PR #13, opened 2020-08-07
fix: isolate PostgREST features from supabase-js

soedirgo:
Currently, some of postgrest-js's functionalities are implemented in supabase-js, like [select](https://github.com/supabase/supabase-js/blob/33939998516e9325e207a63a8a3f10463e89aa41/src/index.js#L137) and [order](https://github.com/supabase/supabase-js/blob/33939998516e9325e207a63a8a3f10463e89aa41/src/index.js#L177). This causes a mismatch of the implementations between the two, which caused https://github.com/supabase/postgrest-js/issues/84.

Since supabase-js is made to act like postgrest-js, we should implement PostgREST-related features solely on postgrest-js and expose that from supabase-js through prototypal inheritance.

I manually tested this on a new Supabase project, but there's some problem with the test suite so I couldn't get them to pass. Will address it in a separate PR.

PR #18, opened 2020-08-10
Add type definitions.

thorwebdev:
This is a WIP draft. Feedback and contributions are very welcome! This relates to https://github.com/supabase/supabase/issues/122

- [x] `createClient`
- [x] `supabase.auth`
- [x] Add detailed comments to `AuthResponse` and `User` params
- [x] Type all other `SupabaseClient` functions
- [x] Figure out how to weave in your custom data types
e.g. this could look like

```ts
interface Message {
  id: number;
  inserted_at: string;
  message: string;
  channel_id: number;
  author?: { name: string };
}

// Resolves with messages: any
const { body: messages } = await supabase
  .from('messages')
  .select('some_column, other_column')

// Resolves with messages: Array<Message>
const { body: messages } = await supabase
  .from<Message>('messages') // Message is the type of row in database
  .eq('channel_id', 1) // IDE will be able to help with the completion of channel_id
  .select(`*, author:user_id(id, username)`) // Not sure if we can do much autocompletion here
  .order('inserted_at', true); // IDE will help with autocompletion
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test:watch": "npm test -- --watch",
     "test:examples": "node examples/",
     "cover": "nyc --check-coverage && npm test",
-    "build": "BABEL_ENV=production babel src --out-dir lib",
+    "build": "BABEL_ENV=production babel src --out-dir lib --copy-files",
     "prepublish": "npm run clean && npm run build",
     "deploy:minor": "npm version minor && npm publish --access=public",
     "deploy:patch": "npm version patch && npm publish --access=public"
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,80 @@
+declare module '@supabase/supabase-js' {
+  interface User {
+    app_metadata: {
+      provider: 'email'
+      [key: string]: any
+    }
+    user_metadata: {
+      [key: string]: any
+    }
+    aud: string
+    created_at: string
+    confirmed_at: string
+    email: string
+    id: string
+    last_sign_in_at: string
+    role: string
+    updated_at: string
+  }
+
+  interface AuthResponse {
+    status: number
+    body: {
+      user: User
+      access_token: string
+      refresh_token: string
+      expires_in: number
+      unauthorized: boolean
+    }
+  }
+
+  interface Auth {
+    /**
+     * Allow your users to sign up and create a new account.
+     * After they have signed up, all interactions using the Supabase JS client will be performed as "that user".
+     */
+    signup: (email: string, password: string) => Promise<AuthResponse>
+    /**
+     * If an account is created, users can login to your app.
+     * After they have logged in, all interactions using the Supabase JS client will be performed as "that user".
+     */
+    login: (email: string, password: string) => Promise<AuthResponse>
+    /**
+     * Get the JSON data for the logged in user.
+     */
+    user: () => Promise<User>
+    /**
+     * After calling log out, all interactions using the Supabase JS client will be "anonymous".
+     */
+    logout: () => Promise<void>
+  }
+
+  interface SupabaseClient {
+    /**
+     * Supabase Auth allows you to create and manage user sessions for access to data that is secured by access policies.
+     */
+    auth: Auth
+  }
+
+  const createClient: (
+    /**
+     * The unique Supabase URL which is supplied when you create a new project in your project dashboard.
+     */
+    supabaseUrl: string,
+    /**
+     * The unique Supabase Key which is supplied when you create a new project in your project dashboard.
+     */
+    supabaseKey: string,
+    options?: {
+      autoRefreshToken: boolean
+      /**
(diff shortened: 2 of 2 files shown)


thorwebdev:
@kiwicopple when I run `npm run build` babel ignores the index.d.ts file. Do you know how we can get babel to copy it over to the `lib` folder?

malobre:
All functions defined in the `Auth` interface may fail for various reasons (wrong credentials, couldn't connect to server, etc) and return a [`ResponseError`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/superagent/index.d.ts#L90).

kiwicopple:
Note on the failing integration tests - this can be ignored. We will have to roll the GoAuth middleware into a new docker image so that it can be run locally. 

We will solve that separately. 

@thorwebdev i'll trust your lead here - it's a non-breaking change so I am happy for you to merge this whenever you're happy. Once it's merged it will be much easier for the community to contribute patches (like the one `malobre` has requested).

I'm happy to merge if you want too :shipit: You're about to make it onto our [humans.txt](https://supabase.io/humans.txt)

thorwebdev:
@malobre the Auth methods don't return the error, but rather throws it which is something you have to handle with a try catch. However this is a good point that this has to be documented here: https://supabase.io/docs/library/user-management

thorwebdev:
@kiwicopple IIRC @awalias wanted to first change the chaining hierarchy in the docs and the dashboard to follow the chaining hierarchy that the types use, e.g. `supabase.from().select().filter().single()`, to make copy and pasting from the docs for TypeScript users more convenient. But I'm happy to not block on that and rather do that as a fast follow, wdyt?

From my pov this is peaceful to merge 👍

malobre:
> @malobre the Auth methods don't return the error, but rather throws it which is something you have to handle with a try catch. However this is a good point that this has to be documented here: https://supabase.io/docs/library/user-management

Nice to know, I wasn't sure how typescript handles errors.

soedirgo:
@thorwebdev IIRC it was because some postgrest-js methods are missing in supabase-js (`limit`, `or`), which https://github.com/supabase/supabase-js/pull/13 is supposed to fix. This delegates the postgrest-js functionality to... postgrest-js, but this means we'll have to wait for the postgrest-js TS transition to get the types.

I personally prefer doing `select()`/`insert()`/`update()` last, but we should let people do both styles.

thorwebdev:
@soedirgo while it seems desirable to be able to call any of the methods at any time, I was worried about the cognitive load (Working memory can generally hold between five and nine items (or chunks) of information at any one time.), and therefore tried to model the major decision tree:

- 1st decision: I want to perform some operation on a table
image
- 2nd decision: I want to read some data
image
- 3rd decision: I want to apply some filters / limits
image
- 4th decision: I want to modify the response shape (single item)
image

This way you can work with the supabase client without needing to consult documentation. The order is only enforced in TypeScript, so any existing JavaScript code will continue working, but I personally think that having a chaining order is generally a good idea, at least for people who don't have all of the available methods memorized. Wdyt?

soedirgo:
You're right, I didn't think of it from a text completion PoV. That way of doing it is more in line with the intended usage.

But say we convert supabase-js to TS. Is it possible to allow arbitrary order of chaining without the compiler yelling? Otherwise I prefer to enforce this order even in JS, since there should at least be method completion even w/o types. This is the best time to do it since I'm working on converting postgrest-js to TS. This will break some code though...

I still think these methods should only be in postgrest-js (`select`, filters, etc.). That way whenever I make a (non-breaking) change on postgrest-js I don't need to update supabase-js.

thorwebdev:
@soedirgo I had a look at how knex.js does this, and it's a bit of a mess IMO:
https://github.com/knex/knex/blob/master/types/index.d.ts
https://lorefnon.tech/2019/05/02/using-boxing-to-prevent-distribution-of-conditional-types/

I personally would be in favor of being opinionated about the chaining order across both JS and TS as it allows us to be coherent across docs and samples, and I think it's a better DX, but also understand that this might be a very subjective opinion. So would need @awalias and @kiwicopple to chime in on this.

> I still think these methods should only be in postgrest-js (select, filters, etc.). That way whenever I make a (non-breaking) change on postgrest-js I don't need to update supabase-js.

Yes, absolutely on board with that 👍

kiwicopple:
We'll have to take a utilitarian approach to developer happiness here. Looks like we're trading flexibility for better/easier docs.

I think the latter gives a better developer experience, so I'm happy to accept Thor's changes & force opinionated chaining.

> This is the best time to do it since I'm working on converting postgrest-js to TS. This will break some code though...

Also we will have to update some of our docs. @soedirgo if you're on board & so is Ant, let's do it. If you think it needs to wait until I'm back to coding, then I can jump into this in ~2 weeks

PR #50, opened 2020-10-14
Feature: migrate to typescript

### LINKED ISSUE #15: `supabase.auth.logout()` throws "Invalid user" error.
I found this to only happen sometimes, but couldn't figure out why:
image

@kiwicopple have you seen this before?

### LINKED ISSUE #20: Auth: Change DX of user management
# Feature request

## Is your feature request related to a problem? Please describe.

As it stands today, you have to do 
```js
try {
const {
  body: { user },
} = await supabase.auth.signup(
  'someone@email.com',
  'password'
)
...
```
To get the user object, and
```js
...} catch (error) {
      console.log('Error message:', error.response.body.error_description);
    } 
```
to get to the error message.

## Describe the solution you'd like

my preferred DX would be 

```js
const { error, user } = await supabase.auth.signup(
  'someone@email.com',
  'password'
);
if (error) alert(error);
```

However, I know that @awalias is a big fan of returning the whole response object, which I personally find rather distracting. Maybe we can do something like

```js
const { error, user, raw_response } = await supabase.auth.signup(
  'someone@email.com',
  'password'
);
if (error) alert(error);
```
To give folks the option to access the raw object, but not overload folks who don't care about most of the stuff in the raw response. Wdyt? This probably warrants a new issue?

## Describe alternatives you've considered

Solve via documentation? Not ideal IMO.

## Additional context

Relates to https://github.com/supabase/supabase/issues/143

### LINKED ISSUE #30: Supabase auth interface missing informiation
# Bug report

## Describe the bug

The Supabase auth object has keys on it like the `accessToken`, `supabaseKey`, `authUrl`, etc. which are missing from the Typescript interface definition: https://github.com/supabase/supabase-js/blob/master/src/index.d.ts#L92

Existing keys:
```
    signup: (email: string, password: string) => Promise<SupabaseAuthResponse>
    login: (email: string, password: string) => Promise<SupabaseAuthResponse>
    user: () => Promise<SupabaseAuthUser>
    logout: () => Promise<void>
```

Keys to be added:
```
	accessToken: string
	authHeader: () => {…}
	authUrl: string
	autoRefreshToken: boolean
	callRefreshToken: () => {…}
	currentUser: SupabaseAuthUser
	persistSession: boolean
	recoverSession: () => {…}
	refreshToken: string
	removeSavedSession: () => {…}
	saveSession: () => {…}
	supabaseKey: string
```

### LINKED ISSUE #31: chore: set up semantic releases
We should automate the releases, with fixes/breaking changes, and releasing to NPM

### LINKED ISSUE #32: Major DX change: response and error handling
## Decision

We've decided to make a major change to support a better developer experience for working with supabase-js. Instead of throwing an error, we will return the error object as part of the response.

## Example

Given a successful query, e.g. `const response = await supabase.from('products').select('*')`, `response` will be:

```
{
  "error": null,
  "data": [...],
  "status": 200,
  "statusCode": 200,
  "statusText": "",
  "body": [...] // for backwards compatibility: `body === data`
}
```

Given a bad query, e.g. `const response = await supabase.from('productx').select('*')`, `response` will be:

```
{
  "error": {
    "code": 404,
    "message": "relation \"public.productx\" does not exist",
    "details": null,
    "hint": null
  },
  "data": null,
  "status": 404,
  "statusCode": 404,
  "statusText": "",
  "body": null // for backwards compatibility: `body === data`
}
```

## Future DX
This will enable the following experience:

```js
const { error, data: products } = await supabase.from('products').select('*')
if (error) {
  alert(error.message)
  return // abort
}
console.log(`Found ${products.length} products.`)
```

## Additional context

Inspiration has been taken from https://github.com/vercel/swr#quick-start:

```js
import useSWR from 'swr'

function Profile() {
  const { data, error } = useSWR('/api/user', fetcher)

  if (error) return <div>failed to load</div>
  if (!data) return <div>loading...</div>
  return <div>hello {data.name}!</div>
}
```

as well as Stripe.js : https://stripe.com/docs/js/payment_methods/create_payment_method
image

### LINKED ISSUE #49: When no `supabaseKey` is passed in it throws an error
# Chore

## Describe the chore
if `createClient` is called and the `supabaseKey` passed in is `null` or `undefined` then a non-specfic error thrown. This can happen if you make a call like this:
``` 
    return createClient(
      process.env.SUPABASE_URL,
      process.env.SUPABASE_PUBLIC_KEY
    );
```
and the env variable is empty.

The error is:
```
TypeError: Cannot read property 'length' of undefined
    at SupabaseClient.initClient (/Users/zlwaterfield/programming/switch-log/node_modules/@supabase/supabase-js/lib/index.js:7:263)
    at SupabaseClient.select (/Users/zlwaterfield/programming/switch-log/node_modules/@supabase/supabase-js/lib/index.js:7:920)
    at CategoryRepository.findByLabel (webpack-internal:///./repositories/CategoryRepository.ts:12:54)
    at LogEntry._findCategory (webpack-internal:///./services/LogEntryService.ts:86:42)
    at LogEntry.timeStart (webpack-internal:///./services/LogEntryService.ts:17:31)
    at __webpack_exports__.default (webpack-internal:///./pages/api/log.ts:11:31)
    at apiResolver (/Users/zlwaterfield/programming/switch-log/node_modules/next/dist/next-server/server/api-utils.js:8:7)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async DevServer.handleApiRequest (/Users/zlwaterfield/programming/switch-log/node_modules/next/dist/next-server/server/next-server.js:45:397)
    at async Object.fn (/Users/zlwaterfield/programming/switch-log/node_modules/next/dist/next-server/server/next-server.js:37:176)
```

We should add validation that the `supabaseKey` is passed in or at least fix [this](https://github.com/supabase/supabase-js/blob/33939998516e9325e207a63a8a3f10463e89aa41/src/index.js#L96) line so the error is better handled.

kiwicopple:

Migrates the repo to Typescript.

- Fixes #32 Major DX change: response and error handling
- Fixes #49 When no `supabaseKey` is passed in it throws an error
- Fixes #31 chore: set up semantic releases
- Fixes #15 `supabase.auth.logout()` throws "Invalid user" error.
- Fixes #20 Auth: Change DX of user management
- Fixes #30 Supabase auth interface missing informiation
- Fixes https://github.com/supabase/supabase/issues/147 https://github.com/supabase/supabase/issues/147
- Partial fix for https://github.com/supabase/realtime-js/issues/53  - if there is no token provided. The error needs to be caught at a socket level.

## BREAKING CHANGES

- Errors are now returned and not thrown
- Auth now uses `@supabase/gotrue-js` interface
- `supabase.getSubscriptions()` only returns open subscriptions


src/Client.ts
```diff
@@ -0,0 +1,168 @@
...
+      },
+      on: (event: string, callback: Function) => {
+        if (!this.realtime.isConnected()) {
+          this.realtime.connect()
+        }
+        return subscription.on(event, callback)
+      },
+    }
+    return builder
+  }
```
kiwicopple: I'm not entirely happy with this one. Is there a way to extend a function that is typescript friendly?

I tried this but Typescript didn't like it because it doesn't seem to pick up the that it is returning a `PostgrestClient` after I call `bind()`

```
  let rest = this._initPostgRESTClient()
  const builder = {
      rest,
      subscription,
      select: rest.from(tableName).select.bind(rest),
      insert: rest.from(tableName).insert.bind(rest),
      update: rest.from(tableName).update.bind(rest),
      delete: rest.from(tableName).delete.bind(rest)
    }
```

I think I need to do some with [TypeScript settings](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#strictbindcallapply)

soedirgo: Doing the above gives me this error (also the case with the TS setting above turned on):

```
Return type of public method from exported class has or is using name 'PostgrestFilterBuilder' from external module [...] but cannot be named.
```

This kind of works:

```js
const builder = rest.from(tableName)
Object.assign(builder, {
  subscription,
  on: (event: 'INSERT' | 'UPDATE' | 'DELETE' | '*', callback: Function) => {
    if (!this.realtime.isConnected()) {
      this.realtime.connect()
    }
    return subscription.on(event, callback)
  },
})
```

But then there's no autocomplete for `on()`. I can't think of a way to do this without creating a new type, e.g. using [mixin](https://www.typescriptlang.org/docs/handbook/mixins.html).

kiwicopple: OK cool - I'm going to try exporting the `PostgrestQueryBuilder` from `postgrest-js`, then I will extend it with the `on()` function.

I have a few other things to tidy up here which I'm working on now

thorwebdev:
When I try to build locally I get a couple of dependency errors:
image
Do you only have these dependencies locally atm? Or maybe package.json is out of date?

kiwicopple:
It looks like the PR for PostgREST didn't get picked up by Semantic Releases: 
https://github.com/supabase/postgrest-js/pull/119

I'll bump both of these today to make sure they are in NPM

src/SupabaseClient.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    // let rest = new SupaabseQueryBuilder(this.restUrl || '', {})
+    let rest = this._initPostgRESTClient()
+    let subscription = new SupabaseRealtimeClient(this.realtime, this.schema, tableName)
+
+    const builder: SupabaseQueryClient<T> = {
+      rest,
+      subscription,
+      select: (columns) => {
+        return rest.from(tableName).select(columns)
```
thorwebdev: this needs to be `return rest.from<T>(tableName).select(columns)` to make sure we pass the type down to postgrest-js. 

Did you have a chance to look into the `SupabaseQueryClient<T> extends PostgrestQueryBuilder<T>` route instead? If that doesn't work this is probably fine as long as we pass down the `<T>` in all the `from`s.

thorwebdev: This reminds me, I think we wanted to call it `SupabaseQueryBuilder` instead of `SupabaseQueryClient`, no?

kiwicopple: Thanks for this - still getting my head around passing through the generics.

Also, do you know why the DocComments don't show up in the intellisense once we get down to the PostgrestFilterBuilder?

image

kiwicopple: I've pushed the latest changes, and adding tests now, but for the life of me I can't figure out why the DocComments aren't getting picked up. @soedirgo do you know why?

FYI, to make it easier, I added a ToDo app in the `example` folder . To use it:

- Run `npm run build`
- Create an app in Supabase and run the To Do List quickstart
- Go into /example 
  - add a `.env.local` file and add `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_KEY`
  -`npm run dev`

This example covers all the functionality, but I can't get the Intellisense to come through 😠

thorwebdev: > I've pushed the latest changes, and adding tests now, but for the life of me I can't figure out why the DocComments aren't getting picked up. 

It does for me
image
You might have to hit `control + space` in order for vscode to fully expand them.

kiwicopple: > You might have to hit control + space in order for vscode to fully expand them.

Wow. you should tweet this. TIL

soedirgo: Ah, looks like it was resolved. `select()` was giving me `any` earlier so I was missing autocompletion.

FWIW, I managed to do this by extending `PostgrestQueryBuilder<T>`:

```ts
// SupabaseClient.ts
const builder = rest.from<T>(tableName) as SupabaseQueryBuilder<T>
builder.subscription = subscription
builder.realtime = this.realtime

// types.ts
export class SupabaseQueryBuilder<T> extends PostgrestQueryBuilder<T> {
  subscription!: SupabaseRealtimeClient
  realtime!: RealtimeClient

  on(event: 'INSERT' | 'UPDATE' | 'DELETE' | '*', callback: Function) {
    if (!this.realtime.isConnected()) {
      this.realtime.connect()
    }
    return this.subscription.on(event, callback)
  }
}

```

kiwicopple: >  `subscription!`

Damn there's no end to obscure TypeScript syntax. This seems better Bobbie, thanks!

thorwebdev:
Couple of updates across libraries:

- [ ] Merge & release https://github.com/supabase/postgrest-js/pull/122
- [ ] Merge & release https://github.com/supabase/gotrue-js/pull/16
- [ ] Merge https://github.com/kiwicopple/supabase-js/pull/1/
- [ ] Update versions of `postgrest-js` & `gotrue-js`
- [ ] Merge this PR 🥳

kiwicopple:
> Merge this PR 🥳

Will do this as soon as all the docs are completed

PR #72, opened 2020-11-11
Add back UMD to npm builds

soedirgo:

Patch update.

When doing the TypeScript transition, we were assuming `<script type="module">` is enough, but it seems being able to use it outside of module `<script>`s is still convenient, so we're adding it back (just for supabase-js).

## What is the new behavior?

You can now use plain `<script>`s to import supabase-js from CDNs, like `<script src="https://cdn.jsdelivr.net/npm/@supabase/supabase-js"></script>` or `<script src="https://unpkg.com/@supabase/supabase-js"></script>` as before.

Then you can use it from a global `supabase` variable:

```html
<script>
const { createClient } = supabase
supabase = createClient("url", "key")
...
</script>
```

## Additional context

Fixes #278.


kiwicopple:
Nice one Bobby 🙏 

Fixes https://github.com/supabase/supabase/issues/285

PR #74, opened 2020-11-15
fix/types exclude

duncanhealy:
- exclude node_modules from tsconfig
- add types export to package.json add types-generate command

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   ],
   "main": "dist/main/index.js",
   "module": "dist/module/index.js",
+  "types": "dist/main/index.d.ts",
   "repository": "supabase/supabase-js",
   "scripts": {
     "clean": "rimraf dist docs",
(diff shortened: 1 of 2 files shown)


kiwicopple:
This is cool @duncanhealy - I'm guessing we run this every time we deploy to NPM? 

If so, we have a few options:

- add it to the GitHub action 
- add `types-generate` to the `build` command
- change `types-generate` to `build:types` and it will get picked up automatically by the build command.

duncanhealy:
The tsc commands are already outputting the index.d.ts 
but it is a little different to the dts-gen output (more fine-grained)
I like to use it for sanity check for the build output like the fixtures / snapshots for jest
if the output changes then you have a breaking change in your library
different surface, can bump minor/major version accordingly

and webpack is using ts-loader so all should  be good

tsc should fail in ci.yml 

```tsc -p tsconfig.json``` should have failed before 
but it may have been the upstream libs (in node_modules) that failed so that shouldn't happen anymore

PR #110, opened 2020-12-30
Add `"sideEffects": false` to pkg.json

BeeeQueue:

Feature/fix

The package is not currently marked as side effect free so tree-shaking will be limited.

## What is the new behavior?

Tree-shaking can be done properly.

---

I didn't see anything that has side effects in the package, but if you know any then this should not be done


kiwicopple:
Thanks for this @BeeeQueue . I had to have a think about it. I'm fairly certain there are no sideEffects here so I'll merge this and keep tabs of it in our builds

PR #112, opened 2021-01-02
Added 'User' to the exports

dshukertjr:

In order to get the `User` type, I had to do a import like this

```
import { User } from '@supabase/gotrue-js/dist/main/lib/types'
```
so I thought it would be better to export `User` from supabase-js

`User` is not exported within supabase-js

## What is the new behavior?

`User` is exported as `AuthUser`. Developers should be able to import the `AuthUser` type like this:

```
import { AuthUser } from "supabase-js";
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,6 @@
 import SupabaseClient from './SupabaseClient'
 import { SupabaseClientOptions, SupabaseRealtimePayload } from './lib/types'
+import { User } from '@supabase/gotrue-js'
 export * from '@supabase/gotrue-js'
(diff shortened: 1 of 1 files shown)


kiwicopple:
Good idea @dshukertjr , but perhaps we should export this one as `AuthUser`? With GoTrue it isn't really necessary, but this library has a long way to go, so hopefully this will prevent conflicts in the future.

dshukertjr:
@kiwicopple Sure, let me update the code right now.

kiwicopple:
Legend, thanks @dshukertjr !

PR #116, opened 2021-01-13
feat: add EventTypes as a Type

ftonato:

Add a new Type to avoid use duplicated code.

We're duplicating code for lack of specific data type.

## What is the new behavior?

Creating this new type, we can remove the duplicated code using the new created type.

## Additional context

JS is amazing, isn't?


kiwicopple:
🔥

kiwicopple:
I've merged this, but now that I think about it, this should probably be in our realtime-js library - 
https://github.com/supabase/realtime-js/

ftonato:
Hey @kiwicopple,

Don’t worry, this is how life works, first we do it, then we think and improve it  🚀

PR #136, opened 2021-03-18
feat: storage api

phamhieu:

Implement storage api with a `next-storage` example app.

To run example app:
- clone `supabase-js` and checkout `feat/storage-api` branch
- build the library with `npm run build`
- `cd example/next-storage/`
- create `.env.local` file under `next-storage` app
- run example app with `npm run dev`

```js
// .env.local for testing
NEXT_PUBLIC_SUPABASE_URL=https://iptgkdvhveztktwedveb.supabase.net
NEXT_PUBLIC_SUPABASE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoiYW5vbiIsImlhdCI6MTYxNjAyNjU5MywiZXhwIjoxOTMxNjAyNTkzfQ.pJ4KSEklFuT9Isou259JoUtfQ9xl1zO1zxyk_2Q51Dw
```

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Screenshot 2021-03-18 at 1 48 41 PM


phamhieu:
after you upload avatar if you refresh the browser it will show `no avatar available`. You have to sign-out and sign-in again to sync user data.

related to this bug on gotrue-js https://github.com/supabase/gotrue-js/pull/67

src/lib/storage/StorageApi.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  /**
+   * Create signed url to download file
+   * @param path the relative file path to be downloaded
+   * @param expiresIn seconds until the signed URL expires
+   */
+  async createSignedUrl(
+    path: string,
+    expiresIn: number
```
kiwicopple, line 189: I wonder if this should be set on the Server? Unsure, but I would imagine we don't want users to generating a tonne of signed URLs which last 1 year?

src/lib/storage/StorageApi.ts
```diff
@@ -0,0 +1,269 @@
...
+    } catch (error) {
+      return { data: null, error }
+    }
+  }
+
+  /**
+   * Download a file
+   * @param path the relative file path to be downloaded
+   */
+  async downloadFile(path: string): Promise<{ data: Blob | null; error: Error | null }> {
```
kiwicopple, line 207: Thoughts on making this `download()`?

kiwicopple: Perhaps not actually - we have `deleteFile` below, which we can't turn into `delete` because it's a reserved word. For consistency and accuracy, `downloadFile()` is probably better

src/lib/storage/StorageApi.ts
inian: I think it is okay to be set on the client. If the client has read access to the object, he can share that object to anyone. There are usecases for both short and long expiry times. Short expiry is useful for generating a temporary download link to the object and long times are useful say if you want to embed the resource on your website markup for example ``.

example/next-storage/components/Avatar.tsx
```diff
@@ -0,0 +1,30 @@
...
+export type UploadButtonProps = {
+  avatar?: string | null
+}
+
+export default function Avatar(props: UploadButtonProps) {
+  const [avatarUrl, setAvatarUrl] = useState<string | null>(null)
+  const { avatar } = props
+
+  useEffect(() => {
+    async function downloadImage(path: string) {
```
kiwicopple: Error here (FF):

image

Debugging now

kiwicopple: wrapped it in a try/catch - data was null, error was `error The requested bucket was not found`

phamhieu: Oh yeah, need to setup the bucket on supabase project first.

Maybe we should extract the default bucket name **avatars** to constants.ts

kiwicopple: I'll add it to the schema in the readme, but I'm still trying to make this work with a project on staging. Do you have any Policies enabled for your Buckets/objects?

phamhieu: I have to disable RLS on `objects` table. @inian gave some examples here for the policy https://github.com/supabase/storage-api/blob/9991eef42de4b6d64941c0883ecaa2eceb936546/src/test/db/03-dummy-data.sql#L39-L46

example/next-storage/pages/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  useEffect(() => {
+    setSession(supabase.auth.session())
+    getPublicProfiles()
+
+    supabase.auth.onAuthStateChange((_event: string, session: AuthSession | null) => {
+      setSession(session)
+    })
+
+    realtimeProfiles = supabase
```
kiwicopple: @phamhieu you had a problem with realtime only returning one item right? What code changes did you make to fix it? I want to make sure we're suggesting the best method for next.js/react

phamhieu: https://github.com/phamhieu/realtime-map/commit/f2d50019e21588c30405243b853d73e45717e72d

I update the code to use `useReducer`. Then on realtime callback i use `dispatch` to trigger change on my UI. It works nicely

kiwicopple: Nice 🔥

src/lib/storage/StorageApi.ts
```diff
@@ -0,0 +1,305 @@
...
+   * Upload a file
+   * @param path the relative file path
+   * @param file the File content
+   */
+  async uploadFile(
+    path: string,
+    file: File
+  ): Promise<{ data: { message: string } | null; error: Error | null }> {
+    try {
+      if (!isBrowser()) throw new Error('No browser detected.')
```
kiwicopple, line 101: I'll have to check what other libs are doing here for the Node environment

PR #137, opened 2021-03-25
Chore/storage refactor

kiwicopple:

Better DX, better docs

### DIFF
--- a/src/lib/storage/StorageApi.ts
+++ b/src/lib/storage/StorageApi.ts
@@ -166,19 +166,19 @@ export class StorageApi {
   /**
    * Renames an existing file.
    *
-   * @param bucketName The bucket which contains the file.
+   * @param bucketId The bucket which contains the file.
    * @param fromPath The original file path, including the current file name. For example `folder/image.png`.
    * @param toPath The new file path, including the new file name. For example `folder/image-copy.png`.
    */
   async renameFile(
-    bucketName: string,
+    bucketId: string,
     fromPath: string,
     toPath: string
   ): Promise<{ data: { message: string } | null; error: Error | null }> {
     try {
       const data = await post(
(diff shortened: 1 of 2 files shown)


src/lib/storage/StorageApi.ts
```diff
@@ -233,17 +244,18 @@ export class StorageApi {
   }
 
   /**
-   * Delete multiple files on the same bucket
-   * @param bucketName
-   * @param paths the relative file paths to be deleted excluded the bucket name
+   * Deletes multiple files within the same bucket
+   *
+   * @param bucketId The bucket which contains the files.
+   * @param paths An array of files to be deletes, including the path and file name. For example [`folder/image.png`].
```
kiwicopple, line 250: it looks like the user doesn't have to include the file name right?

For example, if they put `/folder`, it would delete the entire folder?

src/lib/storage/StorageApi.ts
```diff
@@ -157,20 +164,21 @@ export class StorageApi {
...
-  async renameFile(
-    bucketName: string,
+  async moveFile(
+    bucketId: string,
     fromPath: string,
     toPath: string
   ): Promise<{ data: { message: string } | null; error: Error | null }> {
     try {
       const data = await post(
         `${this.url}/object/rename`,
```
kiwicopple: @inian keep it as `rename` here though right?

src/lib/storage/StorageApi.ts
inian: No. The user will have to get the full path to delete. Having a separate endpoint to delete a folder is on the roadmap.

src/lib/storage/StorageApi.ts
inian: yup. I will update this after the backend change is done

PR #138, opened 2021-03-25
Add additional steps to setup storage example

fracek:

I add a new step in the `next-storage` example readme to let users know they need to create a storage bucket and set the correct policies before they can run the example.


inian:
Hi @fracek, looks like the README was updated with these instructions at some point. Closing the PR. 
Thanks!

PR #139, opened 2021-03-25
Prettify storage example app

joshenlim:

Just prettifying the storage example app. Pardon the mess of a CSS, but at minimum I cleaned up all inline styling into the global.css file

Sneakpeeks:

image


kiwicopple:
This is epic, thanks @joshenlim 🔥 🔥 🔥 🔥

PR #140, opened 2021-03-27
Added storage policies to readme.md of next-storage example app

dshukertjr:

Added sample storage policies to the next-storage example app's readme.md

There are no example policies to get started with. 

## What is the new behavior?

There are SQLs that users can run to get up and running with Supabase Storage!

## Additional context

I added typescript dependencies to the project, because it was missing, but let me know if this was unnecessary.

### DIFF
--- a/example/next-storage/README.md
+++ b/example/next-storage/README.md
@@ -31,4 +31,10 @@ begin;
 commit;
 
 alter publication supabase_realtime add table profiles;
+
+-- Set up Storage!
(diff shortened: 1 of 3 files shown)


dshukertjr:
Obviously there is a lot of speculations of what kind of policies would be best for this particular example app, so I am open to suggestions for editing the current policies.

kiwicopple:
wow very nice @dshukertjr ! I like the substring policies :)

PR #142, opened 2021-03-30
refactor: SupabaseStorageClient

phamhieu:

Separating `StorageBucketApi` and `StorageFileApi`, so that we can call storage api with this syntax

```
supabase.storage.from('avatars').downloadFile('avatar1.png')
```

### DIFF
--- a/src/lib/storage/StorageFileApi.ts
+++ b/src/lib/storage/StorageFileApi.ts
@@ -1,6 +1,6 @@
-import { get, post, put, remove } from './fetch'
+import { get, post, remove } from './fetch'
 import { isBrowser } from './helpers'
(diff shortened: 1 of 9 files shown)


src/lib/storage/StorageFileApi.ts
```diff
@@ -311,23 +239,25 @@ export class StorageApi {
...
         headers: this.headers,
       })
       return { data, error: null }
     } catch (error) {
       return { data: null, error }
     }
   }
+
+  _getFinalPath(path: string) {
+    return `${this.bucketId}\\${path}`
```
inian: Why the backslash here @phamhieu ?

Issue #148, opened 2021-04-05
createSignedURL returning two properties, broken URL generated


ProjectMoon:
# Bug report

## Describe the bug

The storage `createSignedURL` function is returning a result of `{ signedUrl, signedURL }`. The `signedUrl` should be the one with the full URL to download the file. But on the end of this URL, it has `undefined`. The `signedURL` contains the rest of the URL that should be appended to the domain and path in `signedUrl`.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. call `createSignedURL`.
2. Inspect the returned value.

## Expected behavior

Only one `signedUrl` property in the returned object, with the full URL properly present.

## Screenshots

The returned value looks like this:

```
{
  "data": {
    "signedURL": "/object/sign/test/test.png?token=blahblahblah",
    "signedUrl": "https://example.supabase.co/storage/v1undefined",
    "error": null
  }
}
```

inian:
Thanks for the report @ProjectMoon! I have fixed this in supabase-js v1.8.0 
The return value only has signedURL with the fully qualified URL.

Ascarbek:
@inian Hi! the thing is return value should have `signedUrl` not `signedURL` according to it's type. Typescript is pointing to it as an error in my project. 
``` TS2551: Property 'signedURL' does not exist on type '{ signedUrl: string; }'. Did you mean 'signedUrl'?```

inian:
Hi @Ascarbek, I have pushed a fix for this just today. Can you update to the latest version of `supabase-js`?

Ascarbek:
@inian I've upgraded to 1.11.4 but it still returns `signedURL`
I mean the HTTP request to storage is returning `signedURL` I don't think it's `supabase-js` problem

inian:
That's the expected behaviour - Server returns signedURL and supabase-js also returns signedURL. There should be no signedUrl anywhere now.

Ascarbek:
@inian I got it! Now I don't have Typescript error. Thank you!

PR #149, opened 2021-04-06
feat: updates createSignedURL to return full URL

inian:
fixes https://github.com/supabase/supabase-js/issues/148

Previous return value for `createSignedURL`
```
{
  "data": {
    "signedURL": "/object/sign/test/test.png?token=blahblahblah",
    "signedUrl": "https://example.supabase.co/storage/v1undefined",
    "error": null
  }
}
```

New return value
```
{
  "data": {
    "signedURL": "https://example.supabase.co/storage/v1/object/sign/test/test.png?token=blahblahblah",
    "error": null
  }
}
```

### DIFF
--- a/src/lib/storage/StorageFileApi.ts
+++ b/src/lib/storage/StorageFileApi.ts
@@ -145,8 +145,7 @@ export class StorageFileApi {
         { expiresIn },
         { headers: this.headers }
       )
-      const signedUrl = `${this.url}${data.signedUrl}`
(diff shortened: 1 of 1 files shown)


taktran:
In case anyone else like me got confused with `signedURL` no longer working as expected in `v2.0`, it got removed changed back to `signedUrl` in https://github.com/supabase/storage-js/pull/94

Issue #179, opened 2021-05-25
Incorrect Type for callback parameter in .on callback on SupabaseRealtimeClient


PhoenixCreation:
# Bug report
callback function of .on() should have `callback: (payload: SupabaseRealtimePayload<T>) => void` type instead of just function.

## Describe the bug
If you channel the realtime event with .on() then first callback function's parameter gets the type of `SupabaseRealtimePayload` but for second .on() it has type `any`.

Here is a clear and concise description of what the bug is.
let's take this code for example:
```js
const unsubscriber = supabase
          .from(tableName)
          .on("UPDATE", (updatedChat) => {
            console.log("Updated chat", updatedChat.new);
          })
          .on("INSERT", (insertedChat) => {
            console.log("Inserted chat", insertedChat.new);
          })
          .subscribe();
```
In this example if you check the type of "updatedChat" then it is SupabaseRealtimePayload but if you check type of "insertedChat" then it is `any`.

## To Reproduce

Just channel the .on() for one table for two or more events. Screenshots will clear what I am telling better and will give you idea what I am trying to say and how to reproduce it.

## Expected behavior

All the callback parameters of .on() should have SupabaseRealtimePayload type.

## Screenshots

![ok_one](https://user-images.githubusercontent.com/63447220/119447912-389f6100-bd4e-11eb-9a04-c5a8f605198b.png)
![problematic_one](https://user-images.githubusercontent.com/63447220/119447924-3a692480-bd4e-11eb-80af-4d03529764d3.png)

## System information

- OS: Windows 7(32bit)
- Version of supabase-js: ^1.11.14

This is irrelevant so I omitted some info.

## Additional context- Solution

I am not pro in TypeScript so I am not 100% confident about this but I am 99% confident.

changing the file: src/lib/SupabaseRealtimeClient.ts
line no: [35 click to jump to this file.](https://github.com/supabase/supabase-js/blob/329edd0757431478c48a2190796ba97ca73de939/src/lib/SupabaseRealtimeClient.ts#L35)

changes: 
```diff
- on(event: SupabaseEventTypes, callback: Function) {
+ on(event: SupabaseEventTypes, callback: (payload: SupabaseRealtimePayload<any>) => void) {
```

Let me know if this is right solution or not so I can create PR according to it. I didn't wanted to just start a PR without consenting to anyone.

Thank you in advance for such great platform.

PhoenixCreation:
I created PR anyway. merge it if you think this solves the issue. #180

[...]

PR #180, opened 2021-05-25
fix: type error for callback of on() in SupabaseRealtimeClient

PhoenixCreation:
Parameter of callback function of on() should have SupabaseRealtimePayload type instead of any Function.

Type error fix. Solves the #179. Detailed information is given in issue.

callback function has type Function which can be replaced with `(payload: SupabaseRealtimePayload<T>) => void`
#179


kiwicopple:
thanks @PhoenixCreation !

PR #201, opened 2021-06-11
Bump postgrest-js

### LINKED ISSUE #203: [Typescript error] match filter only accepts string values
# Bug report

## Describe the bug
Incorrect types for `match` filter. It should accept any types of values
https://supabase.io/docs/reference/javascript/match

## To Reproduce

Add this code to a typescript file and it'll generate error at line 4 saying
`Type 'number' is not assignable to type 'string'`

```js
  .from('cities')
  .select('name, country_id')
  .match({name: 'Beijing', country_id: 156})

```

I should change the values to string inorder to get rid of the errors

```js
  .from('cities')
  .select('name, country_id')
  .match({name: 'Beijing', country_id: '156'})  <----------

```

## Expected behavior

It shouldn't throw any type error when passing  types other than string

## Screenshots
Screen Shot 2021-06-12 at 17 15 10

Also, I'm importing types form `'@supabase/postgrest-js'`.  `@supabase/supabase-js` doesn't have any exported types.

```js
import { PostgrestFilterBuilder } from '@supabase/postgrest-js';
```

## System information

-  "typescript": "^4.3.2"
- "@supabase/supabase-js": "^1.14.0",

soedirgo:
Implement unwrap() to throw query errors instead of returning them: https://github.com/supabase/postgrest-js/pull/188

fix: change match's value type to unknown: https://github.com/supabase/postgrest-js/pull/192

Closes #203.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@supabase/gotrue-js": "^1.16.2",
-        "@supabase/postgrest-js": "^0.31.0",
+        "@supabase/postgrest-js": "^0.31.1",
         "@supabase/realtime-js": "^1.0.9",
         "@supabase/storage-js": "^1.2.1"
       },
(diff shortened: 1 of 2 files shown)


kiwicopple:
Awesome @soedirgo - did we have any docs for this? If you drop a note to me here I can add them to our website

soedirgo:
Hmm, I don't think we do—I think we can add this as an example in Reference > Data > select():

## Throwing on errors

Instead of handling errors using the returned `error` field, you can make it throw on errors instead:

```js
try {
    .from('missing_table')
    .select()
    .throwOnErrors()
  // do something with `data`
} catch (e) {
  // handle error
}
```

Issue #236, opened 2021-08-15
count for rpc is not passed to postgrest-js


e0:
# Bug report

## Describe the bug

In [`postgrest.js`](https://github.com/supabase/postgrest-js/blob/master/src/lib/PostgrestRpcBuilder.ts#L18), the `rpc` function allows for a parameter that accepts an object with `count`. It looks like the object with `count` [does not get passed](https://github.com/supabase/supabase-js/blob/master/src/SupabaseClient.ts#L99-L102) to `postgrest.js` from `supabase.js` as this is the current function definition:

```typescript
  rpc<T = any>(fn: string, params?: object) {
    const rest = this._initPostgRESTClient()
    return rest.rpc<T>(fn, params)
  }
```

Currently, the [following example](https://supabase.io/docs/reference/javascript/rpc#with-count-option) mentioned in the documentation would not work. (This specific example would currently not work in `postgrest.js` either due to [a related bug](https://github.com/supabase/postgrest-js/issues/208).).

```javascript
const { data, error, count } = await supabase
  .rpc('hello_world', { count: 'exact' })
```

## To Reproduce

Running something like `.rpc('hello_world', { count: 'exact' })` should fail since there would be no matching stored procedure.

Running something like `.rpc('hello', { name: 'world' }, { count: 'exact' })` should return a result where `count` is null since `{ count: 'exact' }` is not passed to `postgrest-js`.

## Expected behavior

When I run something like `.rpc('hello', { name: 'world' }, { count: 'exact' })`, I expect to receive a `count` with an integer value in the result.

## System information

- OS: macOS
- Browser: Safari
- Version of supabase-js: 1.21.0
- Version of Node.js: v16.4.2

PR #237, opened 2021-08-16
Fixes supabase/supabase-js#236

zernonia:
Fix the rpc count missing in the function

`supabase.rpc('fnc_name', { params }, { count: "exact" })` is missing the count object parameter.

## What is the new behavior?
Fix the function, able to create rpc function that return `count`


soedirgo:
Awesome, thanks for the PR @zernonia!

PR #238, opened 2021-08-20
Feat/client info header

awalias:
for https://github.com/supabase/infrastructure/issues/1044


awalias:
oh hold on

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@supabase/supabase-js",
-  "version": "1.0.0",
+  "version": "0.0.0",
```
awalias, line 3: @inian I think this is safe to revert back to 0.0.0? It's not used anywhere, right?

soedirgo: Yeah semantic-release will update this automatically so this is unused.

awalias: we might want to have semantic-release also check in this change

PR #252, opened 2021-09-04
Access localStorage lazily to avoid DOMExceptions in constrained envs when `localStorage` is provided using options

Andarist:
It's a follow up to: https://github.com/supabase/gotrue-js/pull/117 - just adjusting how options are passed around so one can leverage those downstream changes.


soedirgo:
Thanks @Andarist!

soedirgo (approved):

PR #258, opened 2021-09-14
refactor: create a new type to avoid repeat ourselves and add the correct return type for createClient method

ftonato:

Refactor our codebase creating a new type to avoid repeat ourselves and add the correct return type for createClient method

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -1,5 +1,5 @@
 import { DEFAULT_HEADERS } from './lib/constants'
-import { SupabaseClientOptions } from './lib/types'
+import { GenericObject, SupabaseClientOptions } from './lib/types'
 import { SupabaseAuthClient } from './lib/SupabaseAuthClient'
 import { SupabaseQueryBuilder } from './lib/SupabaseQueryBuilder'
 import { SupabaseStorageClient } from '@supabase/storage-js'
@@ -116,7 +116,7 @@ export default class SupabaseClient {
       try {
         await this._closeSubscription(subscription)
 
-        const openSubscriptions = this.getSubscriptions().length
+        const openSubscriptions = this._getSubscriptions().length
         if (!openSubscriptions) {
           const { error } = await this.realtime.disconnect()
           if (error) return resolve({ error })
@@ -137,7 +137,7 @@ export default class SupabaseClient {
   /**
    * Returns an array of all your subscriptions.
    */
-  getSubscriptions(): RealtimeSubscription[] {
+  _getSubscriptions(): RealtimeSubscription[] {
     return this.realtime.channels
   }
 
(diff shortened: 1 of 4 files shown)


adam-beck:
Wouldn't `Record<string, string>` accomplish the same thing? And, since it's a built-in utility type, it would be easier to read and understand.

ftonato:
> Wouldn't `Record<string, string>` accomplish the same thing? And, since it's a built-in utility type, it would be easier to read and understand.

You're totally right, they both do the same, it's just a matter of custom for those who have been using TS since before version 2.1 (which is when the records arrived) 🐶

About being easier to read, to me they both seem simple (again, seems like a particular preference)

src/SupabaseClient.ts
```diff
@@ -137,7 +137,7 @@ export default class SupabaseClient {
   /**
    * Returns an array of all your subscriptions.
    */
-  getSubscriptions(): RealtimeSubscription[] {
+  _getSubscriptions(): RealtimeSubscription[] {
```
w3b6x9: @ftonato i would keep it as `getSubscriptions` b/c developers can use it to grab their subscriptions and it'll break existing code (e.g. `client.getSubscriptions()`)

ftonato: @w3b6x9 Sorry for the change! I thought the method was only used internally in the class, hence the idea of initialising it with underscore. Anyway, I "removed" the change and fixed the conflicts...

w3b6x9: @ftonato no worries and thank you!

src/SupabaseClient.ts
```diff
@@ -254,8 +255,8 @@ export default class SupabaseClient {
     })
   }
 
-  private _getAuthHeaders(): { [key: string]: string } {
-    const headers: { [key: string]: string } = this.headers
+  private _getAuthHeaders(): GenericObject {
+    const headers: GenericObject = DEFAULT_HEADERS
```
w3b6x9: @ftonato i think the merge commit (https://github.com/supabase/supabase-js/pull/258/commits/19647931b45b6ee8a16c62a7608de826b18e91a6) messed this up:

```suggestion
    const headers: GenericObject = this.headers
```

src/lib/SupabaseQueryBuilder.ts
```diff
@@ -19,7 +19,7 @@ export class SupabaseQueryBuilder<T> extends PostgrestQueryBuilder<T> {
       table,
       fetch,
     }: {
-      headers?: { [key: string]: string }
+      headers?: GenericObject
```
w3b6x9, line 22: @ftonato just one more thing. `GenericObject` needs to be imported.

w3b6x9 (changes requested):

ftonato:
@w3b6x9 First and last time I try to resolve conflict by browser hahaha

w3b6x9:
> @w3b6x9 First and last time I try to resolve conflict by browser hahaha

haha wow mobile browser...that's quite the challenge!

w3b6x9 (approved):
lgtm! thanks!

PR #264, opened 2021-09-18
fix: bump gotrue-js@1.18.0 with user metadata on signup

Inzeppelin:

Bump https://github.com/supabase/gotrue-js @1.18.0 

Adding arbitrary data to the auth.user table is only possible via the update call.

## What is the new behavior?

Lets you add metadata when signing up a user instead of having to sign up then update.

## Additional context

This version includes https://github.com/supabase/gotrue-js/pull/130 provided by @danalloway


soedirgo (approved):

Inzeppelin:
@soedirgo awesome! thank you! Is it possible to push new tag with this change?

PR #269, opened 2021-09-22
chore: fix CI release process

darora:
- update to node16 from 12 (semantic-release now requires at least node14)
- disable commenting on PRs/issues included in the release (semantic release seems to be running into some github API issue)

Could pin to the last major release of semantic release to avoid bumping node versions for a little while.

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -16,10 +16,10 @@
     "@semantic-release/commit-analyzer",
     "@semantic-release/release-notes-generator",
     ["@semantic-release/github", {
-      successComment: false,
-      releasedLabels: false,
-      failTitle: false,
-      addReleases: false,
+      "successComment": false,
+      "releasedLabels": false,
+      "failTitle": false,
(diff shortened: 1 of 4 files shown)


.releaserc.json
```diff
@@ -15,7 +15,12 @@
     ],
     "@semantic-release/commit-analyzer",
     "@semantic-release/release-notes-generator",
-    "@semantic-release/github",
+    ["@semantic-release/github", {
+      successComment: false,
+      releasedLabels: false,
+      failTitle: false,
+      addReleases: false,
```
soedirgo: Invalid JSON? Should be quoted

inian (approved):

soedirgo:
How about using Node 14? Since we're using that across the board and 16 isn't LTS.

darora:
done, untested though since in my other repos I've used node16. Works in theory

soedirgo (approved):

PR #271, opened 2021-09-22
fix: change release.yml node version to 14

w3b6x9:

Bug fix

`semantic-release` breaking change raised minimum node version to 14.17 (see [here](https://github.com/semantic-release/semantic-release/pull/2132)) which is breaking our release (our node version is 12).

## What is the new behavior?

Sets our release node version to 14.


w3b6x9:
dupe of https://github.com/supabase/supabase-js/pull/269

PR #281, opened 2021-10-05
Add JWT Validation

### LINKED ISSUE #197: don't send expired JWTs
We might want to verify the JWT expiry date locally before sending to the server to query for data, then if there's a `refresh_token` stored locally we might be able to refresh it before making the request, or otherwise triggering a listenable event on the client

velocity23:

This PR adds the validation of JWTs whenever a new client (of any service) is initialized.

The JWT is not validated, and invalid JWTs are sent to the API if introduced

## What is the new behavior?

The JWT is validated upon client construction and if it is malformed or expired, an exception is thrown.

## Additional context

Closes #197


src/lib/helpers.ts
```diff
@@ -9,5 +11,29 @@ export function uuid() {
...
+    if (pad === 1) {
+      throw new Error('Invalid JWT - base64Url is an invalid length to determine padding')
+    }
+
+    base64Url += new Array(5 - pad).join('=')
+  }
+
+  const base64 = base64Url.replace('-', '+').replace('_', '/')
+
+  return JSON.parse(atob(base64))
```
kiwicopple: It looks like `atob` is only compatible on Node >16 right?

https://developer.mozilla.org/en-US/docs/Web/API/atob#browser_compatibility

Would this need to be polyfilled?

kiwicopple:
Sorry for the delay here - I didnt have notifications on supabase-js.

I like this approach, thanks for PR'ing @velocity23 . I think we need to just double-check compatibility across all the browsers and node environments.

Would you mind adding a unit test for the helper functions to make sure we catch any regression?

velocity23:
Thanks for taking a look @kiwicopple. I've made the requested changes, improved the error messages, and added a check in the client constructor.

julianpoma:
Thanks for this PR!

Would it be a good idea to dispatch an event `JWT_EXPIRED` or trigger an auto refresh based on the value of `autoRefreshToken`, if the token is expired?

velocity23:
Will work on this in the next few days/week or so

velocity23:
If someone else wants to pick up where I left off you're more than welcome - I don't have time to complete this anymore unfortunately

PR #303, opened 2021-11-18
fix: improve auth for realtime row level security

kiwicopple:
- gotrue-js should now be handled across browser windows/tabs by default
- client will keep socket connections fresh with all auth events

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -227,6 +227,10 @@ export default class SupabaseClient {
       if (event === 'SIGNED_OUT' || event === 'USER_DELETED') {
         this.removeAllSubscriptions()
       }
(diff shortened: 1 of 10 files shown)


darora (approved):

src/SupabaseClient.ts
```diff
@@ -209,4 +226,52 @@ export default class SupabaseClient {
...
+    }
+
+    if (event === 'SIGNED_OUT' || event === 'USER_DELETED') {
+      this.removeAllSubscriptions()
+      this.auth.signOut()
+    }
+
+    // @todo: Handle all other auth events - token refresh, etc.
+    // If the token is not undefined, and it's different from this current tab, we should refresh the
+    // auth client in this supabase instance
```
kiwicopple: refresh the current session here

PR #305, opened 2021-11-22
fix: missing fetch during auth init

soedirgo:
We want to pass custom `fetch` to gotrue-js, but the value is not yet
set when we run `_initSupabaseAuthClient()`. Solution: set custom
`fetch` before running `_initSupabaseAuthClient()`.


kiwicopple (approved):

swapnilrajegithubcom:
cb06fc627940d7759c906a428bf45e3b4df4ce7d

Issue #166, opened 2021-05-03
Headers provided to Supabase client are not used in requests


brattonross:
# Bug report

## Describe the bug

Headers passed to the supabase client upon creation are not sent as part of requests made by the client.

## To Reproduce

**Note:** This reproduction example assumes that you are querying a storage bucket that has policies applied so that only authenticated users may read from the bucket, and that the request is being made outside of a browser environment e.g. from a Next.js API route.

1. Initialize a supabase client, passing in custom headers for the client, e.g.

```js
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(url, anonKey, {
  headers: {
    apiKey: anonKey,
    Authorization: `Bearer ${jwt}`
  }
})
```

2. Make a request using that client, e.g.

```js
const result = await supabase.storage.from('avatars').download(avatarURL)
```

3. Observe that the request does not return the expected result, as the request made by the client does not use the `Authorization` header that was provided on creation

## Expected behavior

Supabase client uses the headers provided in the call to `createClient` when making requests.

In the example above this would set the `Authorization` header on the request to the Storage API, allowing the server to make a request to the restricted content on behalf of the user.

## Screenshots

N/A

## System information

- OS: N/A
- Browser: N/A
- Version of supabase-js: 1.11.9
- Version of Node.js: 14.16.1

## Additional context

This issue was created off the back of [this comment](https://github.com/supabase/supabase/discussions/1094#discussioncomment-688755), and is discussed further in that thread.

ftonato:
Hello,

The **data** has been sending to the correct method `_initSupabaseAuthClient` but inside this method, as we can [see here](https://github.com/supabase/supabase-js/blob/906a076c8942fe86765f0b9430be5b0ad076036a/src/SupabaseClient.ts#L138), we aren't using the param `{ headers: apiKey: 'xx', Authorization: 'xx' }`.

dabarrell:
@brattonross I'm looking at writing a PR to add in header handling for the clients, so we can pass in custom headers. I'm unclear about the example you give - passing auth headers to the storage client.

I note that the storage client is initialised here:
https://github.com/supabase/supabase-js/blob/master/src/SupabaseClient.ts#L74-L76

Which pulls the headers from `_getAuthHeaders()`, setting both the `apikey` and the `Authorization` headers. Was that not the case when you created this issue? and if it was the case, what need do you have to provide custom `apikey`/`Authorization` headers?

[...]

PR #311, opened 2021-11-25
Use custom headers in every client

mgm1313:

Feature

Since https://github.com/supabase/supabase-js/pull/218 custom headers passed to the `SupabaseClient` constructor are passed on to the `GoTrueClient`. 
However they are not passed on to the`PostgestClient`, `SupabaseStorageClient` and `SupabaseQueryBuilder`, although implied by the param description: 
https://github.com/supabase/supabase-js/blob/5f7f3d5f088d213ae56e9a806066c1c810c852ab/src/SupabaseClient.ts#L45

## What is the new behavior?

Headers passed through the `options.headers` parameter of the constructor are now stored in a new protected variable `headers`. Since all clients mentioned above call the private method `_getAuthHeaders` I decided to inject the custom headers here, together with the default headers that were already injected inside this method. 

## Additional context

I also created a very minimal test case, just to be sure the `_getAuthHeaders` method indeed returns the custom headers. I'm not an expert on Jest, so as far as convention goes there might be more optimal ways to test the functionality this PR introduces. 
Closes https://github.com/supabase/supabase-js/issues/166

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -33,6 +33,9 @@ export default class SupabaseClient {
   protected storageUrl: string
   protected realtime: RealtimeClient
   protected fetch?: Fetch
+  protected headers: {
+    [key: string]: string
+  }
 
   /**
    * Create a new client for use in the browser.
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
Thanks @mgm1313!

PR #328, opened 2021-12-15
fix: create subscription only when .on method is called

w3b6x9:

Bug fix

1. A new Realtime subscription is created whenever SupabaseClient `from` method is called, even when user does not intend on using Realtime.
2. Realtime channel is not cleaned up properly from the list of channels on the RealtimeClient. Channel is only removed when it is open and then closed. There are cases, like mentioned above, where a subscription is created but never subscribed to. This causes state to be closed which means the subscription is never cleaned up.

## What is the new behavior?

1. Subscription is created only when intended; when SupabaseQueryBuilder `on` method is called.
2. SupabaseClient does a better job of cleaning up subscriptions regardless of the state.

## Additional context

- Related issue: https://github.com/supabase/supabase/issues/4451

- `removeAllSubscriptions` method on the SupabaseClient can be used to clear away all subscriptions and disconnect the socket connection

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -131,12 +131,11 @@ export default class SupabaseClient {
   }
 
   /**
-   * Remove all active subscriptions.
+   * Remove all subscriptions.
    */
-  removeAllSubscriptions() {
-    this.realtime.channels.forEach((sub) => {
-      this.removeSubscription(sub)
-    })
+  async removeAllSubscriptions() {
+    const subscriptions = this.realtime.channels.slice()
+    return await Promise.allSettled(subscriptions.map((sub) => this.removeSubscription(sub)))
   }
 
   /**
@@ -149,12 +148,14 @@ export default class SupabaseClient {
       try {
         await this._closeSubscription(subscription)
 
-        const openSubscriptions = this.getSubscriptions().length
-        if (!openSubscriptions) {
+        const allSubscriptions = this.getSubscriptions()
+        const openSubscriptionsCount = allSubscriptions.filter((chan) => chan.isJoined()).length
+
+        if (!allSubscriptions.length) {
           const { error } = await this.realtime.disconnect()
           if (error) return resolve({ error })
         }
-        return resolve({ error: null, data: { openSubscriptions } })
+        return resolve({ error: null, data: { openSubscriptions: openSubscriptionsCount } })
       } catch (error) {
         return resolve({ error })
       }
@@ -165,6 +166,11 @@ export default class SupabaseClient {
     if (!subscription.isClosed()) {
       await this._closeChannel(subscription)
     }
+
+    return new Promise((resolve) => {
+      this.realtime.remove(subscription)
+      return resolve(true)
+    })
   }
 
   /**
@@ -225,7 +231,6 @@ export default class SupabaseClient {
       subscription
         .unsubscribe()
         .receive('ok', () => {
-          this.realtime.remove(subscription)
           return resolve(true)
         })
         .receive('error', (e: Error) => reject(e))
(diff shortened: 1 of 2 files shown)


GaryAustin1:
@w3b6x9 I believe I finally learned how to clone a github branch and build it and move it to my project (all in phpstorm).
If I was successful (75% chance) What I see now is when a realtime subscription is generated with .on other subscriptions are still generated as "closed", BUT now when I call removeAllSubscriptions() the subscription array comes back empty and the socket disconnects as expected.  I still don't know why you want all these extra closed connections hanging around, and I did not test if I just shut down my single subscription would it close down.  THANKS

src/SupabaseClient.ts
```diff
@@ -131,12 +131,11 @@ export default class SupabaseClient {
...
-   * Remove all active subscriptions.
+   * Remove all subscriptions.
    */
-  removeAllSubscriptions() {
-    this.realtime.channels.forEach((sub) => {
-      this.removeSubscription(sub)
-    })
+  async removeAllSubscriptions() {
+    const subscriptions = this.realtime.channels.slice()
+    await Promise.allSettled(subscriptions.map((sub) => this.removeSubscription(sub)))
```
kiwicopple: let's return something from is function so that our users can `await` the connections to all be closed - useful for cleanups

kiwicopple (approved):

src/SupabaseClient.ts
w3b6x9: @kiwicopple yep, that was my intention! forgot the `return`

kiwicopple:
It's be a while since I looked at this functionality, but the general logic that we wanted was:

- 1 socket per client. Eg, we shouldn't be generating a socket connection every time the user calls `subscribe()`, we should just reuse the same socket connection, connecting to different channels 
- the socket should start closed, and only open if there is a subscription created
- if the user removes all subscriptions, we should close the socket connection

The idea is to limit the load on the realtime server as much as possible.

> now when I call removeAllSubscriptions() the subscription array comes back empty and the socket disconnects as expected.

This sounds good - I haven't tested, but if that's the result we want

GaryAustin1:
Although you have made removeAllSubscriptions() delete all subscriptions even the "fake" closed ones from .from.select which is a big improvement...
1) If I create a single subscription, then go about using the database with .from.selects I still get a closed subscription for each of those.  When I then use removeSubscription(mysubscription) the channel remains open because the "fake" subscriptions are still there.  
2)  A normal user does not know there are extra subscriptions that have to be dealt with on close.
3)  removeAllSubscriptions is NOT a documented anywhere other than in code

I believe I am running your modified code (at least removeAll works now) versus if I use 1.28.5.

Here are the fake subscriptions created by .from.selects operations NOT .ON operations...

subs

w3b6x9:
@GaryAustin1

> If I create a single subscription, then go about using the database with .from.selects I still get a closed subscription for each of those.

I just tested this and I'm not getting any closed subscriptions. Might they be from a previous test you conducted prior to testing out this PR?

> removeAllSubscriptions is NOT documented anywhere other than in code

I made this [PR](https://github.com/supabase/supabase/pull/4594) to add it to docs.

I'm going to merge this PR in as it's an improvement and then we can continue to iterate.

GaryAustin1:
I agree it is an improvement and won't make things worse.  I acknowledge I'm not 100% I got right code merged and built in my test, but then new removeAll did work, so I think I did.  The extra closed subscriptions are only add AFTER you open a real subscription with .on.  None before that cause an extra subscription.  I'll check again with released code when I can just pull it with NPM.  Thanks for the work!

GaryAustin1:
1.28.6 DOES fix the "fake" open subscription problems!  Guess I need to work more on pulling and building intermediate code.

PR #353, opened 2022-01-27
feat: refactor Realtime subscription methods

w3b6x9:

feature

## What is the new behavior?

when a user signs out, the client with update realtime subscriptions from the user's access token, where claims role is `authenticated`, to the access token where role is `anon`. this will allow signed out users to continue to listen to changes depending on rls policies without missing messages as long as they have permission.

Other improvements:

- client.removeSubscription(subscription) will close and remove a subscription and return a promise of an object with error and the number of open subscriptions remaining:

e.g.

```js
{
  data: { openSubscriptions: 10 },
  error: null | Error
}
```

- client.removeAllSubscriptions() will close and remove all subscriptions and return a promise of an array of objects with the subscription and error:

e.g.

```js
[{
  data: { subscription: Subscription Object },
  error: null | Error
}]
```

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -131,82 +131,65 @@ export default class SupabaseClient {
   }
 
   /**
-   * Remove all subscriptions.
+   * Closes and removes all subscriptions and returns a list of removed
+   * subscriptions and their errors.
    */
   async removeAllSubscriptions(): Promise<
-    (
-      | {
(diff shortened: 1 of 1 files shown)


koba-ninkigumi:
@w3b6x9 
https://github.com/supabase/supabase-js/pull/360

soedirgo (approved):

PR #369, opened 2022-02-14
fix: update version in package.json on release

w3b6x9:


burggraf (approved):

soedirgo:
@w3b6x9 any reason for doing this? The version in package.json is updated automatically before being pushed to npm.

w3b6x9:
@soedirgo yea, i'm trying to get the correct version in `src/lib/version.ts` so that it can be passed as headers: https://github.com/supabase/supabase-js/blob/93550120f0e89e9dd3c79e966402a3f412c8b120/src/lib/constants.ts#L3

darora:
That gets generated correctly upon release - it's only not going to be set if you're working within a local copy

PR #392, opened 2022-03-01
feat: support throwing errors by default

darora:


soedirgo:
This is just adding the option right? Not for changing the default behavior?

darora:
Yes, I believe the default included[1] ensures that we're not changing the default behaviour.

[1]: https://github.com/supabase/supabase-js/pull/392/files#diff-4fdf43398676cfbb61729753f6419fc2f0996edce995b37200a9cb693a8c4577R78

soedirgo (approved):

PR #399, opened 2022-03-19
fix: allow overriding urls

DavraYoung:

This PR allows developers to control services urls, in case if they use Supabase behind the proxy.

Current behavior appends /rest/v1, /realtime/v1, etc. paths to the url, which sometimes is not desired, and there is not option to easily disable this behavior.

## What is the new behavior?
New behavior by default works as old one, but if user specifies overrideUrls option, he will be able to change "/*/v1" url append logic

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -67,10 +67,22 @@ export default class SupabaseClient {
     const _supabaseUrl = stripTrailingSlash(supabaseUrl)
     const settings = { ...DEFAULT_OPTIONS, ...options }
 
-    this.restUrl = `${_supabaseUrl}/rest/v1`
-    this.realtimeUrl = `${_supabaseUrl}/realtime/v1`.replace('http', 'ws')
-    this.authUrl = `${_supabaseUrl}/auth/v1`
-    this.storageUrl = `${_supabaseUrl}/storage/v1`
+    this.restUrl =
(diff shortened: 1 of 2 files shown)


marcinkoziej:
@DavraYoung have you tested this? You check if option is null, but when unset, it looks like it will be undefined. hence, when none of these options are given, the urls will be set to undefined..

src/SupabaseClient.ts
```diff
@@ -67,10 +67,22 @@ export default class SupabaseClient {
     const _supabaseUrl = stripTrailingSlash(supabaseUrl)
     const settings = { ...DEFAULT_OPTIONS, ...options }
 
-    this.restUrl = `${_supabaseUrl}/rest/v1`
-    this.realtimeUrl = `${_supabaseUrl}/realtime/v1`.replace('http', 'ws')
-    this.authUrl = `${_supabaseUrl}/auth/v1`
-    this.storageUrl = `${_supabaseUrl}/storage/v1`
+    this.restUrl =
```
fdemir, line 70: it would be better if you use "logical or" operator here.

marcinkoziej: https://github.com/supabase/supabase-js/pull/443

soedirgo:
Closed as per https://github.com/supabase/supabase-js/pull/443#issuecomment-1260660413.

PR #402, opened 2022-03-22
fix: allow overriding Bearer token

soedirgo:
When initializing SupabaseClient, if the `Authorization` header is provided, it will be used for `_getAuthHeaders()` instead of the session's access token or Supabase key.

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -247,7 +247,7 @@ export default class SupabaseClient {
     const headers: GenericObject = this.headers
     const authBearer = this.auth.session()?.access_token ?? this.supabaseKey
     headers['apikey'] = this.supabaseKey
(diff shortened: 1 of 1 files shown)


thorwebdev (approved):
LGTM, thanks

maxkostow:
This causes the `Authorization` header to be cached even after `auth` refreshes the `access_token`. This should probably clone the headers object before assigning the properties.

```js
  private _getAuthHeaders(): GenericObject {
    const headers: GenericObject = { ...this.headers }
    const authBearer = this.auth.session()?.access_token ?? this.supabaseKey
    headers['apikey'] = this.supabaseKey
    headers['Authorization'] = headers['Authorization'] || `Bearer ${authBearer}`
    return headers
  }
```

PR #415, opened 2022-03-28
feat(auth): add cookieOptions for underliying auth client

fernandolguevara:
allow customize cookie options on auth client

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -229,6 +229,8 @@ export default class SupabaseClient {
     localStorage,
     headers,
     fetch,
+    cookieOptions,
+    multiTab,
   }: SupabaseClientOptions) {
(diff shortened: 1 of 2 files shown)


fabianludwig:
Thank you! I was just hitting this issue myself. 

An example ro illustrate the benefit of these changes to others: 
I'm working with supabase auth and next.js. The client-side session is stored in local storage and is valid for up to a week. The session cookies lifetime on the other hand is always eight hours. The server-side auth validation is therefore "out of sync" after eight hours and will prompt the user to log in again, even though the session is still valid. 

I expected the cookie to be "synced" with the local storage session.

fernandolguevara:
@soedirgo can we merge this PR?

kangmingtay (approved):

PR #455, opened 2022-06-07
fix: include apikey in custom fetch

alaister:
fix: include apikey in custom fetch

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -23,6 +23,7 @@ export const resolveHeadersConstructor = () => {
 }
 
 export const fetchWithAuth = (
+  supabaseAnonKey: string,
   getAccessToken: () => Promise<string | null>,
   customFetch?: Fetch
 ): Fetch => {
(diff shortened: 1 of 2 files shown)


src/lib/fetch.ts
```diff
@@ -23,17 +23,22 @@ export const resolveHeadersConstructor = () => {
...
   const HeadersConstructor = resolveHeadersConstructor()
 
   return async (input, init) => {
-    const accessToken = await getAccessToken()
+    const accessToken = (await getAccessToken()) ?? supabaseAnonKey
     let headers = new HeadersConstructor(init?.headers)
 
-    if (!headers.has('Authorization') && accessToken) {
+    if (!headers.has('apikey')) {
+      headers.append('apikey', supabaseAnonKey)
```
soedirgo: Should this be `.set` instead of `.append`? Also we can set this unconditionally since this is only used for Kong

alaister: Nice catch - I'll blame copilot for that one ;)

soedirgo (approved):

src/lib/fetch.ts
alaister: I think the conditional is nice in case you're using the the anon key and override with the service role

PR #458, opened 2022-06-08
Revamp Supabase Constructor for V2

J0:

Alters SupabaseClient constructor and types 
Options all around

## What is the new behavior?

Splits Supabase constructor to use client library specific  scoped options

## Additional context

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -1,5 +1,8 @@
 import { GoTrueClient } from '@supabase/gotrue-js'
 import { RealtimeClientOptions } from '@supabase/realtime-js'
+// TODO(Joel): Validate if this is something we want
+// import { FunctionsOptions } from '@supabase/functions-js'
+// import { StorageOptions } from '@supabase/storage-js'
 
 type GoTrueClientOptions = ConstructorParameters<typeof GoTrueClient>[0]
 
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -13,51 +16,64 @@ export type SupabaseClientOptions = {
...
-  headers?: GenericObject
-  /**
-   * Automatically refreshes the token for logged in users.
-   */
-  autoRefreshToken?: boolean
-  /**
-   * Allows to enable/disable multi-tab/window events
-   */
-  multiTab?: boolean
+  api: {
```
soedirgo: Make this `rest`? API is kinda broad (c.f. API key) and this only affects postgrest-js. cc @inian

J0: Sounds good, will do

inian: Thinking about this a bit more, what about `db` instead?

src/lib/types.ts
```diff
@@ -11,51 +11,55 @@ export type SupabaseClientOptions<SchemaName> = {
...
+  }
 
+  auth?: {
+    /**
+     * Automatically refreshes the token for logged in users.
+     */
+    autoRefreshToken?: boolean
+    /**
+     * Allows to enable/disable multi-tab/window /
+    multiTab?: boolean
```
soedirgo: delete?

soedirgo (approved):

PR #460, opened 2022-06-16
fix: namespace storage key

### LINKED ISSUE #332: Namespace cookie / localstorage
There is a conflict if we use multiple supabase projects in the same website. This is because we use the same key to read from localstorage / cookie. We will need to name space the token per project.

kangmingtay:
* Fixes #332 
* See gotrue-js PR for reference (https://github.com/supabase/gotrue-js/pull/303)


sannajammeh:
Great concept! 

However, this could make the storage key a guessing game as its being inferred and as such might introduce a breaking change for unaware users (pulling from localStorage manually).
I'd recommend leaving it up to the user to define it in `createClient()`.

```ts
createClient(URL, KEY, {
   // Options
   storageKey: "myproject-sb" // or authStorageKey
})
```

kangmingtay:
hey @sannajammeh, it wouldn't be a guessing game because the developer would know what the storage key is based on the project-ref from the supabase url provider when the client is initialised

pixtron:
To me that PR looks good. But - as @sannajammeh says - it is a breaking change. Strictly following semver, this PR would require the release of supabase-js v2.0.0 (breaking change -> new major). 

Until v2.0.0 #332 could be partially fixed by letting consumers pass in a custom storageKey to `createClient` and pass it down to gotrue-js. Basically the same as @kangmingtay already did in supabase/gotrue-js/pull/303. So consumers that have multiple clients on the same domain, can set unique keys for them.

This current PR could always be added in next major and serve as a default, when there is no storageKey set in the `createClient` options. I'd really like to see namespaced storage keys!

Also an update to a version containing this PR would log out every user session as soon as users get the new code (new key !== old key). And there would eventually also be garbage sessions left in storage under the old key. 
This could be solved with a migration. 1. reading session from the old key, 2. if the session is a session for the current authUrl store it under the new key and remove the old key. Although such a migration would be probably a bit over engineered.

alaister (approved):
LGTM! 🔥

kangmingtay:
Hey @pixtron and @sannajammeh, thanks for the feedback! I've made the necessary changes so you'll be able to pass in a custom storage key in the next major version release.

PR #468, opened 2022-06-29
Fix: Realtime setup with GoTrue v2

alaister:

Fix

## What is the current behaviour?

The current setup may not set an access token until the token is refreshed. This is because of a race condition where if the initial `SIGNED_IN` event occurs while the channel is in an `isJoining` state, it won't send the access token.

## What is the new behaviour?

This setup sets the join token if we have it, and always sets the access token after the channel has joined successfully.

## Additional context

It's worth mentioning in the docs that there may (rarely) be a case where realtime connects with the anon token even when the user is logged in before it sends up the current users token.


src/SupabaseClient.ts
```diff
@@ -97,9 +97,9 @@ export default class SupabaseClient<
     this.headers = { ...DEFAULT_HEADERS, ...options?.headers }
     this.shouldThrowOnError = settings.shouldThrowOnError || false
 
+    this.auth = this._initSupabaseAuthClient(settings.auth || {}, this.headers, settings.fetch)
     this.fetch = fetchWithAuth(supabaseKey, this._getAccessToken.bind(this), settings.fetch)
 
-    this.auth = this._initSupabaseAuthClient(settings.auth || {}, this.headers, this.fetch)
```
alaister, line 102: @soedirgo fyi this was a circular reference. Is there any reason why we can't use the regular `settings.fetch` here instead of our `fetchWithAuth`?

soedirgo: looks like the circular reference is `this.auth` -> `this.fetch` -> `this._getAccessToken` -> `this.auth`

not quite familiar with gotrue-js but I can't think of any reason not to use `settings.fetch` off the top of my head (but we def can't use it for `functions` and `storage`)

alaister: Correct. It's only if we were doing something extra in `fetchWithAuth` that `gotrue-js` needed... but yeah it just does a normal fetch so thanks, Bobbie!

src/lib/SupabaseRealtimeClient.ts
```diff
@@ -66,13 +62,30 @@ export class SupabaseRealtimeClient {
    * Enables the channel.
    */
   subscribe(callback: Function = () => {}) {
+    // if the socket already has a good accessToken
+    // we can just use it strait away∏
```
w3b6x9:
```suggestion
    // we can just use it straight away
```

w3b6x9: @alaister did you intend to do `∏`? what even is that?

w3b6x9 (approved):
@alaister realtime changes lgtm!

src/lib/SupabaseRealtimeClient.ts
alaister: Nice catch! It looks like I've been button-mashing again...

PR #485, opened 2022-08-04
fix: bump gotrue-js/next version

kangmingtay:


alaister (changes requested):
I think we'll need to update the `supabase-js` code that uses the now removed methods?

alaister (approved):
lgtm!
Can we also remove the `isBrowser()`? Not sure if it's still used anywhere?

alaister (changes requested):
Oh also looks like the tests need changing to `getSession()` as well

kangmingtay:
@alaister i think we can just remove them since they are using `setAuth` ?

alaister (approved):
@kangmingtay yep I agree – the use case is no longer relevant.

PR #490, opened 2022-08-10
temporarily remove shouldThrowOnError

inian:
Till all the client libraries support this option.

Once they do, we can add this option back in to the parent library as a non-breaking change

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -51,7 +51,6 @@ export default class SupabaseClient<
   protected storageKey: string
   protected fetch?: Fetch
   protected changedAccessToken: string | undefined
-  protected shouldThrowOnError: boolean
 
   protected headers: {
     [key: string]: string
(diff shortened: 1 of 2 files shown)


soedirgo (approved):

logemann:
any news when this is back in ?

Revadike:
@inian @soedirgo any plans on bringing this back?

soedirgo:
We're still planning on adding it back, but no timeline for now

Magnuti:
Any updates on this? It is super annoying having to write error handling logic for every query

```ts
const { data, error } = await supabase.from("...")

if (error) {
  ...
}
```

I would much rather like the function to just throw. The `shouldThrowOnError` option worked perfectly before.

Magnuti:
If anyone is looking for a similar solution to this just use `. throwOnError()` https://github.com/orgs/supabase/discussions/604

PR #492, opened 2022-08-10
Patch revamped constructor

J0:

Draws heavily on the work done in #499 

Here's an image of a manual test:

input options:
CleanShot 2022-08-13 at 18 40 37@2x

Sample output:

CleanShot 2022-08-13 at 18 39 15@2x

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -19,7 +19,6 @@ const DEFAULT_OPTIONS = {
   autoRefreshToken: true,
   persistSession: true,
   detectSessionInUrl: true,
-  multiTab: true,
   headers: DEFAULT_HEADERS,
 }
 
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -9,7 +9,7 @@ import { RealtimeChannel, RealtimeClient, RealtimeClientOptions } from '@supabas
 import { StorageClient as SupabaseStorageClient } from '@supabase/storage-js'
 import { DEFAULT_HEADERS } from './lib/constants'
 import { fetchWithAuth } from './lib/fetch'
-import { stripTrailingSlash } from './lib/helpers'
+import { stripTrailingSlash, applySettingDefaults } from './lib/helpers'
 import { SupabaseAuthClient } from './lib/SupabaseAuthClient'
 import { SupabaseRealtimeChannel } from './lib/SupabaseRealtimeChannel'
 import { Fetch, GenericSchema, SupabaseClientOptions, SupabaseAuthClientOptions } from './lib/types'
@@ -105,7 +105,7 @@ export default class SupabaseClient<
       global: DEFAULT_GLOBAL_OPTIONS,
     }
 
-    const settings = this._applySettingDefaults(options || {}, DEFAULTS)
+    const settings = applySettingDefaults(options || {}, DEFAULTS)
 
     this.storageKey = settings.auth?.storageKey ?? ''
     this.headers = settings.global?.headers ?? {}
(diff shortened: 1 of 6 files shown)


src/SupabaseClient.ts
```diff
@@ -94,20 +95,23 @@ export default class SupabaseClient<
...
+    )
     this.fetch = fetchWithAuth(supabaseKey, this._getAccessToken.bind(this), settings.fetch)
 
     this.realtime = this._initRealtimeClient({ headers: this.headers, ...settings.realtime })
     this.rest = new PostgrestClient(`${_supabaseUrl}/rest/v1`, {
       headers: this.headers,
       schema: options?.db?.schema,
       fetch: this.fetch,
-      throwOnError: this.shouldThrowOnError,
+      throwOnError: options?.db?.shouldThrowOnError || false,
```
inian: options?.db?.shouldThrowOnError ?? false

Here it is okay cos the second argument is false but better to use ??

src/SupabaseClient.ts
```diff
@@ -92,18 +99,22 @@ export default class SupabaseClient<
     const defaultStorageKey = `sb-${new URL(this.authUrl).hostname.split('.')[0]}-auth-token`
     this.storageKey = options?.auth?.storageKey ?? defaultStorageKey
 
-    const settings = { ...DEFAULT_OPTIONS, ...options, storageKey: this.storageKey }
+    const settings = Object.assign(DEFAULT_OPTIONS, options)
```
pixtron: I just have those concerns:

1. If `options.auth.storageKey` is not set, the `defaultStorageKey` seems to not be applied
2. By using `Object.assign()` `DEFAULT_OPTIONS` gets changed, if that should not be the case it is safer to use `Object.assign({}, DEFAULT_OPTIONS, options)`  which creates a new object and assigns to it.
3. `Object.assign()` only does a shallow merge, eg if `options.auth` is set it will completely replace `DEFAULT_OPTIONS.auth`

J0: Hey @pixtron,

Thanks for raising these points!  You are right, the previous draft implementation which makes use of `Object.assign` doesn't work  as `Object.assign` does a shallow copy. This was flagged two days back but we hadn't gotten around to fixing it.

I've made some amendments by drawing heavily on the changes you've suggested in the branch and will tag you for review once I clean up and write tests.

Once again, thanks so much for the feedback!

pixtron:
LGTM. Just the docs aren't yet up to date. Probably better to change them in an own PR though.

https://github.com/supabase/supabase-js/blob/72b7ff3c6a361bb95ca02cffe85569dbef84c209/src/SupabaseClient.ts#L70-L76

src/SupabaseClient.ts
```diff
@@ -90,19 +98,29 @@ export default class SupabaseClient<
...
-    this.storageKey = options?.auth?.storageKey ?? defaultStorageKey
+    const DEFAULTS = {
+      db: DEFAULT_DB_OPTIONS,
+      realtime: DEFAULT_REALTIME_OPTIONS,
+      auth: { ...DEFAULT_AUTH_OPTIONS, storageKey: defaultStorageKey },
+      global: DEFAULT_GLOBAL_OPTIONS,
+    }
 
-    const settings = { ...DEFAULT_OPTIONS, ...options, storageKey: this.storageKey }
+    const settings = applySettingDefaults(options || {}, DEFAULTS)
```
inian: `options ?? {}`

src/SupabaseClient.ts
```diff
@@ -90,19 +98,29 @@ export default class SupabaseClient<
...
 
-    this.headers = { ...DEFAULT_HEADERS, ...options?.headers }
+    this.storageKey = settings.auth?.storageKey ?? ''
```
inian, line 110: will storageKey ever be nullish here since we are setting it to the defaultStorageKey above?

src/SupabaseClient.ts
```diff
@@ -90,19 +98,29 @@ export default class SupabaseClient<
...
+    this.headers = settings.global?.headers ?? {}
```
inian, line 111: will headers be nullish here since we set DEFAULT_HEADERS above?

src/SupabaseClient.ts
```diff
@@ -90,19 +98,29 @@ export default class SupabaseClient<
...
 
-    this.auth = this._initSupabaseAuthClient(settings.auth || {}, this.headers, settings.fetch)
-    this.fetch = fetchWithAuth(supabaseKey, this._getAccessToken.bind(this), settings.fetch)
+    this.auth = this._initSupabaseAuthClient(
+      settings?.auth || {},
```
inian: will settings.auth be nullish here since we set DEFAULT_AUTH_OPTIONS above?

inian (approved):

alaister (approved):

src/SupabaseClient.ts
J0: Nope,  you're right that it shouldn't be Nullish. However, I think the compiler thinks it might be Null-ish because `applySettingDefaults` returns `SupabaseClientOptions` which allows for null-ish storage key.

This is probably something that should be fixed in the typing (e.g. via a new type or other), will leave it there and merge for now(so as not to block other changes) but am open to suggestions on how we can best fix.

src/SupabaseClient.ts
J0: Same as above

J0:
Thanks everyone, merging for now so as to unblock the follow up PRs but if any changes come to mind feel free to comment.

PR #499, opened 2022-08-12
fix: respect default options

pixtron:

Bug fix

After #458, default options of supabase-js does not have any effect on auth anymore as they are expected in `settings.auth` now.  `auth` is undefined in `DEFAULT_OPTIONS`, although default options specifies default auth settings like `autoRefreshToken`. 

## What is the new behavior?

Properly merges default options and user options. Although a testable utility method would probably be superior.
A shallow copy of default options and user options is not sufficient as a partial auth setting passed by the user would override the default options not set in the user options. 

```js
const settings = {
  ...{auth: {a: true}}
  ...{auth: {b: true}}
} 

--> {auth: {b: true}}
```

## Additional context

There seems to be another bug (or is that intentional?) in `_initSupabaseAuthClient`, overriding default gotrue-js options with undefined.

For example when calling `this._initSupabaseAuthClient({persistSession: false});` GotrueClient will be constructed with these options:

```js
{
  persistSession: false,
  storageKey: undefined,
  autoRefreshToken: undefined,
  detectSessionInUrl: undefined,
  storage: undefined,
  cookieOptions: undefined,
 }
```
These undefined values then override the gotrue-js default options
```js
const DEFAULT_OPTIONS = {autoRefreshToken: true}
const settings = {...DEFAULT_OPTIONS, {autoRefreshToken: undefined, persistSession: false}}
--> {autoRefreshToken: undefined, persistSession: false}
```

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -14,12 +14,20 @@ import { SupabaseAuthClient } from './lib/SupabaseAuthClient'
 import { SupabaseRealtimeChannel } from './lib/SupabaseRealtimeChannel'
 import { Fetch, GenericSchema, SupabaseClientOptions, SupabaseAuthClientOptions } from './lib/types'
 
-const DEFAULT_OPTIONS = {
+const DEFAULT_OPTIONS: SupabaseClientOptions<'public'> = {
+  headers: DEFAULT_HEADERS,
+}
+
+const DEFAULT_DB_OPTIONS = {
   schema: 'public',
+}
+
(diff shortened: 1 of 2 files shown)


J0:
Hey @pixtron,

Thanks for taking the time to look into this!

This is a known issue and we're actually looking into patching it in #492

I'll take the suggestions here into account when updating the PR.

Thanks!
Joel

pixtron:
Great you are already aware and working on a fix. Closing this PR.

PR #555, opened 2022-09-21
Update api.js

Sujaruu:
adding validation if (.env) missing out


Sujaruu:
adding validation if (.env) missing out

soedirgo (approved):

PR #584, opened 2022-10-10
fix: custom authorization header should not be overwritten

kangmingtay:
* Fixes a bug with `_initSupabaseAuthClient` where it overwrites the custom `Authorization` header passed into `SupabaseClient` with the default authorization header (the supabase key)
* The auth client should be using the custom `Authorizaion` header instead


thorwebdev (approved):
LGTM, thanks

PR #672, opened 2022-12-30
fix: unset `changedAccessToken` on sign out

hf:
When a user signs out, the access token should no longer be kept in RAM.

See: https://github.com/supabase/gotrue-js/issues/524

Thanks @j4w8n.


j4w8n:
@hf haha, I already had https://github.com/supabase/supabase-js/pull/645 for this.

hf:
> @hf haha, I already had https://github.com/supabase/supabase-js/pull/645 for this.

Oh sorry wasn't aware but also didn't check 😅. Thanks let's go with yours!

PR #704, opened 2023-02-04
Fix - Set custom jwt as auth token

itsgouthamraj:

Bug fix - Initializing realtime client failing when custom JWT sent through Authorization header 
Please find the issue [here](https://github.com/supabase/realtime/issues/470)

If the custom JWT is sent through global Authorization Header then this custom JWT is used as apikey while making new realtime connection. This behaviour is added in this [pr](https://github.com/supabase/supabase-js/pull/699)

## What is the new behavior?

If the custom JWT is sent through headers, we create new realtime client using supabasekey as apikey and then we use setAuth function to set the custom JWT

## Additional context

Please find this [issue](https://github.com/supabase/realtime/issues/470) created in realtime repo

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -274,6 +274,7 @@ export default class SupabaseClient<
     if ( authToken ) {
       newRealtimeClient.setAuth(authToken)
     }
+    return newRealtimeClient
   }
 
   private _listenForAuthEvents() {
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -267,10 +267,13 @@ export default class SupabaseClient<
   private _initRealtimeClient(options: RealtimeClientOptions) {
     const authHeader = options.headers?.Authorization
     const authToken = authHeader?.startsWith('Bearer ') && authHeader.split(' ')[1]
-    return new RealtimeClient(this.realtimeUrl, {
+    const newRealtimeClient = new RealtimeClient(this.realtimeUrl, {
       ...options,
-      params: { ...{ apikey: authToken || this.supabaseKey }, ...options?.params },
+      params: { ...{ apikey: this.supabaseKey }, ...options?.params },
     })
+    if ( authToken ) {
+      newRealtimeClient.setAuth(authToken)
+    }
   }
 
   private _listenForAuthEvents() {


itsgouthamraj:
@w3b6x9 Can you please review it?

w3b6x9:
@itsgouthamraj take a look at https://supabase.com/docs/guides/realtime/postgres-changes#custom-tokens.

Also, `setAuth` only pushes the token to the channels on the server if the channels are already connected: https://github.com/supabase/realtime-js/blob/63242ff54958ecedb924345d3c07ffbff69cf830/src/RealtimeClient.ts#L279. I'm actually surprised this worked for you since you can only create and connect to channels after creating the Supabase client.

gouthamraj-r:
Thanks @w3b6x9 . In my code, I'm calling setAuth only after the realtime connection is created so its worked for me.

itsgouthamraj:
> @itsgouthamraj take a look at https://supabase.com/docs/guides/realtime/postgres-changes#custom-tokens.
> 
> Also, `setAuth` only pushes the token to the channels on the server if the channels are already connected: https://github.com/supabase/realtime-js/blob/63242ff54958ecedb924345d3c07ffbff69cf830/src/RealtimeClient.ts#L279. I'm actually surprised this worked for you since you can only create and connect to channels after creating the Supabase client.

@w3b6x9 It doesn't fix the issue. Also about `setAuth`, https://github.com/supabase/realtime-js/blob/63242ff54958ecedb924345d3c07ffbff69cf830/src/RealtimeClient.ts#L274 you can see here that the token sent to `setAuth` is assigned to `this.accessToken` so it not only updates the existing channels but also all the new channels created using this supabase client instance will be created with updated token only. Am I Right? I think this might be reason my code is working.

w3b6x9:
@itsgouthamraj yes t's setting `this.accessToken` and then on the next heartbeat it'll call `setAuth` which will send it to the server when there are channels subscribed during that window: https://github.com/supabase/realtime-js/blob/4bae15bee7aa3e951a169b772d0a8fc606e4e920/src/RealtimeClient.ts#L404.

itsgouthamraj:
@w3b6x9 Got it. Also https://supabase.com/docs/guides/realtime/postgres-changes#custom-tokens, this doesn't fix this. In fact, it throws a web-socket connection error when we use any key other than the anon key. Since `setAuth` is not exposed as an API, I cannot use my current workaround without modifying the superbase-js source code. Can we get this PR merged? Or do we have any work around here?

itsgouthamraj:
Hey supabase team, any updates on this?

w3b6x9:
@itsgouthamraj I'll address it here: https://github.com/supabase/realtime/issues/470.

PR #720, opened 2023-03-04
Bump GoTrue-js to v2.16.0

J0:

Bumps GoTrue-js version to v2.16.0 which includes WhatsApp release.


soedirgo (approved):
Lots of checksums are deleted in the package-lock.json, is that expected?

J0:
Hmm, not too sure - I deleted `node_modules` and re-installed and the `package-lock.json` seems to have the `shasum`s reinstated

PR #748, opened 2023-04-10
feat: bump gotrue to v2.20

J0:
`USER_DELETED` event was removed as it was deprecated, additional option `flow_type` passed in

Relevant PRs: 
User deleted event removed: https://github.com/supabase/gotrue-js/pull/653

PKCE settting: https://github.com/supabase/gotrue-js/pull/646

Note, we should also check that the latest changes on GoTrue-js don't break anything before merging, seems like build is failing. Putting this out here to solicit early feedback

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -30,6 +30,7 @@ const DEFAULT_AUTH_OPTIONS: SupabaseAuthClientOptions = {
   autoRefreshToken: true,
   persistSession: true,
   detectSessionInUrl: true,
+  flowType: 'implicit',
 }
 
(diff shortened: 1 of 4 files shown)


src/SupabaseClient.ts
```diff
@@ -291,7 +294,7 @@ export default class SupabaseClient<
       this.realtime.setAuth(token ?? null)
 
       this.changedAccessToken = token
-    } else if (event === 'SIGNED_OUT' || event === 'USER_DELETED') {
```
silentworks, line 294: This is potentially a breaking change if someone was relying on this event.

PR #778, opened 2023-05-26
feat: edge functions support for custom domains and vanity domains

delgado3d:

feature

## What is the new behavior?

Invoke functions with `/functions/v1` path prefix allowing us to support custom domains and vanity domains for edge functions


thorwebdev (approved):
LGTM, thanks 👍

PR #792, opened 2023-06-13
chore: type for optional params

Ananya2001-an:

just used a better way to define optional params instead of writing `undefined` explicitly


soedirgo (approved):
Looks good, thanks!

Ananya2001-an:
Awesome...though a test is failing is it, because of my change? Probably not I hope :)

soedirgo:
I don't think so - retrying

soedirgo:
Looks good now, thanks again!

Ananya2001-an:
Happy to help :)

PR #793, opened 2023-06-13
ci: added paths-ignore option

### LINKED ISSUE #791: ci: add `paths-ignore` to ci GHA
# Chore

## Describe the chore
 
Running tests when making changes to .md files or maybe .ignore files can be a bit unnecessary so I guess adding something like below would be good:

```yaml
on:
  push:
    branches:
      - master
      - next
      - rc
    paths-ignore:
      - *.md
      - .gitignore
```

I would like to work on it if it's approved :)

Ananya2001-an:

Added paths-ignore option for running ci tests only when certain files change

closes #791

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,21 @@ name: CI
 
 on:
   pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '**/*.md'
+      - '.prettierrc'
+      - '**/*ignore'
   push:
     branches:
       - master
       - next
       - rc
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,11 +2,25 @@ name: CI
 
 on:
   pull_request:
+    branches:
+      - master
+      - next
+      - rc
```
soedirgo: Let's keep it applied to all branches? Still useful to run the CI in these cases

soedirgo (approved):

PR #828, opened 2023-08-05
Add dynamic schemas on a per-query basis

sjones6:
Resolves https://github.com/supabase/postgrest-js/issues/280

Feature

Currently, the supabase client is created and defaults to the `public` schema when no option is supplied:

```ts
createClient(supabaseUrl, supabaseKey)
```

Or, you can supply an alternative schema:

```ts
createClient(supabaseUrl, supabaseKey, {
  db: {
    schema: 'alternative'
  }
})
```

However, this means that all methods/queries for the client are associated to the single schema. Using multiple schemas requires multiple clients, which presents a challenge for auth where all clients need to be authenticated.

See pre-existing issue:

https://github.com/supabase/postgrest-js/issues/280

## What is the new behavior?

Extends on behavior added to underlying `PostgrestClient`: https://github.com/supabase/postgrest-js/pull/455

The method added to the supabase client allows for dynamic schema selection on a per-query basis.

```ts
const supabase = createClient(supabaseUrl, supabaseKey);

supabase.from('some_table_in_public').select();

supabase.schema('alt').from('table_in_alt').select()
```

If a `Database` generic type is supplied, type inference works as expected without additional generics:
Screenshot 2023-08-06 at 1 12 31 PM

## Additional context

### DIFF
--- a/test/types.ts
+++ b/test/types.ts
@@ -1,0 +1,201 @@
+export type Json = string | number | boolean | null | { [key: string]: Json } | Json[]
+
+export interface Database {
+  personal: {
+    Tables: {
+      users: {
+        Row: {
+          age_range: unknown | null
+          data: Json | null
+          status: Database['public']['Enums']['user_status'] | null
+          username: string
+        }
+        Insert: {
+          age_range?: unknown | null
+          data?: Json | null
+          status?: Database['public']['Enums']['user_status'] | null
+          username: string
+        }
+        Update: {
+          age_range?: unknown | null
+          data?: Json | null
+          status?: Database['public']['Enums']['user_status'] | null
+          username?: string
+        }
+        Relationships: []
+      }
+    }
+    Views: {
+      [_ in never]: never
+    }
+    Functions: {
+      get_status: {
+        Args: {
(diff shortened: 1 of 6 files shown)


dshukertjr:
Thanks for the feedback! We do have an issue discussing a solution for this here https://github.com/supabase/postgrest-js/issues/280 . We will pick this one post launch week. Thanks for your patience!

soedirgo:
Thanks @sjones6! We'll need the change both here and in postgrest-js, so keeping this open.

sjones6:
@soedirgo , updated this PR now that https://github.com/supabase/postgrest-js/pull/455 is published.

test/types.ts
```diff
@@ -0,0 +1,201 @@
+export type Json = string | number | boolean | null | { [key: string]: Json } | Json[]
+
+export interface Database {
```
sjones6, line 3: Copied from postgrest-js. Didn't bring over the the whole database setup that postgrest-js has, since that seems like a lot for this change.

soedirgo (approved):

Issue #612, opened 2022-10-26
Can't resolve 'encoding'


thorwebdev:
Module not found: Can't resolve 'encoding' in '/Users/thorwebdev/Documents/code/supabase/quickstarts/supabase-next-13/node_modules/.pnpm/node-fetch@2.6.7/node_modules/node-fetch/lib'

This warning is printed when using supabase-js in nextjs 13 project: https://github.com/supabase/auth-helpers/issues/340

It's related to an old version of node-fetch being used in cross-fetch: https://github.com/node-fetch/node-fetch/issues/412  https://github.com/lquixada/cross-fetch/issues/146

It should be safe to ignore as it's an optional dependency in node-fetch, but might be worthwhile to see if we can get rid of the warning.

boompikachu:
Is there a way to silent the warning?

cdedreuille:
I'm seeing the same issue.

```
Module not found: Can't resolve 'encoding' in '/Users/cdedreuille/Sites/kapture/node_modules/node-fetch/lib'

Import trace for requested module:
./node_modules/node-fetch/lib/index.js
./node_modules/cross-fetch/dist/node-ponyfill.js
./node_modules/@supabase/supabase-js/dist/main/lib/fetch.js
./node_modules/@supabase/supabase-js/dist/main/SupabaseClient.js
./node_modules/@supabase/supabase-js/dist/main/index.js
./lib/supabase.ts
./app/app/page.tsx

./node_modules/node-fetch/lib/index.js
Module not found: Can't resolve 'encoding' in '/Users/cdedreuille/Sites/kapture/node_modules/node-fetch/lib'

Import trace for requested module:
./node_modules/node-fetch/lib/index.js
./node_modules/cross-fetch/dist/node-ponyfill.js
./node_modules/@supabase/supabase-js/dist/main/lib/fetch.js
./node_modules/@supabase/supabase-js/dist/main/SupabaseClient.js
./node_modules/@supabase/supabase-js/dist/main/index.js
./lib/supabase.ts
./app/app/components/image.tsx
./app/app/components/grid.tsx
```

Everything is working as expected but it would be great to make it go away.

SharadKumar:
Yes, am getting the same with the latest Nextjs 13.1.6. Everything works though.

kamto7:
I am using node 18+ version, It does not need `cross-fetch`.

```shell
pnpm add -D encoding
```
With this you can run.

cdedreuille:
[...]

PR #850, opened 2023-09-11
fix: replace cross-fetch with @supabase/node-fetch

soedirgo:
Fixes https://github.com/supabase/supabase-js/issues/612.


kevcodez (approved):

soedirgo:
Running `npm i @supabase/supabase-js@2.33.2` or any version after that should do it I believe.

PR #855, opened 2023-09-19
feat: bump `@supabase/gotrue-js` to v2.53.0

hf:
Bumps `@supabase/gotrue-js` to v2.53.0.


J0 (approved):

J0:
THink we still need to update the types on the auth.debug flag to fix the CI. Otherwise should be good

PR #933, opened 2023-12-22
feat: switch to `@supabase/auth-js`

hf:
Switches to using [@supabase/auth-js]() instead of @supabase/gotrue-js. Note that these two packages are **equvalent** and **guaranteed to remain equivalent for as long as version 2 is maintained.**


soedirgo (approved):

hf:
Hold this, realized that people with pinned gotrue-js versions will probably have an issue with this.

PR #935, opened 2024-01-02
fix: stop using unstable type params from postgrest-js

soedirgo:
We're currently duplicating passing the type params to postgrest-js classes - we may change the order and/or add new type params on the postgrest-js side, so the type params in supabase-js can get out of sync.

This also fixes the issue in this [demo](https://github.com/daichi5/supabase-demo/blob/91385f2c1d509109938cb0e02d2a3b00201cf044/src/index.ts#L12) (`users` in `data` should be `T[]`, not `T | null`).


thorwebdev (approved):
lgtm

PR #973, opened 2024-02-13
fix: Use class methods instead of bound arrow functions for rest calls

kamilogorek:
`rest` is already `PostgrestClient<Database, SchemaName>`, so all return types should be infered in the same way as before without using explicit return types.

(I changed `FnName` and `Fn` so it fits on one line, as VSCode is messing up syntax highlighting when generics are split into multiple lines 👀 )

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -130,7 +130,7 @@ export default class SupabaseClient<
    *
    * @param relation - The table or view name to query
    */
-  from: PostgrestClient<Database, SchemaName>['from'] = (relation: string) => {
+  from(relation: string) {
     return this.rest.from(relation)
   }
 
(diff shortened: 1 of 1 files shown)


soedirgo (approved):

benaduggan:
Upgrading to this version blew up the types for me 👀 Am I doing something wrong? or did this PR have unintended consequences? 

Lots of things like this:
image

image

peterje:
> Upgrading to this version blew up the types for me 👀 Am I doing something wrong? or did this PR have unintended consequences?
> 
> Lots of things like this: image
> 
> image

Bump. This change broke my team's CI today.

soedirgo:
Sorry about that! This should be fixed in v2.39.5, can you give it a try?

PR #984, opened 2024-03-01
feat: typecheck table names in from clause

vonovak:

Please take this as a sort of a question because I'm new to supabase. (and I'm aware of the `// NOTE: signatures must be kept in sync with PostgrestClient.from`).

### Motivation

I want to query supabase using the client and want to do that with the best possible guidance from TS.

Right now, TS (although it will suggest valid table names) won't complain if I try to query a non-existent table.
To me, this is a fairly unexpected limitation and I don't understand why it's there.

So I'm suggesting this change here (I can also make it in the other place) with the aim of achieving higher code safety.
If this is not possible, why? (Are there other cases like this, with low-hanging fruit for TS improvements?)

Thank you! :)

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -130,10 +130,11 @@ export default class SupabaseClient<
   }
 
   // NOTE: signatures must be kept in sync with PostgrestClient.from
-  from<TableName extends keyof Schema['Tables'], Table extends Schema['Tables'][TableName]>(
-    relation: TableName
-  ): PostgrestQueryBuilder<Schema, Table, TableName>
-  from<ViewName extends keyof Schema['Views'], View extends Schema['Views'][ViewName]>(
+  from<
+    TableName extends string & keyof Schema['Tables'],
+    Table extends Schema['Tables'][TableName]
+  >(relation: TableName): PostgrestQueryBuilder<Schema, Table, TableName>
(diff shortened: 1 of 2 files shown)


kamilogorek:
Hey, this looks like a good addition. We did this previously because we needed a support for any generic name, due to materialized views not being supported in the generated types. Now that they are, we can update the types just fine.

As for your change, the only required one is the last line, which is removal of the 3rd overload.
We need to use `extends string & keyof` instead of just `extends keyof`, as `symbol` type can also be passed as a type without that restriction, and it'll fail inside `postgrest-js`, where we do string interpolation.

So now to make it land, what we need is:
- change the PR to only include last overload removal
- updated `postgrest-js` to make it inline with this change
- bump `postgrest-js` version here

Thanks for catching this one! (not that I'll be OOO for the next 2 weeks, so can be slow to respond) cc @soedirgo

soedirgo:
We also need to support generating types for foreign tables: https://github.com/supabase/postgres-meta/pull/754

Otherwise, this change looks good to me - will take care of the typegen updates and postgrest-js changes 👍

vonovak:
@soedirgo I made the necessary change 🙂

soedirgo:
Thanks! The typegen support for foreign tables has been deployed, so merging this in 🚀

soedirgo (approved):

PR #998, opened 2024-03-28
fix: bump auth to v2.63.0

kangmingtay:
* Bump auth to v2.63.0 which includes the `signInAnonymously` method


soedirgo (approved):

hawkcookie:
@kangmingtay @soedirgo 
This changes may causes what is pointed out in the following issues.
https://github.com/supabase/supabase-js/issues/1001

PR #1004, opened 2024-04-01
feat: add third-party auth support

hf:
Adds support for the `accessToken` option on the Supabase client which can be used to provide a third-party authentication (e.g. Auth0, Clerk, Firebase Auth, ...) access token or ID token to be used instead of Supabase Auth.

When set, `supabase.auth.xyz` cannot be used and an error will be thrown.

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -51,6 +51,7 @@ export default class SupabaseClient<
   protected storageKey: string
   protected fetch?: Fetch
   protected changedAccessToken?: string
+  protected accessToken?: () => Promise<string>
 
   protected headers: Record<string, string>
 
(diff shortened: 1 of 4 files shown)


hf:
Merge after platform changes are released.

j4w8n:
Correct me if I'm wrong, but couldn't this also be used to authenticate supabase clients, for RLS, during API requests? This assumes a Supabase JWT is being used as the API key.

So instead of adding the JWT to the global header, you'd use `accessToken`, because I believe supabase-js will use this for db fetch requests.
```diff
/* Some API endpoint that you user hits. */
const jwt = 'get-from-request-authorization-header'

const supabase = createClient(
  env.SUPABASE_URL, 
  env.SUPABASE_ANON_KEY, {
+ accessToken: async () => { return `${jwt}` }
- global: {
-   headers: {
-     Authorization: `Bearer ${jwt}`
-   }
- },
- auth: {
-   persistSession: false,
-   detectSessionInUrl: false,
-   autoRefreshToken: false
- }
})

const { data, error } = await supabase.from('table').select('column')
```

mansmahamat (approved):

hf:
> Correct me if I'm wrong, but couldn't this also be used to authenticate supabase clients, for RLS, during API requests? This assumes a Supabase JWT is being used as the API key.
> 
> So instead of adding the JWT to the global header, you'd use `accessToken`, because I believe supabase-js will use this for db fetch requests.

Absolutely. No more needing to patch the `global` option thing.

src/SupabaseClient.ts
```diff
@@ -95,11 +96,24 @@ export default class SupabaseClient<
...
+    } else {
+      this.accessToken = settings.accessToken
+
+      this.auth = new Proxy<SupabaseAuthClient>({} as any, {
+        get: (prop) => {
+          throw new Error(
+            `@supabase/supabase-js: Supabase Client is configured with the accessToken option, accessing supabase.auth.${prop} is not possible`
+          )
+        },
+      })
```
kangmingtay: @hf why would we not want folks to access this if the access token is set?

kangmingtay: is it because we want to prevent supabase auth from overriding the access token set?

hf: No, it is because this overrides any Supabase Auth access token, so using `supabase.auth.,,,` makes no more sense, and this prevents that confusion for now.

What access token will storage receive? Not Supabase Auth's.

PR #1005, opened 2024-04-01
Export GenericSchema types

alexdlm:

Improvement

Database / Schema types can be supplied / inferred.

## What is the new behavior?

Able to extend/reuse the types for schema interfaces.

## Additional context

It's not strictly necessary but I think is friendlier to export all the named types that are referenced in exported types.

I'm making a library that expects a table with a specific schema to be defined, it's nicer to be able to require the `SupabaseClient` that a user passes through has the table with the expected schema defined, plus all usage of the client conforms to the declared interfaces. Also declaring my schema as extends `GenericSchema` catches issues from future type changes in the right place, and also documents to consumers what the schema/database objects are.

It's also nicer for future breaks in the interfaces to cause errors in the types that use them instead of in function calls spread around.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,18 @@ export {
 } from '@supabase/functions-js'
 export * from '@supabase/realtime-js'
 export { default as SupabaseClient } from './SupabaseClient'
-export type { SupabaseClientOptions, QueryResult, QueryData, QueryError } from './lib/types'
+export type {
+  GenericFunction,
+  GenericNonUpdatableView,
+  GenericSchema,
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks for the PR! Unfortunately these are internal types and we don't plan to export them.

You can of course reimplement these types in your library, and things should keep working since the shape of the `Database` type is expected to be backward-compatible as we make changes to it - this way you won't have unexpected changes to the types when you upgrade supabase-js (e.g. we may decide to throw away all these helper types when we do a major refactor).

PR #1044, opened 2024-05-15
fix: publish to jsr.

thorwebdev:
Publish to https://jsr.io/@supabase/supabase-js

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,10 @@
   "module": "dist/module/index.js",
   "types": "dist/module/index.d.ts",
   "sideEffects": false,
-  "repository": "supabase/supabase-js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/supabase/supabase-js.git"
+  },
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -19,7 +19,10 @@
   "module": "dist/module/index.js",
   "types": "dist/module/index.d.ts",
   "sideEffects": false,
-  "repository": "supabase/supabase-js",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/supabase/supabase-js.git"
```
laktek: I'm not sure if anything depended on `repository` previously. If it did we may need to check if the new format doesn't break those.

laktek (approved):

package.json
thorwebdev: Yah, good point. semantic-release should be able to understand the old format. so I'll revert to that. Thanks.

soedirgo (approved):

PR #1045, opened 2024-05-15
feat: allow overriding default PostgREST route prefix

pcnc:

* adds the ability to specify a custom route prefix for a project's PostgREST service

## Additional context

https://supabase.slack.com/archives/C01D6TWFFFW/p1715697051055689

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -17,6 +17,7 @@ import {
   DEFAULT_DB_OPTIONS,
   DEFAULT_AUTH_OPTIONS,
   DEFAULT_REALTIME_OPTIONS,
+  DEFAULT_POSTGREST_OPTIONS,
 } from './lib/constants'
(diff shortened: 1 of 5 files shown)


src/SupabaseClient.ts
```diff
@@ -88,6 +89,7 @@ export default class SupabaseClient<
       realtime: DEFAULT_REALTIME_OPTIONS,
       auth: { ...DEFAULT_AUTH_OPTIONS, storageKey: defaultStorageKey },
       global: DEFAULT_GLOBAL_OPTIONS,
+      postgrest: DEFAULT_POSTGREST_OPTIONS,
```
pcnc, line 92: There already is a `settings.db` project, although that targets database access-related modifiers.
This addition simply targets how the lib interfaces with PostgREST

kamilogorek:
Oh, why is it gone? 👀

PR #1260, opened 2024-08-23
fix: bump storage-js v2.7.0

fenos:

Dep updated

## What is the new behavior?

storage-js version update, with support for metadata and exists.


alaister (approved):
LGTM!

PR #1276, opened 2024-10-03
Fix: Authorization Cache Poisoning

andrisecops:
By sending a crafted HTTP request, it is possible to poison the cache of a non-dynamic server-side rendered route in the pages router (this does not affect the app router). When this crafted request is sent it could coerce Next.js to cache a route that is meant to not be cached and send a Cache-Control: `s-maxage=1, stale-while-revalidate` header which some upstream CDNs may cache as well.

CWE-349
CWE-639
[CVE-2024-46982](https://nvd.nist.gov/vuln/detail/CVE-2024-46982)

### DIFF
--- a/examples/next-ts/yarn.lock
+++ b/examples/next-ts/yarn.lock
@@ -2,67 +2,67 @@
 # yarn lockfile v1
 
 
-"@next/env@14.1.1":
-  version "14.1.1"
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks! Closing in favor of https://github.com/supabase/supabase-js/pull/1277

PR #1283, opened 2024-10-17
Updates supabase sub-modules to latest versions

lourd:
I just ran into the issue described in https://github.com/supabase/realtime/issues/282 — the `subscribe` does not mean that I'm successfully subscribed to `postgres_changes` - instead I need to listen on `'system'` messages for that confirmation, checking the `extension` and `status` properties, as @chasers described in that thread.

I searched around to see if the types for `@supabase/supabase-js` to include `'system'`, and was happily surprised to see that @filipecabaco added it in https://github.com/supabase/realtime-js/pull/422 last month.

Then I was confused to discover that this module hadn't been updated since then.

Thankfully a simple fix! Just updating these version numbers. I went ahead and update the other supabase modules that are behind their latest versions, too.

It would be ideal if this module was automatically updated when its supabase sub-modules are updated, as I think everyone is mostly using them via this module, as the docs say to do. Or perhaps a more lenient version constraint at least so users can more easily update them on their own?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
       "version": "0.0.0-automated",
       "license": "MIT",
       "dependencies": {
-        "@supabase/auth-js": "2.65.0",
-        "@supabase/functions-js": "2.4.1",
+        "@supabase/auth-js": "2.65.1",
+        "@supabase/functions-js": "2.4.3",
         "@supabase/node-fetch": "2.6.15",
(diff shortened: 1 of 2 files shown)


soedirgo (approved):

lourd:
Thanks @soedirgo! What's an ETA on publishing a new version you think?

soedirgo:
Oops, meant to release a new version immediately - gimme a moment

PR #1293, opened 2024-10-27
feat: do not send non-JWTs in `Authorization` header

hf:
In preparation for the introduction of new API keys, the client library should not pass non-JWT values in the `Authorization` header, as requests like this will definitely fail.

Two backward compatible changes are introduced:

- Configuring a `createClient()` with a non-JWT Supabase Key will not forward it as the `Authorization: Bearer <...>` header.
- Configuring a `createClient()` with global header options that contains a non-JWT in the `Authorization` header will throw an error.

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -66,3 +66,60 @@ export function applySettingDefaults<
 
   return result
 }
+
+export const BASE64URL_REGEX = /^([a-z0-9_-]{4})*($|[a-z0-9_-]{3}$|[a-z0-9_-]{2}$)$/i
+
+/**
+ * Checks that the value somewhat looks like a JWT, does not do any additional parsing or verification.
+ */
+export function isJWT(value: string): boolean {
+  if (value.startsWith('Bearer ')) {
+    value = value.substring('Bearer '.length)
+  }
+
+  value = value.trim()
+
+  if (!value) {
+    return false
+  }
+
+  const parts = value.split('.')
+
+  if (parts.length !== 3) {
+    return false
+  }
+
+  for (let i = 0; i < parts.length; i += 1) {
+    const part = parts[i]
+
+    if (part.length < 4 || !BASE64URL_REGEX.test(part)) {
+      return false
+    }
+  }
+
+  return true
+}
+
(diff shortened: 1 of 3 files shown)


src/lib/helpers.ts
```diff
@@ -66,3 +66,60 @@ export function applySettingDefaults<
...
+  }
+
+  return true
+}
+
+export function checkAuthorizationHeader(headers: { [header: string]: string }) {
+  if (headers.authorization && headers.Authorization) {
+    console.warn(
+      '@supabase-js: Both `authorization` and `Authorization` headers specified in createClient options. `Authorization` will be used.'
+    )
```
kangmingtay, line 107: can we initialise the `headers` field using the `Headers` interface instead since it's case-insensitive? (https://developer.mozilla.org/en-US/docs/Web/API/Headers)

src/lib/helpers.ts
```diff
@@ -66,3 +66,60 @@ export function applySettingDefaults<
...
+    delete headers.authorization
+  }
+
+  const authorization = headers.Authorization ?? headers.authorization ?? null
+
+  if (!authorization) {
+    return
+  }
+
+  if (authorization.startsWith('Bearer ') && authorization.length > 'Bearer '.length) {
```
kangmingtay, line 118: > `authorization.length > 'Bearer '.length` 

not immediately obvious what this check is for?

kangmingtay (commented):
@hf just a few things to confirm:
1. Users don't need to upgrade their `supabase-js` version to use the new API keys
2. This change just helps to ease the transition since upgrading the client lib version will provide more information about any unforseen errors. Our api-gateway already ignores checking the `Authorization` header and only looks at the `apikey` header.

src/lib/helpers.ts
j4crev: Looks like it's making sure that we only check for a valid JWT if the authorization header actually has a value after "Bearer ".

It's worth noting that this check will not cover a scenario where someone only passes in "Bearer" (without a trailing space), but I'm guessing this whole scenario would be extremely rare anyway. Still, you might consider covering both.

mandarini:
After discussing with @hf and receiving more context, I am closing this PR.

PR #1232, opened 2024-06-27
fix: use global fetch instead of @supabase/node-fetch

m-shaka:
## What kind of change does this PR introduce?

Fix #914 

## Additional context
As I posted here, this won't work on Node.js <18
https://github.com/supabase/supabase-js/issues/914#issuecomment-2193910905

PR #1306, opened 2024-11-09
chore(lib/fetch): dynamic import `@supabase/node-fetch`

### LINKED ISSUE #1303: Request to dynamic import `@supabase/node-fetch`
# Chore

## Describe the chore

Request update the  static import of `@supabase/node-fetch`:

https://github.com/supabase/supabase-js/blob/4c7f57197c0109b9393080db5971543347a6397a/src/lib/fetch.ts#L2-L17

https://github.com/supabase/supabase-js/blob/4c7f57197c0109b9393080db5971543347a6397a/src/lib/fetch.ts#L18-L24

Use dynamic `import('@supabase/node-fetch')` to avoid possible `Cannot read property 'bind' of undefined` error.

## Additional context

Related to this issue https://github.com/supabase/postgrest-js/issues/574.

Using `@supabase/supabase-js`, a user is able to provide custom fetch by providing a `global.fetch` option. However, with @supabase/node-fetch` is statically imported, there could be `Cannot read property 'bind' of undefined` error at runtime.

```
TypeError: Cannot read property 'bind' of undefined
    at path/to/node_modules/@supabase/node-fetch/browser.js
```

The error is thrown when the runtime environment does not have native `fetch` on `globalObject` at all.

https://github.com/supabase/node-fetch/blob/bd8f50a4ae647e56cda296b8301d7dff01d4a87b/browser.js#L18.
```javascript
/* @supabase/node-fetch/browser.js#L18 */
export default globalObject.fetch.bind(globalObject);
``` 

In some other `@supabase/` packages like `@supabase/auth-js`, the fetch is resolved using a `resolveFetch` function: 

https://github.com/supabase/auth-js/blob/8af88b6f4e41872b73e84c40f71793dab6c62126/src/lib/helpers.ts#L93-L104.
```typescript
/* @supabase/auth-js/src/lib/helpers.ts#L93-L104. */
export const resolveFetch = (customFetch?: Fetch): Fetch => {
  let _fetch: Fetch
  if (customFetch) {
    _fetch = customFetch
  } else if (typeof fetch === 'undefined') {
    _fetch = (...args) =>
      import('@supabase/node-fetch' as any).then(({ default: fetch }) => fetch(...args))
  } else {
    _fetch = fetch
  }
  return (...args) => _fetch(...args)
}
``` 

This way, the `node-fetch` import is safe. We could also implement this for `nodeFetch` and `Headers` imported here.

liyachun01:
Dynamic import('@supabase/node-fetch') in `lib/fetch` to avoid possible `Cannot read property 'bind' of undefined` error.

Closes #1303

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -1,27 +1,5 @@
-// @ts-ignore
-import nodeFetch, { Headers as NodeFetchHeaders } from '@supabase/node-fetch'
-
-type Fetch = typeof fetch
-
-export const resolveFetch = (customFetch?: Fetch): Fetch => {
-  let _fetch: Fetch
-  if (customFetch) {
-    _fetch = customFetch
-  } else if (typeof fetch === 'undefined') {
-    _fetch = nodeFetch as unknown as Fetch
-  } else {
-    _fetch = fetch
-  }
-  return (...args: Parameters<Fetch>) => _fetch(...args)
-}
-
-export const resolveHeadersConstructor = () => {
-  if (typeof Headers === 'undefined') {
-    return NodeFetchHeaders
-  }
-
-  return Headers
-}
+import { resolveFetch, resolveHeadersConstructor } from './helpers'
+import { Fetch } from './types'
 
 export const fetchWithAuth = (
   supabaseKey: string,
@@ -29,10 +7,11 @@ export const fetchWithAuth = (
   customFetch?: Fetch
 ): Fetch => {
   const fetch = resolveFetch(customFetch)
-  const HeadersConstructor = resolveHeadersConstructor()
 
   return async (input, init) => {
     const accessToken = (await getAccessToken()) ?? supabaseKey
+
+    const HeadersConstructor = await resolveHeadersConstructor()
     let headers = new HeadersConstructor(init?.headers)
 
     if (!headers.has('apikey')) {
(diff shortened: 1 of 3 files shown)


liyachun01:
Similar updates are made to `@supabase/postgrest-js`, see https://github.com/supabase/postgrest-js/pull/576

liyachun01:
Hi @soedirgo, please kindly look at this PR, thanks~

jsprds:
In my opinion this is a very good improvement, thanks for the work @liyachun01!  This will probably also solve problems when using `supabase-js` in an Angular project with SSR enabled. 

I hope that this PR can be reviewed and merged in the near future. 🤞

mandarini:
Hi @liyachun01 ! Thank you so much for your contribution and for taking the time to address issue #1303!

I will not be merging this PR now, as the `async` approach introduces a performance regression for all browser requests (every request now requires an unnecessary `await` even when Headers is immediately available).

That being said, once we officially drop support for Node 18 (October 31st) [as we recently announced](https://github.com/orgs/supabase/discussions/37217), and remove the `@supabase/node-fetch` dependency on November 1st, this issue will be resolved entirely without any performance impact.

I know this may not be the immediate solution you were hoping for, but I want to avoid introducing performance costs that affect all users when we have a clean, permanent fix coming in just a few months. I do hope you understand, and thank you very very much for your patience. I am also going to refer you to [this PR](https://github.com/supabase/supabase-js/pull/1232), where I replied almost the same thing. I do hope my rationale makes sense for you.

If you think there's another solution that would keep everyone happy until November, please let me know! :D

Again, thank you very much for your patience and for contributing to Supabase!

PR #1329, opened 2024-12-10
fix: call set auth for realtime on token changes

filipecabaco:

call set auth for realtime on token changes

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -318,18 +318,19 @@ export default class SupabaseClient<
   }
 
   private _listenForAuthEvents() {
-    let data = this.auth.onAuthStateChange((event, session) => {
-      this._handleTokenChanged(event, 'CLIENT', session?.access_token)
+    let data = this.auth.onAuthStateChange(async (event, session) => {
+      await this._handleTokenChanged(event, 'CLIENT', session?.access_token)
     })
     return data
   }
 
-  private _handleTokenChanged(
+  private async _handleTokenChanged(
     event: AuthChangeEvent,
     source: 'CLIENT' | 'STORAGE',
     token?: string
   ) {
-    this.realtime.setAuth()
+    // On any token change we call the realtime setAuth function that it's by itself a callback and lets Realtime run the flow needed by it.
+    await this.realtime.setAuth()
     if (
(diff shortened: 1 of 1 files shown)


src/SupabaseClient.ts
```diff
@@ -329,13 +329,13 @@ export default class SupabaseClient<
     source: 'CLIENT' | 'STORAGE',
     token?: string
   ) {
+    this.realtime.setAuth()
```
avallete: **question**

Not a lot of context here, but I wonder why we call it on top of the function an not instead something like:

```ts
_handleTokenChanged() {
   ...
   this.realtime.setAuth(this.changedAccessToken)
}
```

Or is it not the same token ? And we need the `setAuth` to generate another one for the realtime ?

filipecabaco: We changed it to be a callback function so on token change realtime runs the `accessToken` callback and pulls credentials as defined by said callback and pushes it to connected channels (required for proper Realtime operations)

https://github.com/supabase/realtime-js/blob/9ef764907abc202ce0efd5769149660b4e0f928c/src/RealtimeClient.ts#L357

filipecabaco: Here's the code on the supabase-js side https://github.com/supabase/supabase-js/blob/325c2c9b253a81cf1f02ec7e2bdbced7318467f8/src/SupabaseClient.ts#L122

avallete (approved):

src/SupabaseClient.ts
```diff
@@ -318,24 +318,25 @@ export default class SupabaseClient<
...
     return data
   }
 
-  private _handleTokenChanged(
+  private async _handleTokenChanged(
     event: AuthChangeEvent,
     source: 'CLIENT' | 'STORAGE',
     token?: string
   ) {
+    // On any token change we call the realtime setAuth function that it's by itself a callback and lets Realtime run the flow needed by it.
```
avallete: Would go for something like:
```suggestion
// On token change, call Realtime's `setAuth` to sync auth and connections.
// Realtime handles token retrieval internally which may involve customizations, so we don't pass the token directly.
```

avallete (approved):

PR #1340, opened 2024-12-18
DRAFT: (breaking change) createAdminClient()

kiwicopple:
throws an error if the user runs createClient() with a service_role key

This is a spike for internal discussion. It would be considered a breaking change as developers would need to change their Adming init code from

```js

const admin = createClient('SUPABASE_URL', 'SERVICE_ROLE_KEY')
```
to 

```js
import { createAdminClient } from '@supabase/supabase-js'

const admin = createAdminClient('SUPABASE_URL', 'SERVICE_ROLE_KEY')
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,8 +21,46 @@ export * from '@supabase/realtime-js'
 export { default as SupabaseClient } from './SupabaseClient'
 export type { SupabaseClientOptions, QueryResult, QueryData, QueryError } from './lib/types'
 
+interface JwtPayload {
+  role: string
+  iss: string
+  iat: number
+  exp: number
+}
+
+function decodeJWT(jwt: string): JwtPayload | null {
+  try {
+    const base64Url = jwt.split('.')[1]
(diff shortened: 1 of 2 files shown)


kiwicopple:
For now, perhaps we start simply with a prominent warning:

```
console.log(
  '%c⚠️ WARNING ⚠️',
  'color: white; background-color: red; font-size: 20px; font-weight: bold; padding: 5px 10px;'
);
console.log(
  '%cYou are using a Supabase admin key in a browser environment. This poses a serious security risk and could expose your database to unauthorized access. NEVER use an admin key in client-side code!',
  'color: red; font-size: 16px; font-weight: bold;'
);
```
image

grdsdev:
Yes @kiwicopple preferable to go first with the warning

kiwicopple:
Discussed internally - we are going to raise these alerts in the API logs inside the project dashboard (and email users). This is more targeted to the developer who is building the project

PR #1341, opened 2024-12-18
fix: Realtime set auth on handleTokenChanged

filipecabaco:

Realtime set auth on handleTokenChanged

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -336,10 +336,11 @@ export default class SupabaseClient<
       this.changedAccessToken !== token
     ) {
       this.changedAccessToken = token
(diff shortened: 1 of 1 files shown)


salmoro:
Thanks @filipecabaco

If this solves what I think it does then it's huge.
We've been noticing `nvalidJWTToken: Invalid value for JWT claim "exp" with value {number}` from realtime in our app which will randomly kill our users session.
Seems like this should fix it.

filipecabaco:
Hey @salmoro , yes I think this will help but it's not enough.. I might need to do some more changes to realtime-js to ensure we send the proper token before subscribe and that might mean a deeper change to the way we subscribe

Issue #1338, opened 2024-12-15
Warning: "The ?? operator will always return the left operand" in SupabaseClient.ts


leongobbs:
I encountered a warning while using `supabase-js@2.47.7` with Deno. The warning states that the `??` operator will always return the left operand because the left-hand expression can never be `null` or `undefined`. This seems to be unnecessary usage of the `??` operator.

Here’s the exact warning:

```
[WARNING] The "??" operator here will always return the left operand [suspicious-nullish-coalescing]

https://jsr.io/@supabase/supabase-js/2.47.7/src/SupabaseClient.ts:309:68:
  309 │       hasCustomAuthorizationHeader: 'Authorization' in this.headers ?? false,
      ╵                                                                     ~~
The left operand of the "??" operator here will never be null or undefined, so it will always be returned.
```

**Steps to Reproduce:**

1. Import `supabase-js` in a Deno project:
```
  import { createClient } from "jsr:@supabase/supabase-js@2";

const options = {
    global: {
        headers: { "x-my-custom-header": "my-app" }, 
    },
    auth: {
        persistSession: true, 
        detectSessionInUrl: true, 
        autoRefreshToken: true, 
    },
};

const supabaseUrl = "https://ayomz.supabase.co";
const supabaseKey =
    "my_key";

const supabase = createClient(supabaseUrl, supabaseKey, options);

export { supabase };
```

2. Run the project in Deno with the following command:
   ```bash
   deno run --allow-net main.ts
   ```

3. Observe the warning in the terminal.

**Expected Behavior:**
No warning should be emitted, as this seems like an unnecessary check.

**Actual Behavior:**
The following warning is displayed in the terminal:
```
[WARNING] The "??" operator here will always return the left operand [suspicious-nullish-coalescing]
```

**Proposed Fix:**
It seems that the nullish coalescing operator (`?? false`) can be removed in this case since `'Authorization' in this.headers` will always return a boolean (`true` or `false`).

The current code:
```typescript
hasCustomAuthorizationHeader: 'Authorization' in this.headers ?? false,
```

Suggested fix:
```typescript
hasCustomAuthorizationHeader: 'Authorization' in this.headers,
```

**Environment:**
- Library version: `supabase-js@2.47.7`
- Environment: Deno
- Deno version: `1.37.0`

**Additional Context:**
This warning does not break functionality but clutters the logs during development.

j4w8n:
Nice catch @leongobbs! Would you like to submit a PR for this?

[...]

PR #1344, opened 2024-12-23
fix(lint): nullish coalescing operator issue in hasCustomAuthorizationHeader

raghavyuva:
fixes the following bug : 

The left operand of the "??" operator here will never be null or undefined, so it will always be returned.

Bug fix

Please link any relevant issues here.
#1338


avallete (approved):
Thank you for your contribution 👍

leongobbs:
Now how can I fix this for myself?

PR #1356, opened 2025-01-17
feat(types): also export type FunctionsResponse from functions-js

maximilian-hammerl:

Also export type `FunctionsResponse` from functions-js

As the return type of the `invoke` method of the `FunctionsClient` class, the `FunctionsResponse` type is significant if one wants to type the response when calling the `invoke` method.

Currently, one has to have functions-js as a dependency just for the `FunctionsResponse` type.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ export {
   FunctionsError,
   type FunctionInvokeOptions,
   FunctionRegion,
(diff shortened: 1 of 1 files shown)


maximilian-hammerl:
Hi @avallete, anything missing for this PR to get approved and merged? Anything I can help with?

PR #1357, opened 2025-01-17
feat(types): export more PostgrestResponse, PostgrestBuilder and PostgrestClient types from postgrest-js

maximilian-hammerl:

Export more `PostgrestResponse`, `PostgrestBuilder` and `PostgrestClient` types from postgrest-js

As the return type of, among others, the filter methods of the `PostgrestFilterBuilder` class, the types are significant if one wants to type the response when calling the filter methods.

Currently, one has to have postgrest-js as a dependency just for the `PostgrestResponse`, `PostgrestBuilder` and `PostgrestClient` types.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,8 +5,15 @@ export * from '@supabase/auth-js'
 export type { User as AuthUser, Session as AuthSession } from '@supabase/auth-js'
 export {
   type PostgrestResponse,
+  type PostgrestResponseFailure,
+  type PostgrestResponseSuccess,
   type PostgrestSingleResponse,
   type PostgrestMaybeSingleResponse,
+  type PostgrestClient,
+  type PostgrestQueryBuilder,
+  type PostgrestFilterBuilder,
+  type PostgrestTransformBuilder,
+  type PostgrestBuilder,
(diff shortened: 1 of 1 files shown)


maximilian-hammerl:
Hi @avallete, anything missing for this PR to get approved and merged? Anything I can help with?

avallete:
Actually I'm not too sure we want to expose those kind of types for external use.

People might feel encouraged to build on top of it and expect their behavior to remain stable while we could potentially break them, or remove them entirely as they are a technical implementation detail in our side. Potentially making us to have to make a new major release because of potential breaking changes in types every-time we fix a behavior. 

ping @soedirgo what do you think ?

If we don't mind making more frequent breaking changes/major releases then I guess we could go for it.

soedirgo:
Yeah, these types are internal implementation detail as far as supabase-js is concerned, so we don't want users to be dependent on their stability.

For your use case, you can consider using TypeScript's `ReturnType` instead - I don't foresee that causing issues if we change the return types of filter methods etc.

soedirgo:
Closing this as we don't plan to export these types

PR #1411, opened 2025-05-12
fix: bump up realtime-js; await on token change

filipecabaco:

 bump up realtime-js; await on token change

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -117,11 +117,13 @@ export default class SupabaseClient<
     }
 
     this.fetch = fetchWithAuth(supabaseKey, this._getAccessToken.bind(this), settings.global.fetch)
+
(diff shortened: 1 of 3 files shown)


src/SupabaseClient.ts
```diff
@@ -319,14 +321,13 @@ export default class SupabaseClient<
     })
   }
 
-  private _listenForAuthEvents() {
-    let data = this.auth.onAuthStateChange((event, session) => {
+  private async _listenForAuthEvents() {
```
grdsdev, line 324: the caller doesn't use an `await`, isn't this an issue?

grdsdev (approved):

src/SupabaseClient.ts
filipecabaco: yes indeed, changing

PR #1412, opened 2025-05-12
fix: prevent async calls in _listenForAuthEvents

filipecabaco:

prevent async calls in _listenForAuthEvents ( https://supabase.com/docs/reference/javascript/auth-onauthstatechange )


grdsdev (approved):

Snehil-Shah:
@filipecabaco I'd presume this eliminates the need for the hack to avoid deadlocks as mentioned in the [docs](https://supabase.com/docs/reference/javascript/javascript), right?

filipecabaco:
can you be more specific?

Snehil-Shah:
@filipecabaco The docs ([here](https://supabase.com/docs/reference/javascript/auth-onauthstatechange)) mention that awaiting async supabase functions inside the callback can cause a deadlock, and recommends dispatching it after the callback ends, as a hack. Does this PR fix that?

filipecabaco:
That still needs to be avoided as this only fixed in this specific scenario and not fully as this flow really should not be async

PR #1425, opened 2025-05-21
fix: preserve path segments in supabaseUrl when constructing endpoints.

### LINKED ISSUE #1421: Regression using paths in supabaseUrl since 2.49.5
# Bug report

## Describe the bug

Starting from PR #1417 (released with 2.49.5), the way Supabase client URLs are constructed has changed. Previously, endpoint URLs (e.g., auth, realtime) were generated by concatenating the base Supabase URL with the corresponding paths. The new implementation uses the [URL constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) instead.

However, because the endpoint paths begin with a leading `/`, any path component included in the base Supabase URL is discarded. This means base URLs that include a path no longer behave as expected.

## To Reproduce

This constructs the URLs in a similar way like it is done in the [current code](https://github.com/supabase/supabase-js/blob/bc4bce1cae69546d42a8cea9e9a20846d5654be3/src/SupabaseClient.ts#L78)

```
const realtimeUrl = "/realtime/v1";
const supabaseUrlWithoutTrailingSlash = "https://example.com/path/to/supabase"; // Worked in versions ≤ 2.49.4
const constructorResultUrl1 = new URL(realtimeUrl, supabaseUrlWithoutTrailingSlash);
// ==>  https://example.com/realtime/v1

const supabaseUrlWithTrailingSlash = "https://example.com/path/to/supabase/";
const constructorResultUrl2 = new URL(realtimeUrl, supabaseUrlWithTrailingSlash);
// ==>  https://example.com/realtime/v1

```

## Expected behavior

The base Supabase URL should preserve its path component until the last forward-slash when constructing service URLs, as it did in versions up to and including 2.49.4.

## Possible fix:

Remove leading slash from endpoint path and do NOT strip the trailing slash from the supabaseUrl.

```
const realtimeUrlNew = "realtime/v1";
const supabaseUrlWithTrailingSlash = "https://example.com/path/to/supabase/";
const constructorResultUrl3 = new URL(realtimeUrlNew, supabaseUrlWithTrailingSlash);
// ==> https://example.com/path/to/supabase/realtime/v1
```

## System information

- Version of supabase-js: 2.49.5

etzelc:

This PR ensures that path segments in the `supabaseUrl` are preserved when constructing service endpoint URLs (e.g., auth, realtime, etc.). 

Fixes #1421.

Currently, any path component in the supabaseUrl is discarded when constructing endpoint URLs using the URL constructor.

Example:  `const supabase = createClient('https://example.com/path/to/supabase/', 'public-anon-key')` the resulting endpoint is incorrectly resolved as `https://example.com/realtime/v1`.

See #1421 for more context.

## What is the new behavior?

This fix preserves the full path in supabaseUrl when resolving endpoint URLs.
For example, if `supabaseUrl` is `https://example.com/path/to/supabase/`, the realtime endpoint will now correctly resolve to `https://example.com/path/to/supabase/realtime/v1`.

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -19,7 +19,7 @@ import {
   DEFAULT_REALTIME_OPTIONS,
 } from './lib/constants'
 import { fetchWithAuth } from './lib/fetch'
-import { stripTrailingSlash, applySettingDefaults } from './lib/helpers'
+import { ensureTrailingSlash, applySettingDefaults } from './lib/helpers'
 import { SupabaseAuthClient } from './lib/SupabaseAuthClient'
 import { Fetch, GenericSchema, SupabaseClientOptions, SupabaseAuthClientOptions } from './lib/types'
 
@@ -75,7 +75,8 @@ export default class SupabaseClient<
     if (!supabaseUrl) throw new Error('supabaseUrl is required.')
     if (!supabaseKey) throw new Error('supabaseKey is required.')
 
-    const baseUrl = new URL(supabaseUrl)
+    const _supabaseUrl = ensureTrailingSlash(supabaseUrl)
+    const baseUrl = new URL(_supabaseUrl)
 
     this.realtimeUrl = new URL('realtime/v1', baseUrl)
     this.realtimeUrl.protocol = this.realtimeUrl.protocol.replace('http', 'ws')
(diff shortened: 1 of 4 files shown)


etzelc:
You're right — good catch. My initial approach aimed to stay aligned with the behavior of `new URL(...)`, but I agree that's not ideal for a bugfix in a patch release. I'll revise the fix to match the behavior from versions ≤ 2.49.4

etzelc:
I haven't updated the docs. Should I do that as part of this PR, or is it typically handled during the release process?
As part of this change, I removed `stripTrailingSlash(...)` and introduced a new helper `ensureTrailingSlash(...)`.

grdsdev (approved):

grdsdev:
Thanks for the contribution @etzelc

PR #1430, opened 2025-05-27
test: run integration tests in CI

grdsdev:

Add a new `integration` CI job for running:
- Integration tests
- Browser compatibility tests


.github/workflows/ci.yml
```diff
@@ -46,3 +46,44 @@ jobs:
...
+          version: latest
+
+      - name: Start Supabase
+        run: |
+          supabase start
+
+      - name: Run tests
+        run: |
+          npm clean-install
+          npm run test:integration || npm run test:integration
```
grdsdev, line 84: need to run this twice, the first one always fail in CI, but the second passes

@filipecabaco any ideia why? maybe first call "warms" realtime, and it fails because it isn't ready yet?

filipecabaco (approved):

.github/workflows/ci.yml
filipecabaco: no clue honestly ... supabase cli should guarantee that it's in a good state to be used

PR #1472, opened 2025-06-30
feat: use dedicated storage host for storage lib (allows >50GB uploads)

itslenny:

Feature

Storage lib is initialized using the same host as all other Supabase APIs. This zone does not allow uploading of large files.

## What is the new behavior?

Initialize storage lib using new `project-ref.storage.supabase.co/v1` URI which supports uploads up to 500GB

## Additional context

This is specific to the Supabase infrastructure

`project-ref.supabase.co/storage/v1` -> `project-ref.storage.supabase.co/v1`


itslenny:
Moved change to storage-js instead: https://github.com/supabase/storage-js/pull/230

PR #1475, opened 2025-07-02
fix: setAuth ran automatically for realtime; prioritize setAuth tasks

filipecabaco:

* setAuth ran automatically for realtime
* prioritize setAuth tasks
* tried a new approach to test where we move to protected visibility so we have a test implementation for SupabaseClient that mostly binds protected to public methods


hf (approved):
Looks OK

PR #1506, opened 2025-07-16
fix: add Node 18 deprecation notice

grdsdev:

As announced in https://github.com/orgs/supabase/discussions/37217, Node 18 support has been deprecated.

This PR:
- adds a console warning when this lib is used on Node 18 environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,8 +86,5 @@
   },
   "jsdelivr": "dist/umd/supabase.js",
   "unpkg": "dist/umd/supabase.js",
-  "engines": {
-    "node": ">=20"
-  },
   "packageManager": "pnpm@10.11.0+sha512.6540583f41cc5f628eb3d9773ecee802f4f9ef9923cc45b69890fb47991d4b092964694ec3a4f738a420c918a333062c8b925d312f42e4f0c263eb603551f977"
 }
--- a/package.json
+++ b/package.json
@@ -86,5 +86,8 @@
   },
   "jsdelivr": "dist/umd/supabase.js",
   "unpkg": "dist/umd/supabase.js",
+  "engines": {
+    "node": ">=20"
+  },
   "packageManager": "pnpm@10.11.0+sha512.6540583f41cc5f628eb3d9773ecee802f4f9ef9923cc45b69890fb47991d4b092964694ec3a4f738a420c918a333062c8b925d312f42e4f0c263eb603551f977"
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,8 +8,8 @@ importers:
   .:
     dependencies:
       '@supabase/auth-js':
-        specifier: 2.71.0
-        version: 2.71.0
+        specifier: 2.70.0
+        version: 2.70.0
       '@supabase/functions-js':
         specifier: 2.4.5
         version: 2.4.5
@@ -17,8 +17,8 @@ importers:
         specifier: 2.6.15
         version: 2.6.15
       '@supabase/postgrest-js':
-        specifier: 1.19.4
-        version: 1.19.4
+        specifier: 1.21.0
+        version: 1.21.0
       '@supabase/realtime-js':
         specifier: 2.11.15
         version: 2.11.15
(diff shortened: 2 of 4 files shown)


soedirgo:
I just realized `engines` is [advisory only](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) and won't get enforced unless `engine-strict` is set. Should we do that instead?

grdsdev:
Great @soedirgo , didn't know that.

I will add the engine field

package.json
```diff
@@ -86,5 +86,8 @@
   },
   "jsdelivr": "dist/umd/supabase.js",
   "unpkg": "dist/umd/supabase.js",
+  "engines": {
+    "node": ">=20"
```
lforst: Just sharing my experience here: Many many people see this as a breaking change as some package managers will outright fail if they encounter conflicts with the used node version which can cause breakage in CI.

.github/workflows/ci.yml
```diff
@@ -92,6 +92,19 @@ jobs:
...
+  coveralls-finish:
+    name: Coveralls Finished
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      - name: Coveralls Finished
+        uses: coverallsapp/github-action@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel-finished: true
```
mandarini: This change seems out of context. Maybe it should be added in the description, or in a separate PR?

package.json
mandarini: Should we at least add a note that it's a potential breaking change in some configurations then in this case?

pnpm-lock.yaml
```diff
@@ -8,17 +8,17 @@ importers:
...
         specifier: 2.4.5
         version: 2.4.5
       '@supabase/node-fetch':
         specifier: 2.6.15
         version: 2.6.15
       '@supabase/postgrest-js':
-        specifier: 1.21.0
-        version: 1.21.0
+        specifier: 1.19.4
+        version: 1.19.4
```
mandarini: What happened here? Why downgraded? (and why the others upgraded in any case, maybe we should look into it?)

pnpm-lock.yaml
```diff
@@ -622,10 +622,10 @@ packages:
       }
     engines: { node: 4.x || >=6.0.0 }
 
-  '@supabase/postgrest-js@1.21.0':
+  '@supabase/postgrest-js@1.19.4':
```
mandarini: What happened here? Why downgraded? (and why the others upgraded in any case, maybe we should look into it?)

pnpm-lock.yaml
```diff
@@ -5599,7 +5599,7 @@ snapshots:
     dependencies:
       whatwg-url: 5.0.0
 
-  '@supabase/postgrest-js@1.21.0':
+  '@supabase/postgrest-js@1.19.4':
```
mandarini: What happened here? Why downgraded? (and why the others upgraded in any case, maybe we should look into it?)

package.json
grdsdev: If that can be treated as breaking change by some package managers, then I'd rather remove it and keep only the `console.warning`.

.github/workflows/ci.yml
grdsdev: That was done mainly because coveralls was failing for some reason, but it makes more sense to break it into another PR.

pnpm-lock.yaml
grdsdev: This happened because we currently have both `npm` and `pnpm`, and since pnpm is less used, sometimes it gets outdated.

package.json
mandarini: What is said [here](https://github.com/supabase/supabase-js/pull/1506#issuecomment-3105582248) 
> won't get enforced unless `engine-strict` is set

this narrows down the cases this can be a breaking change, but there still are such cases.

I think what you suggest is safer, have the warning, and roll out the `engines.node` setting after some time?

.github/workflows/ci.yml
grdsdev: moved to https://github.com/supabase/supabase-js/pull/1510

package.json
grdsdev: I rollback a few changes, please review it again @mandarini

mandarini (approved):

package.json
soedirgo: @lforst good call; had a feeling that could be the case. Let's go ahead with the `console.warning`.

soedirgo (approved):

PR #1525, opened 2025-08-08
test(types): use `tstyche` instead of `tsd`

mrazauskas:

This PR migrates type test to TSTyche. Similar to https://github.com/supabase/postgrest-js/pull/633

## What is the new behavior?

This repo has only one test file, so I took an opportunity to demo TSTyche specific features: `test()` helpers and assertions like `.toBe()`, `.not.toBeCallableWith()`.

In the PR of `supabase/postgrest-js` repo, checks of `// @ts-expect-error` are used. Which is a good idea, of course. Only that it makes testing agains many versions of TypeScript less smooth, because of changing message text.

In a way the message text is only implementation detail. In this PR I used `.not.toBeCallableWith()` which tries to call the given expression and passes if there is any error. (It also checks if the expression is actually callable.) This way testing can be done across several TypeScript versions.

By the way, the actual calling of the expression is implemented via erased syntax:

```ts
expect(supabase.from).type.not.toBeCallableWith(42)
```

roughly becomes:

```ts
;     (supabase.from)                          (42)
```

## Additional context

Current version of `tsd` ships and uses TypeScript 5.3.3. TSTyche is using the installed version of TypeScript which currently is 4.7.4.

As suggested in the above mentioned PR, in CI it would be possible to run tests agains a range of TypeScript versions: `tstyche --target '>=4.7'`. The exact range should be decided.

---

Sorry, I can’t figure out how to update `deno.lock`.


package.json
```diff
@@ -39,7 +39,7 @@
     "test:watch": "jest --watch --verbose false --silent false",
     "test:node:playwright": "cd test/integration/node-browser && npm install && cp ../../../dist/umd/supabase.js . && npm run test",
     "test:bun": "cd test/integration/bun && bun install && bun test",
-    "test:types": "run-s build:module && tsd --files test/types/*.test-d.ts && jsr publish --dry-run",
+    "test:types": "run-s build:module && tstyche && jsr publish --dry-run",
```
mrazauskas, line 42: What if the `jsr` part would be move to a separate `test:jsr` script? This way `test:types` could be passed `--target '>=4.7'` in CI.

avallete:
Tstyche will be kept as a way to cross-test our typings for multiples TS version with the basic `expectType(TypeEqual)` syntax for now.

Once we're comfortable with it, we might migrate more of the codebase to it. Closing for now.

mrazauskas:
Fine with me (; Thanks for feedback!

PR #1528, opened 2025-08-11
fix: make Authorization header check case-insensitive

### LINKED ISSUE #1043: New check for a custom Authorization header is case sensitive.  Headers are case insensitive
# Bug report

- [x ] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The new code added 3 weeks ago is checking for an Authorization header.  It looks for 'Authorization' in the header array.  http headers are case insensitive.   

Just worked on a user issue where they loaded the header in authorization = .... and then used the header option `authorization` in the global headers object.
This should work.

The change was here @kangmingtay 
(https://github.com/supabase/supabase-js/blame/4dff472a15cb754bf9fa81af495a84b1f0c42b14/src/SupabaseClient.ts#L282)

## To Reproduce
use 
```
global: {
    headers: { 'authenticated': 'bearer myjwt' }
```
in createClient

## Expected behavior

Case of header name should not matter.

Louis454545:
## Summary
Fixes #1043

This PR makes the Authorization header detection case-insensitive, which is the correct behavior according to HTTP RFC standards. Previously, the code only checked for the exact string 'Authorization' in headers, but HTTP headers should be case-insensitive.

## Changes Made
- Replace case-sensitive `'Authorization' in this.headers` check with case-insensitive `Object.keys(this.headers).some(key => key.toLowerCase() === 'authorization')`
- This allows headers like 'authorization', 'Authorization', 'AUTHORIZATION' to all work correctly

## Testing
- [x] All existing unit tests pass
- [x] Build completes successfully
- [x] Manual testing with various header case combinations (lowercase, uppercase, mixed case)
- [x] Verified that no authorization header case still works correctly

## Root Cause
The issue was introduced in the hasCustomAuthorizationHeader check which was meant to detect custom auth headers but was using case-sensitive string matching instead of proper case-insensitive header handling.

## Additional Notes
This is a small but important fix that improves compatibility with various HTTP client implementations that may use different casing for the Authorization header.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -315,7 +315,9 @@ export default class SupabaseClient<
       fetch,
       // auth checks if there is a custom authorizaiton header using this flag
       // so it knows whether to return an error when getUser is called with no session
-      hasCustomAuthorizationHeader: 'Authorization' in this.headers,
+      hasCustomAuthorizationHeader: Object.keys(this.headers).some(
+        (key) => key.toLowerCase() === 'authorization'
+      ),
     })
   }
 


mandarini (approved):
Thank you for the contribution @Louis454545 !

PR #1537, opened 2025-08-19
feat: reland postgrest 13 support

soedirgo:
pending on https://github.com/supabase/postgrest-js/pull/637

Additionally reimplement support for accepting a schema name in the second type param (e.g. `SupabaseClient<Database, 'my_schema'>`. We have a lot of users doing this: https://github.com/search?q=createClient%3CDatabase%2C+%22&type=code

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -52,7 +52,11 @@ export default class SupabaseClient<
     : never,
   ClientOptions extends { PostgrestVersion: string } = SchemaNameOrClientOptions extends string &
     keyof Omit<Database, '__InternalSupabase'>
-    ? { PostgrestVersion: '12' }
+    ? // If the version isn't explicitly set, look for it in the __InternalSupabase object to infer the right version
+      Database extends { __InternalSupabase: { PostgrestVersion: string } }
+      ? Database['__InternalSupabase']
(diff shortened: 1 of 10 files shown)


src/SupabaseClient.ts
```diff
@@ -30,12 +30,32 @@ import { Fetch, GenericSchema, SupabaseClientOptions, SupabaseAuthClientOptions
  */
 export default class SupabaseClient<
   Database = any,
-  SchemaName extends string & keyof Database = 'public' extends keyof Database
+  // The second type parameter is also used for specifying db_schema, so we
+  // support both cases.
+  // TODO: Allow setting db_schema from ClientOptions.
+  SchemaNameOrClientOptions extends
+    | (string & keyof Database)
+    | { PostgrestVersion: string } = 'public' extends keyof Database
```
avallete: I guess what we wants in case of schemaName is to omit the '__InternalSupabase' ?
```suggestion
  SchemaNameOrClientOptions extends
    | (string & keyof Omit<Database, '__InternalSupabase'>)
    | { PostgrestVersion: string } = 'public' extends keyof Omit<Database, '__InternalSupabase'>
```

soedirgo: Yep good catch!

PR #1545, opened 2025-08-29
feat: expose auth `userStorage` option to facilitate lower cookie sizes

hf:
Adds the `userStorage` option to facilitate [the experimental split session storage capability in `auth-js`](https://github.com/supabase/auth-js/pull/1023). It will be propagated down to the `@supabase/ssr` package hopefully significantly decreasing cookie sizes.


src/lib/types.ts
```diff
@@ -38,6 +38,12 @@ export type SupabaseClientOptions<SchemaName> = {
      * A storage provider. Used to store the logged-in session.
      */
     storage?: SupabaseAuthClientOptions['storage']
+    /**
+     * A storage provider to store the user profile separately from the session. Useful when you need to store the session information in cookies, without bloating the data with the redundant user object.
```
cemalkilic:
```suggestion
     * A storage provider to store the user profile separately from the session.
     * Useful when you need to store the session information in cookies,
     * without bloating the data with the redundant user object.
```

just to make it look prettier

cemalkilic (approved):

Issue #1239, opened 2024-07-03
JS Client logs sensitive credentials to stdout/stderr


dmmulroy:
# Bug report
Hi, I was working with supabases' js client attempting to get storage upload working and I incorrectly configured the client. This resulted in my credentials being logged to stdout - the error message clearly indicates that it is knowledgable of the credentials and logs them out anyways. I was also streaming on twitch at the time and exposed my creds and had to roll them. I don't the think client should log sensitive creds to stdout/stderr.

![image](https://github.com/supabase/supabase-js/assets/2755722/e12b50cd-23da-4f7e-9f42-8a3eb8efc5d5)

PR #1547, opened 2025-09-01
fix: add validation for supabaseUrl

maxmetcalfe:

This PR adds basic url validation for the supaseUrl.

The client does not validate the supaseUrl (besides requiring it)

## What is the new behavior?

The client runs simple validation on the supaseUrl.

## Additional context

This will improve the experience for new Supabase users as well as help prevent the leaking of credentials to stdout/sterr (see: https://github.com/supabase/supabase-js/issues/1239).

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -86,12 +86,12 @@ export function validateSupabaseUrl(supabaseUrl: string): URL {
     throw new Error('supabaseUrl is required.')
   }
 
-  if (!supabaseUrl.match(/^https?:\/\/.+/i)) {
+  if (!trimmedUrl.match(/^https?:\/\//i)) {
     throw new Error('Invalid supabaseUrl: Must be a valid HTTP or HTTPS URL.')
   }
 
   try {
-    return new URL(ensureTrailingSlash(supabaseUrl))
+    return new URL(ensureTrailingSlash(trimmedUrl))
   } catch {
     throw Error('Invalid supabaseUrl: Provided URL is malformed.')
   }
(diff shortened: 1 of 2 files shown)


mandarini:
Hi there @maxmetcalfe! Thank you so much for this PR! Adding URL validation is a great security improvement that will help prevent accidental credential exposure (as mentioned in #1239). Really appreciate you taking the initiative on this.

The implementation looks solid! I have a few small suggestions that would make it even more robust:

### Suggested improvements:

1. **Trim the URL input**: Consider trimming whitespace before validation to be more forgiving of user input:
```typescript
   export function validateSupabaseUrl(supabaseUrl: string): URL {
     const trimmedUrl = supabaseUrl?.trim()
     
     if (!trimmedUrl) {
       throw new Error('supabaseUrl is required.')
     }
     // ... rest of validation
   }
```
2. **Simplify the regex**: The current regex `/^https?:\/\/.+/i` might be a bit restrictive. Consider using `/^https?:\/\//i` instead, since the URL constructor will handle the rest of the validation:
```typescript
   if (!trimmedUrl.match(/^https?:\/\//i)) {
     throw new Error('Invalid supabaseUrl: Must be a valid HTTP or HTTPS URL.')
   }
```
3. **Add edge case tests**: Would be great to add a couple more test cases:
```typescript
   // Test with leading/trailing whitespace
   expect(() => createClient('  https://xyz123.supabase.co  ', KEY)).not.toThrow()
   
   // Test with URL containing auth (should work but might want to warn in future)
   expect(() => createClient('http://user:pass@localhost:54321', KEY)).not.toThrow()
```
These are all minor suggestions, the core functionality is great as-is! Thanks again for your work on this! 🙏

maxmetcalfe:
thanks @mandarini, updated!

mandarini (approved):
Thank you @maxmetcalfe !

PR #1551, opened 2025-09-11
fix: use setAuth for realtime in signup flows also

filipecabaco:

use setAuth for realtime in signup flows also


TANGAZA-CO:
****

PR #1557, opened 2025-09-17
chore: do not fail on coveralls error

mandarini:

Add a `fail-on-error: false` to `coveralls` uploads. 

Coveralls seems to be failing, and being flaky. 

## What is the new behavior?

If `coveralls` upload fails, the CI will not fail.


uchitwai:
https://yoe-yar-san-build.lovable.app/

PR #1561, opened 2025-09-23
fix(realtime): prevent empty Authorization header in REST fallback

Maharshi2403:
- Avoid sending empty Authorization header when no session exists
- Use global.headers.Authorization or supabaseKey as fallback
- Fixes 500 error on /realtime/v1/api/broadcast
- Closes #38820

**Issue**: When calling `channel.send()` without `channel.subscribe()`, the Realtime client falls back to `/realtime/v1/api/broadcast`. Without a session, it sends an empty `Authorization: \r\n` header, causing a 500 error.

**Fix**: Updated `_getAccessToken` to:
- Check for a valid session before returning an access token.
- Use `global.headers.Authorization` or `supabaseKey` if no session exists.
- Prevents empty `Authorization` header in REST requests.

**Repro**: See issue #38820
- Before: `send ok? timed out` (500 error).
- After: `send ok? ok` with correct headers.

**Tests**: Verified with local repro and existing test suite.

### DIFF
--- a/src/SupabaseClient.ts
+++ b/src/SupabaseClient.ts
@@ -301,10 +301,21 @@ export default class SupabaseClient<
     if (this.accessToken) {
       return await this.accessToken()
     }
-
+  
     const { data } = await this.auth.getSession()
-
-    return data.session?.access_token ?? this.supabaseKey
+  
+    // If no session exists, check for global Authorization header or fall back to supabaseKey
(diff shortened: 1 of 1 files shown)


mandarini:
Closing in favor of https://github.com/supabase/supabase-js/pull/1937

Thank you for your contribution @Maharshi2403 !

Issue #438, opened 2022-05-06
Custom fetch per request


enyo:
# Feature request

When using the `load` function in Svelte Kit, it provides a [`fetch`](https://kit.svelte.dev/docs/loading) implementation that can be used.

From the official docs:

> SvelteKit's load receives an implementation of fetch, which has the following special properties:
> - it has access to cookies on the server
> - it can make requests against the app's own endpoints without issuing an HTTP call
> - it makes a copy of the response when you use it, and then sends it embedded in the initial page load for hydration

## Describe the solution you'd like

I think that it would be necessary to specify the fetch on a "per request" basis. Something like this:

```js
supabase.withFetch(fetch).from('mytable').select();
```

Full example:

```html
<script context="module">
  export const prerender = true;
  import { supabase } from "$lib/initSupabase";

  export async function load({ params, fetch, session, stuff }) {
    let { data, error } = await supabase.from("users").select();

    if (error) {
      return {
        status: 500,
      };
    } else {
      return {
        props: {
          users: data,
        },
      };
    }
  }
</script>
```

## Describe alternatives you've considered

Maybe I'm overthinking this. The way I've solved this so far is by creating a function:

```js
export const getSupabase = (fetch) => createClient(url, anonKey, { fetch });
```

I then use it in my `load` function.

Is there a downside in creating multiple clients like this? If not, this issue can probably be closed.

soedirgo:
Not sure about Auth and Realtime, but for REST this should be fine.

Alternatively (and this is a bit hacky):

```js
// initSupabase.js
let currentFetch;

export function setFetch(fetch) { // call this per request
  currentFetch = fetch;
}

export supabase = createClient(..., { fetch: (...args) => currentFetch(...args) });
```

w3b6x9:
I don't think this will impact Realtime.

soedirgo:
There shouldn't be an issue with using multiple clients, so closing this.

5hee75:
>There shouldn't be an issue with using multiple clients, so closing this.

I have definitely encountered an issue with multiple clients, specifically when using auth listeners.   I created a client factory, similar to OP, where each request to the SDK would create a new client.  I did this so I could inject dynamic headers and settings into the client for each execution.

```js
const supa = (...params) => createClient(...params);

supa().from...
supa().auth...
// etc
```

The issue I found was that auth listeners/events are registered to a specific client instance, not to the auth events themselves.  I'm not sure if this is intended or not.

[...]

PR #1570, opened 2025-09-30
feat(postgrest): expose query builder options to .from method - custom fetch per request

stenlan:

Introduces a way to pass options to the PostgrestQueryBuilder instance on a per-request basis.

There is no way to specify per-request options currently.
https://github.com/supabase/supabase-js/issues/438

## What is the new behavior?

An optional parameter is introduced that can be passed to the `.from` method, where users can optionally specify a custom `fetch` option.

## Additional context

This is a supporting PR for https://github.com/supabase/postgrest-js/pull/648 that updates the method signature in the @supabase/supabase-js package.

## Additions since the original description

- Per-request `retry` and `urlLengthLimit` are also accepted and override the client defaults.
- In supabase-js a per-request `fetch` is wrapped like the client fetch, so `apikey`, `Authorization` and trace headers are still injected. The caller's options object is not mutated.
- `PostgrestQueryBuilderOptions` is exported from `@supabase/postgrest-js` and re-exported from `@supabase/supabase-js`.
- Tests: supabase-js unit tests (custom fetch used, auth injection including schema clients, retry override, `db.timeout`, no mutation), server-free postgrest-js tests for the `from()` option plumbing and fetch decoration, and tsd assertions for the `from(relation, options)` signature.
- The `headers` option and its `mergeHeaders` helper were dropped after review. Headers for a single query go inside the custom `fetch`.
- postgrest-js `PostgrestClient` gains a `wrapFetch` constructor option. It is applied to the client fetch and to any per-request `fetch`, and carried over by `schema()`; the client `timeout` applies to per-request fetches too. supabase-js passes its auth wrapper as `wrapFetch`, so `supabase.from()` and `supabase.schema().from()` behave the same.
- Docs: JSDoc on `from()`, `schema()` and the new option.
- Rebased onto current master; `sdk-compliance.yaml` lists the options type as a supporting symbol.

### DIFF
--- a/packages/core/postgrest-js/src/utils.ts
+++ b/packages/core/postgrest-js/src/utils.ts
@@ -1,0 +1,22 @@
+/**
+ * Non-destructively merges an optional {@link HeadersInit} into a base
+ * {@link Headers} object. Right-side entries take precedence over left.
+ */
+export function mergeHeaders(left: Headers, right?: HeadersInit): Headers {
+  const merged = new Headers(left)
+
+  if (!right) return merged
+
+  const entries =
+    right instanceof Headers
+      ? right.entries()
+      : Array.isArray(right)
+        ? right
+        : Object.entries(right)
+
+  for (const [key, value] of entries) {
+    merged.set(key, value)
+  }
+
+  return merged
+}
--- a/packages/core/postgrest-js/src/utils.ts
+++ b/packages/core/postgrest-js/src/utils.ts
@@ -7,16 +7,9 @@ export function mergeHeaders(left: Headers, right?: HeadersInit): Headers {
 
   if (!right) return merged
 
-  const entries =
-    right instanceof Headers
-      ? right.entries()
-      : Array.isArray(right)
-        ? right
-        : Object.entries(right)
-
-  for (const [key, value] of entries) {
+  new Headers(right).forEach((value, key) => {
     merged.set(key, value)
-  }
+  })
 
   return merged
 }
--- a/packages/core/postgrest-js/src/PostgrestClient.ts
+++ b/packages/core/postgrest-js/src/PostgrestClient.ts
@@ -2,6 +2,8 @@ import PostgrestQueryBuilder from './PostgrestQueryBuilder'
 import PostgrestFilterBuilder from './PostgrestFilterBuilder'
 import { Fetch, GenericSchema, ClientServerOptions } from './types/common/common'
 import { GetRpcFunctionFilterBuilderByArgs } from './types/common/rpc'
+import { PostgrestQueryBuilderOptions } from './types/types'
+import { mergeHeaders } from './utils'
 
 /**
  * PostgREST client.
@@ -156,12 +158,17 @@ export default class PostgrestClient<
   from<
     TableName extends string & keyof Schema['Tables'],
     Table extends Schema['Tables'][TableName],
-  >(relation: TableName): PostgrestQueryBuilder<ClientOptions, Schema, Table, TableName>
+  >(
+    relation: TableName,
+    options?: PostgrestQueryBuilderOptions
+  ): PostgrestQueryBuilder<ClientOptions, Schema, Table, TableName>
   from<ViewName extends string & keyof Schema['Views'], View extends Schema['Views'][ViewName]>(
-    relation: ViewName
+    relation: ViewName,
+    options?: PostgrestQueryBuilderOptions
   ): PostgrestQueryBuilder<ClientOptions, Schema, View, ViewName>
   from(
-    relation: (string & keyof Schema['Tables']) | (string & keyof Schema['Views'])
+    relation: (string & keyof Schema['Tables']) | (string & keyof Schema['Views']),
+    options?: PostgrestQueryBuilderOptions
   ): PostgrestQueryBuilder<ClientOptions, Schema, any, any> {
     if (!relation || typeof relation !== 'string' || relation.trim() === '') {
       throw new Error('Invalid relation name: relation must be a non-empty string.')
@@ -169,11 +176,11 @@ export default class PostgrestClient<
 
     const url = new URL(`${this.url}/${relation}`)
     return new PostgrestQueryBuilder(url, {
-      headers: new Headers(this.headers),
+      headers: mergeHeaders(this.headers, options?.headers),
       schema: this.schemaName,
-      fetch: this.fetch,
-      urlLengthLimit: this.urlLengthLimit,
-      retry: this.retry,
+      fetch: options?.fetch ?? this.fetch,
+      urlLengthLimit: options?.urlLengthLimit ?? this.urlLengthLimit,
+      retry: options?.retry ?? this.retry,
     })
   }
 
--- a/packages/core/postgrest-js/src/index.ts
+++ b/packages/core/postgrest-js/src/index.ts
@@ -27,6 +27,7 @@ export type {
   PostgrestResponseSuccess,
   PostgrestSingleResponse,
   PostgrestMaybeSingleResponse,
+  PostgrestQueryBuilderOptions,
 } from './types/types'
 export type { ClientServerOptions as PostgrestClientOptions } from './types/common/common'
 // https://github.com/supabase/postgrest-js/issues/551
--- a/packages/core/postgrest-js/src/types/types.ts
+++ b/packages/core/postgrest-js/src/types/types.ts
@@ -55,12 +55,6 @@ export type PostgrestQueryBuilderOptions = {
   retry?: boolean
 }
 
-/** @internal */
-export type PostgrestQueryBuilderOptionsWithSchema<TSchema extends string> =
-  PostgrestQueryBuilderOptions & {
-    schema?: TSchema
-  }
-
 export type DatabaseWithOptions<Database, Options extends ClientServerOptions> = {
   db: Database
   options: Options
--- a/packages/core/postgrest-js/src/types/types.ts
+++ b/packages/core/postgrest-js/src/types/types.ts
@@ -1,7 +1,7 @@
 import PostgrestError from '../PostgrestError'
 import { ContainsNull } from '../select-query-parser/types'
 import { SelectQueryError } from '../select-query-parser/utils'
-import { ClientServerOptions } from './common/common'
+import { ClientServerOptions, Fetch } from './common/common'
 
 /**
  * Response format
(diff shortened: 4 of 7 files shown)


mandarini:
Hi @stenlan ! Thank you very much for your contribution. Can you please import your changes from [this PR](https://github.com/supabase/postgrest-js/pull/648) here, and make the changes needed in the new monorepo? I will be archiving the old repos soon. You will still be able to view the PR content, don't worry.

I understand that this may be frustrating, but thank you very much for your time and contribution!

stenlan:
Thank you @mandarini for the response. I have integrated the changes into the new monorepo structure to the best of my ability, but because of time constraints I have not yet been able to successfully set up the monorepo locally. For some reason, in `packages/core/supabase-js`, it seems to not be able to resolve an entirely unrelated package `@supabase/storage-js`, nor can it resolve the `PostgrestQueryBuilderOptions` that I have added to `@supabase/postgrest-js` anymore, since it seems to be using a different version of this package, even if I set up a local file link to the package. Are there any specific environment setup instructions I should be following?

packages/core/supabase-js/package.json
```diff
@@ -54,21 +54,21 @@
   "dependencies": {
     "@supabase/auth-js": "*",
     "@supabase/functions-js": "*",
+    "@supabase/node-fetch": "2.6.15",
     "@supabase/postgrest-js": "*",
     "@supabase/realtime-js": "*",
-    "@supabase/storage-js": "*",
-    "@supabase/node-fetch": "2.6.15"
```
mandarini: please revert changes to `package.json`

package-lock.json
```diff
@@ -36412,7 +36412,7 @@
         "@supabase/auth-js": "*",
         "@supabase/functions-js": "*",
         "@supabase/node-fetch": "2.6.15",
-        "@supabase/postgrest-js": "*",
```
mandarini: please revert this change

mandarini:
@stenlan thank you for your contribution! Nothing special is needed to set up your local environment. Did you go through our contributing guide? Just clone the repo, run `npm i` and it should work out of the box. Is this not your experience? What errors are you seeing? Can you share some logs?

mandarini:
@stenlan you're right, there were some issues in our lock file. Please pull, delete all your local node_modules and dist folders, and run `npm i` again on the root of the repo! Should be fixed now!

steve-chavez:
What is the outcome of this PR? Does it provide an interface like the one mentioned on 

> supabase
  .withFetch((...args) => fetch(args[0], { 
    ...args[1], 
    cache: 'force-cache', 
    next: { revalidate: 900 } 
  }))
  .select()
  // ...
https://github.com/supabase/supabase-js/issues/438#issuecomment-3117530326?

stenlan:
I will get back to this as soon as I have the time, maybe this weekend. I'm pretty busy at the moment, my apologies.

stenlan:
@mandarini should be all good now, sorry about the *slight* delay 😅. This time I'm getting typescript errors on my side, but I think that's just my local env.

mandarini (changes requested):
Hi @stenlan, thank you so much for sticking with this one through the monorepo migration, the lockfile mess, and the long gaps in between. That is a lot of friction for a contribution and I really appreciate you seeing it through. Also, apologies for taking so long to review.

I did a full review of the current state of the PR and there are a few things that need to change before this can merge, plus one open question from @steve-chavez that we should settle.

**Things that need fixing:**

1. `PostgrestQueryBuilderOptions` includes `retry` and `urlLengthLimit`, but `PostgrestClient.from` never actually reads them from `options`, it always uses `this.retry` and `this.urlLengthLimit` instead. Right now calling `.from('users', { retry: false })` silently does nothing. Either wire these through or drop them from the type so it doesn't promise something it doesn't do.

2. `SupabaseClient.from` mutates the caller's `options` object directly (`options.fetch = this._createFetchWithAuth(options.fetch)`). If someone reuses the same options object across multiple `.from()` calls in one request, which is a very likely pattern for the SSR use case this is meant for, the fetch gets wrapped again on every call. Please build a new object instead of assigning back onto the caller's `options`.

3. This is the important one: `_createFetchWithAuth` uses `this._getAccessToken.bind(this)`, but master has since moved the default client fetch over to `this._getSessionToken.bind(this)`. That change exists specifically to stop new-format API keys (`sb_publishable_...` / `sb_secret_...`) from ever being sent as a Bearer token, which `_getAccessToken`'s old fallback would do. As written, a per-request fetch override would bypass that protection. Please rebase `_createFetchWithAuth` against the current `fetchWithAuth`/`_getSessionToken` in `lib/fetch.ts` and `SupabaseClient.ts` rather than resolving this as a plain merge conflict, the auth logic underneath has changed since this PR was opened.

4. No tests were added. At minimum we'd want coverage for: the per-request fetch actually being used instead of the client default, header merge precedence (a per-request header overriding a client-level one of the same name), and confirming auth headers are still injected when a custom fetch is passed.

5. The package.json / package-lock.json changes flagged in the earlier review are still there, please revert those.

**On the API shape:** @steve-chavez asked whether this delivers the chainable `.withFetch()` interface floated in the issue thread. After going through it, we'd like to stick with the `.from(relation, options)` shape from this PR rather than switch to a chainable method, since the headers/fetch merging is already built out here and reworking it into a chainable form would be a bigger change for not much benefit. So no need to change direction, just the fixes above.

No rush given everything else going on, just wanted to get you a clear list so the next pass can land. Thanks again for the persistence on this one.

stenlan:
Hi @mandarini , thanks for the in-depth review.

**\#1 - PostgrestQueryBuilderOptions includes retry and urlLengthLimit, but PostgrestClient.from never actually reads them from options**
Fixed, it now passes these options as well

**\#2 - SupabaseClient.from mutates the caller's options object directly (options.fetch = this._createFetchWithAuth(options.fetch)).**
Fixed, it now creates a new `options` object in this case.

**\#3 - _getAccessToken ->_getSessionToken refactor**
Done, _createFetchWithAuth now uses _getSessionToken instead. `this.functionsFetch`'s initialization logic has also been abstracted into `_createFetchWithAuth` and `_createFetchWithAuth` has received an additional optional parameter to accommodate this.

**\#4 - Tests**
I have attempted to write some tests for the situations you describe. Their general structure should be okay, but I cannot for the life of me get the testing environment properly set up. I get missing/invalid schema errors, typescript throwing out Jest errors, Docker errors, and even an error on the `mergeHeaders` function when running Postgrest tests. I have tried WSL and read through all your setup docs but things still keep failing. Unfortunately, I am not willing to spend any more time on this PR in that regard. If someone else (like @7vignesh) would be willing to pick those up, feel free. Otherwise, feel free to close this PR - possibly in favor of theirs; I saw them mention a different approach using `.withFetch()` instead which also seems like a good one.

**\#5 - package.json / package-lock.json changes**
I have reverted those.

7vignesh:
Hi @mandarini, I'd be happy to pick up the remaining work here (mainly the tests and any final fixes). Since @stenlan mentioned they're unable to get the test environment working and offered for someone else to continue, would you prefer I:

1. Open a new PR based on stenlan's changes with the tests added and any remaining fixes applied
2. Push commits directly to this branch (if stenlan has maintainer edits enabled)

Either way I'll make sure to credit stenlan for the original implementation. Just let me know which approach works best and I'll get started.

cc @stenlan hope that's okay with you!

stenlan:
> cc @stenlan hope that's okay with you!

@7vignesh 

Of course. No credits or anything needed, thanks for the assistance. Let me know if you need maintainer access to my repo or else feel free to fork or just copy it as well, based on the approach you decide on.

7vignesh:
@stenlan that would be great! Please add me as a collaborator on your fork so I can push directly to this branch. My GitHub username is 7vignesh. I'll add the tests and fix anything remaining.

7vignesh:
Hi @mandarini @stenlan, I've reimplemented the feature from scratch on latest master and force-pushed to this branch. All of mandarini's review points are addressed:

1. retry and urlLengthLimit are wired through (per-request options override client defaults)
2. Caller's options object is never mutated (a new object is created when wrapping fetch)
3. Auth logic uses the current _getSessionToken + fetchWithAuth path, no _getAccessToken
4. Tests added: custom fetch usage, header precedence, auth header injection, retry override, and non-mutation of caller options
5. No unrelated package.json/lock changes

The implementation is a single clean commit on master. All 28 postgrest-js type tests pass, supabase-js unit tests pass, and both packages build successfully.

Thanks @stenlan for the original implementation, I just cleaned it up for the current master state.

7vignesh:
@mandarini @stenlan Quick note for the merge commit: the latest push (reimplementation on current master, tests, conflict resolution) was done by me. Would appreciate a co-author credit in the squash merge message if possible:

Co-authored-by: 7vignesh <vigneshgopikrishnan7@gmail.com>

stenlan:
Hey @mandarini not trying to push at all (since I barely have time to invest in this myself anyway), just curious on the status of things here? Anything that @7vignesh or I could do to help this merge along? Any steps that are required still?

mandarini:
@stenlan sorry I have been busy with some other things the past few weeks and let this sit. I will try to fix this PR myself and merge it

mandarini:
@stenlan @7vignesh I am pushing some fixes, and I requested a review from the postgrest team, just to be sure. if they give the thumbs up, I will merge this

mandarini (dismissed):

packages/core/postgrest-js/src/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+  const merged = new Headers(left)
+
+  if (!right) return merged
+
+  new Headers(right).forEach((value, key) => {
+    merged.set(key, value)
+  })
+
+  return merged
+}
```
steve-chavez: If the PR is about passing a custom fetch object, why do we need this new work here?

avallete: Agree

`mergeHeaders` is only for the extra `headers` option on `.from()`, not for custom fetch. Fetch is swapped on the next line, this copies the client header map and lets per-request keys win so you don’t drop `apikey` / `Authorization`.

That’s a separate feature from #438 though. I’d drop `headers` + `mergeHeaders` from this PR and keep `.from(relation, { fetch })` (and `retry` / `urlLengthLimit` if we want those). You can already set headers inside a custom fetch. We can add a dedicated overlay later if a caller actually needs it.

avallete (commented):
LGTM beside steve thread.

packages/core/postgrest-js/src/utils.ts
mandarini: Addressed in last commit! Thanks for review

steve-chavez (approved):

PR #1582, opened 2025-10-02
feat(auth): add OAuth 2.1 client admin endpoints

cemalkilic:
## Summary
Adds OAuth 2.1 client administration endpoints to `@supabase/auth-js` based on the OpenAPI specification from https://github.com/supabase/auth/pull/2165.

This PR implements a new `admin.oauth` namespace with full CRUD operations for managing OAuth clients when the OAuth 2.1 server is enabled in Supabase Auth

New admin.oauth methods:
  - listClients() - List OAuth clients with pagination (for now, the same api with `listUsers`)
  - createClient() - Register new OAuth client
  - getClient() - Get client details by ID
  - deleteClient() - Remove OAuth client
  - regenerateClientSecret() - Regenerate client secret

All methods include proper error handling, follow existing patterns (similar to admin.mfa), and are only relevant when the OAuth 2.1 server is enabled in Supabase Auth.

### DIFF
--- a/packages/core/auth-js/src/GoTrueAdminApi.ts
+++ b/packages/core/auth-js/src/GoTrueAdminApi.ts
@@ -21,12 +21,22 @@ import {
   PageParams,
   SIGN_OUT_SCOPES,
   SignOutScope,
+  GoTrueAdminOAuthApi,
+  CreateOAuthClientParams,
+  OAuthClientResponse,
+  OAuthClientListResponse,
 } from './lib/types'
 import { AuthError, isAuthError } from './lib/errors'
 
 export default class GoTrueAdminApi {
   /** Contains all MFA administration methods. */
   mfa: GoTrueAdminMFAApi
+
+  /**
+   * Contains all OAuth client administration methods.
+   * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+   */
+  oauth: GoTrueAdminOAuthApi
 
   protected url: string
   protected headers: {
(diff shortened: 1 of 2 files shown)


mandarini:
For PR release, try this URL for now:

```bash
npm i https://pkg.pr.new/supabase/supabase-js/@supabase/auth-js@5acdc8e
```

Once we have a `latest` release on `npm` from this repo, normal URLs will work!

packages/core/auth-js/src/lib/types.ts
```diff
@@ -1448,3 +1448,145 @@ export interface JWK {
...
+  | {
+      data: { clients: [] }
+      error: AuthError
+    }
+
+/**
+ * Contains all OAuth client administration methods.
+ * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+ */
+export interface GoTrueAdminOAuthApi {
```
fadymak, line 1552: nit: do we want to continue using the GoTrue name? Feel free to ignore if we do this elsewhere!

packages/core/auth-js/src/GoTrueAdminApi.ts
```diff
@@ -349,4 +366,150 @@ export default class GoTrueAdminApi {
...
+
+  /**
+   * Gets details of a specific OAuth client.
+   * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+   *
+   * This function should only be called on a server. Never expose your `service_role` key in the browser.
+   */
+  private async _getOAuthClient(clientId: string): Promise<OAuthClientResponse> {
+    try {
+      return await _request(this.fetch, 'GET', `${this.url}/admin/oauth/clients/${clientId}`, {
```
fadymak, line 445: Not sure if we want to validate the client ID as a UUID, similar to what we do here: https://github.com/supabase/supabase-js/blob/master/packages/core/auth-js/src/GoTrueAdminApi.ts#L303C5-L303C17

grdsdev: good catch, if client ID is a UUID then it should be validated.

packages/core/auth-js/src/lib/types.ts
mandarini: I think we use it in other places as well. Let's leave as is for now.

PR #1627, opened 2025-10-06
chore(postgrest): setup prettier for jest snapshot update

avallete:
## 🔍 Description

Jest inline snapshot require prettier <3.x when updating use compatible version for test snapshot update, keep the monorepo version for code formatting.

This occured when migrating to an unique prettier version when setting up monorepo, since the error only happen when updating the snapshots it went under the radar.

### What changed?

- Set back `postgrest-js` specific prettier < 3 and configure jest to use it for `inlineSnapshot` update
- Update `format` in favor of a node script to keep all the codebase in sync with the monorepo prettier version.

### To test:

Before:
  - Go to `packages/core/postgrest-js/test` change a line in one of the inline snapshot, run `npm run test:update` see the error
After:
   - The snapshot update back to the expected value.

### DIFF
--- a/packages/core/postgrest-js/scripts/format.js
+++ b/packages/core/postgrest-js/scripts/format.js
@@ -8,12 +8,12 @@ const monorepoRoot = path.resolve(__dirname, '../../../../')
 
 // Run prettier from the monorepo root
 const command =
-    process.argv[2] === 'check'
-        ? 'npx prettier --check "packages/core/postgrest-js/**/*{ts,mjs,json,yml,yaml}"'
-        : 'npx prettier --write "packages/core/postgrest-js/**/*{ts,mjs,json,yml,yaml}"'
+  process.argv[2] === 'check'
+    ? 'npx prettier --ignore-path packages/core/postgrest-js/.gitignore --check "packages/core/postgrest-js/**/*{ts,js,mjs,json,yml,yaml}"'
+    : 'npx prettier --ignore-path packages/core/postgrest-js/.gitignore --write "packages/core/postgrest-js/**/*{ts,js,mjs,json,yml,yaml}"'
 
 try {
-    execSync(command, { cwd: monorepoRoot, stdio: 'inherit' })
+  execSync(command, { cwd: monorepoRoot, stdio: 'inherit' })
 } catch (error) {
-    process.exit(error.status || 1)
+  process.exit(error.status || 1)
 }
--- a/packages/core/postgrest-js/jest.config.ts
+++ b/packages/core/postgrest-js/jest.config.ts
@@ -1,4 +1,6 @@
-module.exports = {
+import type { Config } from '@jest/types'
+
+const config: Config.InitialOptions = {
   preset: 'ts-jest',
   testEnvironment: 'node',
   collectCoverageFrom: ['src/**/*', '!src/types.ts'],
@@ -16,3 +18,4 @@ module.exports = {
     },
   },
 }
+export default config


packages/core/postgrest-js/package.json
```diff
@@ -67,6 +67,7 @@
     "tstyche": "^4.3.0",
     "type-fest": "^4.32.0",
     "wait-for-localhost-cli": "^3.0.0",
-    "zod": "^3.25.76"
+    "zod": "^3.25.76",
+    "prettier": "^2.6.2"
```
mandarini, line 71: Why older version of `prettier`? We have this in the root: https://github.com/supabase/supabase-js/blob/master/package.json#L53

packages/core/postgrest-js/scripts/format.js
```diff
@@ -0,0 +1,19 @@
...
+const { execSync } = require('child_process')
+const path = require('path')
+
+// Get the monorepo root directory (go up from scripts/ to postgrest-js/ to core/ to packages/ to root)
+const monorepoRoot = path.resolve(__dirname, '../../../../')
+
+// Run prettier from the monorepo root
+const command =
+  process.argv[2] === 'check'
+    ? 'npx prettier --ignore-path packages/core/postgrest-js/.gitignore --check "packages/core/postgrest-js/**/*{ts,js,mjs,json,yml,yaml}"'
```
mandarini, line 12: You can run `npx nx format` technically, but if this works better for the `postgrest` use case, let's leave as is, and I can investigate at a later time!

packages/core/postgrest-js/package.json
avallete: As I mentioned in the description, the bug occurs when trying to update `inlineSnapshot` with `jest < 30.xx` and `prettier > 2.XX`. In such case you get an error, so this older version is here to allow the `require.resolve` within `jest.config.ts` to resolve a version that can be used by jest to update inline snapshots.

See: https://github.com/jestjs/jest/issues/14305

Other way to go would be to migrate to jest latest >30. I haven't tried it yet as I have other pending work using current jest major. Once this is passed though, we could migrate the tests toward latest `jest` (or even better `vitest`) then, all this specifics won't be required anymore.

mandarini: Cool! I'll add a task for me to update to jest 29 at some point!

packages/core/postgrest-js/scripts/format.js
avallete: Tried this, but when ran from the `packages/core/postgrest-js` context, the command fail with this error:

```
➜  postgrest-js git:(chore/upgrade-jest-to-latest-for-snapshot) ✗ npx nx format --all                             
➜  postgrest-js git:(chore/upgrade-jest-to-latest-for-snapshot) ✗ npx nx format postgrest-js
[error] No files matching the pattern were found: "nx.json".
[error] No files matching the pattern were found: "tsconfig.base.json".
nx format:write

Overwrite un-formatted files.

Run command using --base=[SHA1] (affected by the committed, uncommitted and untracked changes):
...
Error: Command failed: node "xxx/supabase-js/node_modules/prettier/bin/prettier.cjs" --write --list-different nx.json tsconfig.base.json
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at checkExecSyncError (node:child_process:882:11)
    at execSync (node:child_process:954:15)
    at write (xxx/supabase-js/node_modules/nx/src/command-line/format/format.js:149:43)
    at xxx/supabase-js/node_modules/nx/src/command-line/format/format.js:46:42
    at Array.forEach (<anonymous>)
    at Object.format (xxx/supabase-js/node_modules/nx/src/command-line/format/format.js:46:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Object.handler (xxx/supabase-js/node_modules/nx/src/command-line/format/command-object.js:21:9) {
  status: 2,
  signal: null,
  output: [ null, null, null ],
  pid: 72406,
  stdout: null,
  stderr: null
}
```

Note that if I mention the `--all` flag, then things works properly. It also works if I use `npx nx format:check postgrest-js`.

PR #1633, opened 2025-10-07
chore(repo): add fixed version of node with nvmrc for consistent install

avallete:


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
mandarini: maybe better to use `20`? Just to make sure no ambiguation. other than that, i think that's a solid feat we should add!

mandarini (approved):

PR #1736, opened 2025-10-07
refactor(supabase): reduce codedup between PostgrestClient and Supaba…

avallete:
…seClient leverage inference

We currently duplicate a lot of the typing logic between postgrest-js and supabase-js This makes it very hard to change anything type-wise in postgrest-js without needing to update supabase-js as well. This is mainly because we re-declare functions type returns instead of relying over return type inference.

With this change, as long as the arguments passed to a rest function don't change, the result type will be infered from the this.rest call.

This will allows postgrest-js to alter the result types / implementation details without having to duplicate all changes over to supabase-js.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -184,6 +184,7 @@ export default class SupabaseClient<
    *
    * @param relation - The table or view name to query
    */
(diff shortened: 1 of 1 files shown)


avallete:
Closed in favor of: https://github.com/supabase/supabase-js/pull/1632/commits/36b4cd509b7d832e4fc0faabc71ee0c782882d44

PR #1737, opened 2025-10-07
chore(ci): generate token for supabase repo

mandarini:
Generating token for both `supabase` and `supabase-js` repos, so that it can authenticate and trigger the workflow on the `supabase/supabase` repo.

This will not work yet, because:
1. The GitHub App needs to be installed on the `supabase/supabase` repository.
2. In the App settings, under "Repository permissions":
    - Workflows to Read and write
    - Actions to Read and write as well (optional but recommended by the github assistant)


grdsdev (approved):

staaldraad (approved):
🤞🏼

PR #1749, opened 2025-10-08
fix(realtime): realtime explicit REST call

filipecabaco:
## 🔍 Description

Added new public method for explicit usage of REST API for broadcast 
### What changed?

New method `httpSend` added

Our current approach leads users to use broadcast without knowing.

## 🔄 Breaking changes

- [ ] New log warning to inform that users are using REST when using `send` while not connected.

## 📋 Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md)
- [X] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `<type>(<scope>): <description>`
- [X] I have run `npx nx format` to ensure consistent code formatting
- [X] I have added tests for new functionality (if applicable)

### DIFF
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -436,6 +436,68 @@ export default class RealtimeChannel {
     return this._on(type, filter, callback)
   }
   /**
+   * Sends a broadcast message explicitly via REST API.
+   *
+   * This method always uses the REST API endpoint regardless of WebSocket connection state.
+   * Useful when you want to guarantee REST delivery or when gradually migrating from implicit REST fallback.
+   *
+   * @param event The name of the broadcast event
+   * @param payload Payload to be sent (required)
+   * @param opts Options including timeout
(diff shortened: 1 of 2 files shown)


packages/core/realtime-js/src/RealtimeChannel.ts
```diff
@@ -435,6 +435,68 @@ export default class RealtimeChannel {
...
+   * @param opts Options including timeout
+   * @returns Promise resolving to object with success status, and error details if failed
+   */
+  async postSend(
+    event: string,
+    payload: any,
+    opts: { timeout?: number } = {}
+  ): Promise<{ success: true } | { success: false; status: number; error: string }> {
+    if (!this.socket.accessTokenValue) {
+      return Promise.reject('Access token is required for postSend()')
```
edgurgel, line 455: What if the secret key is being used? Would we still need an access token?

filipecabaco:
https://github.com/supabase/supabase-js/pull/1751 as we need this to be a minor version change

Issue #1704, opened 2024-05-15
PKCE flow issue with other than supabase `code` query in URL


vachmara:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I am using the @nuxtjs/supabase package and I encounter a bug described in this [issue](https://github.com/nuxt-modules/supabase/issues/289).

I am not able to use other PKCE flows because each time the third app redirects to my main app, GoTrueClient tries to refresh the session with the incorrect code parameter in the URL despite using `detectSessionInUrl` at initialization of GoTruClient.

I believe this function [_isPKCEFlow](https://github.com/supabase/auth-js/blob/4ecfdda65188b71322753e57622be8eafe97ed6b/src/GoTrueClient.ts#L1537) should only watch specific URLs to manage other PKCE flows.

## To Reproduce

1. Setup a project with nuxt/supabase.
2. Build a simple authentification system.
3. On any page, use a query parameter `?code=random`.

## Expected behavior

Automatically, the GoTrueClient will try to set up a session at initialization and logout current user which is problematic.

## Screenshots

![image](https://github.com/supabase/supabase-js/assets/55046446/23423708-18e6-409d-8cd2-265b3bd8459c)

Issue #1697, opened 2024-09-24
GoTrueClient intercepts non-Supabase initiated OAuth preventing manual login flows


uncvrd:
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Hi there, I've run in to an issue where Supabase Auth is intercepting url params/fragments that contain the keyword `access_token` in them even when auth requests aren't coming from a Supabase Auth initiated request.

For example, I use Supabase Auth for OAuth Login for social providers so I need to have `detectSessionInUrl` enabled. However, in my application I have a page for users to authenticate through Facebook Login to connect their Facebook Pages and Instagram Accounts.

I initiate this flow manually (outside of Supabase) using: https://developers.facebook.com/docs/instagram-platform/instagram-api-with-facebook-login/business-login-for-instagram

However the callback URL returns a string that looks like this:

`https://my-clever-redirect-url.com/success/?#access_token=EAAHm...&data_access_expiration_time=1658889585&expires_in=4815&long_lived_token=ABAEs...`

As you can see there is an `access_token` fragment. And if the user is logged in via Supabase already, this automatically logs the user out since it detects an `access_token` being returned in the URL, which is invalid and Supabase Auth uses this as a trigger to log the user out...

This issue occurs due to the check for implicitGrant here: https://github.com/supabase/auth-js/blob/8a1ec0602792191bd235d51fd45c0ec2cabdf216/src/GoTrueClient.ts#L311C31-L311C49

Where it checks if there are params for either `(params.access_token || params.error_description)`

Is there a way to prevent running this check on certain pages (like my redirect page) so there isn't this kind of conflict?

To prevent this, I thought I could just proxy my response through a server and rewrite the url to not interfere with Supabase but since this value is passed as a URL fragment, this data is not sent to a server and is only accessible in the browser...

## Expected behavior

I expect to be able to use OAuth flow that isn't triggered by Supabase Auth without there being a conflict and Supabase intercepting non Supabase initiated Auth flows.

## Screenshots

<img width="1139" alt="image" src="https://github.com/user-attachments/assets/5ce7bd2f-b34c-41ec-b8c1-ce650fc98fd9">

## System information

- OS: MacOS
- Browser chrome
- Version of supabase-js: 2.43.5
[...]

PR #1759, opened 2025-10-09
fix(auth): code verifier remains in storage during edge cases

mandarini:
Moved from: https://github.com/supabase/auth-js/pull/986

Author: @j4w8n 

Bug fix

In some scenarios, the `sb-<project-id>-auth-token-code-verifier` will get created in storage, but is either not used or an error will happen during calls like `updateUser()`, etc. In these cases, the code verifier remains in storage, which can cause issues if ~~third-party code~~ any non-Supabase code uses the `code` search param in a url later on. See issue https://github.com/supabase/supabase-js/issues/1704 for an example.

This requires a client-side supabase client to be running on the page that recieves the `code` param.

## What is the new behavior?

When the code verifier is set, and the function throws an error or returns an error (in one case), we now remove the code verifier from storage. To cover success scenarios, we remove any potential code verifier whenever a session is saved; as I wasn't sure where else to handle this.

This should resolve the issue when strictly using supabase-js, but this is not necessarily the case for the ssr library because it only enforces storage changes when a whitelisted auth event also occurs (e.g. SIGNED_IN, TOKEN_REFRESHED). Since none of the existing events seem relevant in this case, I created a new one called `STORAGE_UPDATED`, although this might be too generic. I tried to ensure that the session is passed along with the event; so we hopefully don't disrupt any action logic within `onAuthStateChange()` that a lot of devs put there. The ssr PR that coincides with this is https://github.com/supabase/ssr/pull/82.

I'm not sure if this is how the team wants to resolve this issue or not. Feel free to delete this PR or slice and dice it; as it didn't spend too much time on it.

UPDATE: Noting this PR will no longer resolve this issue for users of the ssr library. I've also closed the related ssr PR.

## Alternatives Considered

After some short discussion with @silentworks, one thing which could be done is renaming `code`; which would be breaking of course.

One other idea is to somehow introduce a new name while keeping the old name. Might be a bit complicated, even if it is possible. Then, deprecate the old usage and remove in auth-js v3.

Or, you could implement something like I've proposed here, as a temporary fix, then deprecate `code` in favor of something else in auth-js v3, and then roll my proposed changes back. Although, if one of the goals is to resolve the "orphaned" code verifier, then you may want to keep some of the changes.

Also, to make my proposed change a bit smaller, perhaps the ssr library can be tweaked to where storage changes are written immediately, instead of waiting for an auth state change event. That would allow you to remove the `STORAGE_UPDATED` event proposed here.

## Additional context

fixes supabase/supabase-js#1704
fixes supabase/supabase-js#1697

https://discord.com/channels/839993398554656828/1312930639066169365

As seen here, https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L1535, the `_isPKCEFlow()` check only returns `true` if there is a `code` search param in the url and the code verifier is in storage. So, if we can cleanup the code verifier once it's no longer needed in these edge cases, this resolves the issue.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -635,6 +635,7 @@ export default class GoTrueClient {
       const { data, error } = res
 
       if (error || !data) {
+        await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
         return this._returnResult({ data: { user: null, session: null }, error: error })
       }
 
@@ -648,6 +649,7 @@ export default class GoTrueClient {
 
       return this._returnResult({ data: { user, session }, error: null })
     } catch (error) {
+      await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
       if (isAuthError(error)) {
         return this._returnResult({ data: { user: null, session: null }, error })
       }
@@ -1138,13 +1140,13 @@ export default class GoTrueClient {
       }
       return this._returnResult({ data: { ...data, redirectType: redirectType ?? null }, error })
     } catch (error) {
+      await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
       if (isAuthError(error)) {
         return this._returnResult({
           data: { user: null, session: null, redirectType: null },
           error,
         })
       }
-
       throw error
     }
   }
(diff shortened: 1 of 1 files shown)


uncvrd:
just want to say thanks a lot for working on this

vachmara:
thx ❤️

Issue #1688, opened 2025-01-14
Support throwing errors instead of returning them


soedirgo:
# Feature request

## Describe the feature

Add a `throwOnError` client option to have errors be thrown instead of returning it. As an example, right now we do:

```ts
const auth = new AuthClient({ url: AUTH_URL })
const { data, error } = await auth.signUp({ email: 'myemail@example.com', password: 'foobar' })
```

But with `throwOnError` we do:

```ts
const auth = new AuthClient({ url: AUTH_URL, throwOnError: true })
const { data } = await auth.signUp({ email: 'myemail@example.com', password: 'foobar' })
```

## Additional context

Follows https://github.com/supabase/postgrest-js/pull/188 & https://github.com/supabase/postgrest-js/pull/248

PR #1766, opened 2025-10-10
feat(auth): support throwing errors instead of returning them

mandarini:
Moved from: https://github.com/supabase/auth-js/pull/1022
Author: @kiwicopple 

Fixes supabase/supabase-js#1688

Add support for `throwOnError` option in `AuthClient` to throw errors instead of returning them.

* **GoTrueClient Changes:**
  - Add `throwOnError` option to `GoTrueClient` constructor.
  - Modify methods to throw errors when `throwOnError` is true.
  - Update methods like `signUp`, `signInWithPassword`, `updateUser`, and others to throw errors if `throwOnError` is true.

* **AuthClient Changes:**
  - Update `AuthClient` to pass `throwOnError` option to `GoTrueClient`.

* **Tests:**
  - Add tests in `test/GoTrueClient.test.ts` to verify `throwOnError` functionality.

### `ConstructorParameters`

Replace `ConstructorParameters` type extraction with direct import of `GoTrueClientOptions`:

```typescript
// Before
type AuthClientOptions = ConstructorParameters<typeof AuthClient>[0]
export interface SupabaseAuthClientOptions extends AuthClientOptions {}

// After
import { GoTrueClientOptions } from '@supabase/auth-js'
export interface SupabaseAuthClientOptions extends GoTrueClientOptions {}
```

This was introduced because `ConstructorParameters` wasn't resolving properly across package boundaries. Direct type extension is more explicit and ensures all auth-js properties are recognized.

### Other

Also fix two test failures in the throwOnError feature:

1. Fix window is not defined error in Node.js tests

**Issue**: `_initialize()` in `GoTrueClient.ts` accessed `window.location.href` without checking if running in a browser environment, causing `Node.js` tests to fail with `ReferenceError: window is not defined`.

**Root Cause**: This bug was introduced in commit f9775cfa but wasn't caught until this PR added new tests that instantiate `SupabaseClient` directly in Node.js environment.

**Fix**: Wrap URL parsing in an `isBrowser()` check, consistent with how browser APIs are accessed elsewhere in the codebase.

```typescript
// Before
const params = parseParametersFromURL(window.location.href)

// After
let params: { [parameter: string]: string } = {}
 if (isBrowser()) {
   params = parseParametersFromURL(window.location.href)
   // ... check for callback types
}
```

2. Fix incorrect test expectations for `signInWithSSO()`

**Issue**: Test `signInWithSSO() should return error shape; calling code can opt to throw` expected the error to be returned instead of thrown, causing test failure.

**Root Cause**: The test was incorrectly written when added in commit 002886ac. The test name suggested alternative behavior, but this contradicts the feature design and all other tests in the suite.

**Why this is a test bug, not a code bug:**

1. **Feature Design**: `throwOnError` is a client-level setting. When `throwOnError: true`, ALL methods throw on error consistently via the `_returnResult()` helper.

2. **Test Suite Consistency**: All other tests in the `GoTrueClient with throwOnError option` suite expect throws:
  - `signUp()` expects throw
  - `signInWithPassword()` expects throw
  - `updateUser()` expects throw
  - `signInWithOtp()` expects throw
  - `signInWithSSO()` expected return (incorrect)
  - 
3. **Implementation Consistency**: `signInWithSSO()` uses `_returnResult()` just like all other methods - there's no special case handling.

**Fix**: Update test to expect throwing behavior and rename to accurately reflect what it tests.

```typescript
// Before
test('signInWithSSO() should return error shape; calling code can opt to throw',
  async () => {
    const res = await client.signInWithSSO({ domain: 'nonexistent.example.com' })
    expect(res.error).not.toBeNull()
})

// After  
test('signInWithSSO() should throw on error when throwOnError is true', async ()
  => {
    await expect(client.signInWithSSO({ domain: 'nonexistent.example.com'
  })).rejects.toThrow()
  })
```

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -121,6 +121,7 @@
         "npm:eslint@^8.57.1",
         "npm:faker@^5.3.1",
         "npm:husky@^9.1.7",
+        "npm:jest-environment-jsdom@^29.5.0",
         "npm:jest-mock-server@0.1",
         "npm:jest@^29.7.0",
         "npm:jiti@2.4.2",
@@ -144,8 +145,7 @@
         "packageJson": {
           "dependencies": [
             "npm:@supabase/node-fetch@2.6.15",
-            "npm:jest@^28.1.3",
-            "npm:ts-jest@^28.0.7",
+            "npm:prettier@^2.8.8",
             "npm:tslib@2.8.1"
           ]
         }
@@ -167,10 +167,8 @@
           "dependencies": [
             "npm:@supabase/node-fetch@2.6.15",
             "npm:chokidar-cli@3",
-            "npm:jest@^28.1.0",
             "npm:node-abort-controller@^3.0.1",
             "npm:prettier@^2.6.2",
-            "npm:ts-jest@^28.0.3",
             "npm:tslib@2.8.1",
             "npm:tstyche@^4.3.0",
             "npm:type-fest@^4.32.0",
@@ -203,7 +201,6 @@
           "dependencies": [
             "npm:@supabase/node-fetch@2.6.15",
             "npm:form-data@4",
-            "npm:jest-environment-jsdom@^29.5.0",
             "npm:ts-loader@^9.4.2",
             "npm:tslib@2.8.1",
             "npm:webpack-cli@^5.0.1",
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -438,12 +438,16 @@ export default class GoTrueClient {
    */
   private async _initialize(): Promise<InitializeResult> {
     try {
-      const params = parseParametersFromURL(window.location.href)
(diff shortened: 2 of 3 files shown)


packages/core/auth-js/src/lib/types.ts
```diff
@@ -107,6 +107,11 @@ export type GoTrueClientOptions = {
    * @experimental
    */
   hasCustomAuthorizationHeader?: boolean
+  /**
+   * If there is an error with the query, throwOnError will reject the promise by
+   * throwing the error instead of returning it as part of a successful response.
+   */
+  throwOnError?: boolean
```
hf, line 114: Don't forget this needs to be propagated all the way to the supabase-js createClient options

mandarini: Done

grdsdev (approved):

PR #1769, opened 2025-10-13
fix(realtime): manipulate URLs using URL object

grdsdev:
## 🔍 Description

Change `httpEndpointURL` to manipulate url using URL JavaScript type.

### What changed?

No behavior change, just use correct types for url manipulation instead of string-based manipulation.

To avoid possible security issues with url manipulation.

## 🔄 Breaking changes

## 📋 Checklist

### DIFF
--- a/packages/core/realtime-js/src/lib/transformers.ts
+++ b/packages/core/realtime-js/src/lib/transformers.ts
@@ -251,8 +251,21 @@ export const toTimestampString = (value: RecordValue): RecordValue => {
 }
 
 export const httpEndpointURL = (socketUrl: string): string => {
-  let url = socketUrl
-  url = url.replace(/^ws/i, 'http')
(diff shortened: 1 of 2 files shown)


grdsdev:
> Nit: I think that `wsUrl.pathname === ''` `pathname` here will never be an empty string, but no hard done in leaving this check there!

Why it can't be empty @mandarini ?

mandarini:
> Why it can't be empty @mandarini ?

As mentioned in the [`pathname` spec](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname), URLs with hierarchical schemes (`ws` URLs follow a hierarchical scheme) "_always have at least one (invisible) path segment: the empty string. The pathname value for such URLs will therefore always have at least one `/` character._".

We can leave the empty check though in any case!

PR #1752, opened 2025-10-09
feat(storage): storage vectors and analytics in storage-js

mandarini:
## Description

Adds vector storage and analytics capabilities to `@supabase/storage-js`, enabling management of vector buckets for AI/ML workloads and analytics data.

## Key Changes

- **Vector Bucket API**: New `VectorBucketApi` class for creating, managing, and querying vector storage buckets
- **Analytics Integration**: Analytics bucket support with custom endpoints
- **Documentation**: Added comprehensive TypeDoc documentation across all new modules

## Testing

- [x] Unit tests added for VectorBucketApi and analytics functionality
- [x] Integration tests passing

@fenos

PR #1777, opened 2025-10-15
chore(repo): configure TypeScript project references

mandarini:
### TypeScript Project References
- Add `composite: true` and `declaration: true` to tsconfig.base.json for incremental builds
- Configure project references in root tsconfig.json to link all packages
- Update core packages (auth-js, realtime-js, postgrest-js, functions-js, storage-js) to:
    - Override `moduleResolution: "Node"` for CommonJS compatibility
    - Override `isolatedModules: false` to avoid requiring `export type` for type re-exports
    - Override `noImplicitOverride: false` for existing code compatibility
    - Add project references pointing to dependencies (managed by `nx sync`)

### Nx Local Caching
- Enable caching for all build targets (`build`, `build:main`, `build:module`, `build:umd`, `webpack:build`)
- Enable caching for all test targets (`test`, `test:unit`, `test:coverage`, `test:integration`, `test:ci`, `vite:test`)
- Enable caching for code quality targets (`lint`, `format`, `typecheck`)
- Enable caching for documentation targets (`docs`, `docs:json`)
- Configure proper output paths for cache restoration:
    - Build outputs: `{projectRoot}/dist`, `.tsbuildinfo` files
    - Test outputs: `{projectRoot}/coverage`
    - Docs outputs: `{projectRoot}/docs`

###  PostgreSQL Docker Image Optimization

- Replace `supabase/postgres:15.1.0.37` (1.184GB) with `postgis/postgis:15-3.4-alpine` (~300MB) for `postgrest-js` tests
  - 74% size reduction eliminates CI timeout failures caused by slow Docker image downloads
  - Audited test database schema and confirmed only PostGIS extension is needed (for .geojson() test with geometry columns)
  - The Supabase Postgres image includes 20+ extensions (pgsodium, pgvector, pg_graphql, etc.) that are not tested by postgrest-js
  - PostgREST is an open-source project independent of Supabase - tests only require standard PostgreSQL 15 + PostGIS
  - Added compute_query_id=on configuration to maintain PostgreSQL EXPLAIN test compatibility
  - Updated test snapshots to reflect minor differences in PostgreSQL default settings between images

## Breaking Changes

**None** - This is purely a build optimization:
- No changes to public APIs
- No changes to build outputs
- No changes to package.json dependencies
- Fully backward compatible

### DIFF
--- a/packages/core/postgrest-js/test/db/docker-compose.yml
+++ b/packages/core/postgrest-js/test/db/docker-compose.yml
@@ -29,7 +29,7 @@ services:
     depends_on:
       - db
   db:
-    image: postgis/postgis:15-3.4-alpine
+    image: supabase/postgres:15.1.0.37
     ports:
       - '5432:5432'
     volumes:
@@ -38,7 +38,6 @@ services:
       POSTGRES_DB: postgres
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
-    command: postgres -c compute_query_id=on
   pgmeta:
     image: supabase/postgres-meta:v0.91.5
     ports:
--- a/packages/core/postgrest-js/test/transforms.test.ts
+++ b/packages/core/postgrest-js/test/transforms.test.ts
@@ -405,6 +405,7 @@ test('explain with options', async () => {
           "Plan": Any<Object>,
           "Query Identifier": Any<Number>,
           "Settings": Object {
+            "effective_cache_size": "128MB",
             "search_path": "\\"public\\", \\"extensions\\"",
           },
         },


packages/core/postgrest-js/test/db/docker-compose.yml
```diff
@@ -29,7 +29,7 @@ services:
     depends_on:
       - db
   db:
-    image: supabase/postgres:15.1.0.37
+    image: postgis/postgis:15-3.4-alpine
```
grdsdev: I’m not sure about this, but I think we should test against the exact stack that Supabase uses, both hosted and self-hosted.

In the future, we should eliminate the use of `docker-compose.yml` files and instead spin up a local Supabase instance using the CLI, similar to how we do for some packages in Python.

mandarini: Ohhhh good point! I'll add this in my to-do list!

packages/core/postgrest-js/tstyche.config.json
```diff
@@ -1,7 +1,4 @@
 {
-  "$schema": "https://tstyche.org/schemas/config.json",
-  "testFileMatch": ["./test/**/*.test.ts", "./test/**/*.test-d.ts"],
-  "checkSourceFiles": false,
-  "checkSuppressedErrors": true,
-  "tsconfig": "./tsconfig.test.json"
```
avallete, line 6: **issue**

I think we want to keep some of those fields:

1. `$schema` allow for in editor autocomplete for the valid fields handled by tstyche config
2. `checkSuppressedErrors` allow assertion about the error that should be raised by the types not only if an error is raised
3. `checkSourceFiles` allow to exclude the source files that aren't our tests files.
4. We are now moving toward a `type + runtime` test in each test rather than `type-d.ts / runtime` tests. This allows to easily see if a runtime result properly match the expected type all within the same context. Which is why this was run over both `.test.ts` and `.test-d.ts`, so I think both should be kept.

mandarini: DM'd you. I'll make this my top priority tomorrow/next week. Can we mark as ok for now, because we need this PR to go in, so that we can also merge https://github.com/supabase/supabase-js/pull/1752 by tomorrow.

avallete (approved):

packages/core/postgrest-js/test/db/docker-compose.yml
mandarini: https://github.com/supabase/supabase-js/pull/1981

PR #1778, opened 2025-10-16
fix(repo): convert postbuild to explicit codegen

mandarini:
## Description

This PR fixes build script issues and converts the `postbuild` type sync mechanism to an explicit codegen workflow with automated checks.

### What changed?

1. Fixed nested Nx invocations causing build failures:
  - Changed package build scripts in `supabase-js`, `auth-js`, and `storage-js` to  use `npm run` instead of `npx nx`
  - Previously, build scripts calling `npx nx build:main package-name` caused nested Nx processes that conflicted over the project graph
  - This was causing `[readCachedProjectGraph] ERROR: No cached ProjectGraph is available` errors

2. Fixed nested `common/common/` directory creation
  - The `postbuild` script in `postgrest-js` was using `cp -rf ./src/types/common  ../supabase-js/src/lib/rest/types/common`
  - When the destination already existed, it would copy the directory INTO it instead of replacing contents
  - This created nested `common/common/common.ts` and `common/common/rpc.ts` files

3. Converted postbuild to explicit codegen workflow
  - Removed automatic `postbuild` script from `postgrest-js/package.json`
  - Created `scripts/sync-common-types.js` for explicit type synchronization
  - Added `npm run codegen` and `npm run codegen:check` commands
  - Generated files now include AUTO-GENERATED header comments with clear instructions
  - Added pre-commit hook that automatically runs codegen when source types change
  - Added CI check (`codegen:check`) in `ci-core.yml` to ensure types stay in sync

**Postbuild Issues:**
The automatic `postbuild` approach had several issues:
  1. **Builds modifying source files** - Anti-pattern that creates non-deterministic builds
  2. **Git workflow confusion** - Running builds could create uncommitted changes
  3. **Nested directory bug** - `cp -rf` behavior created invalid directory structures
  4. **Hidden behavior** - Developers weren't aware files were being auto-generated

**Why Explicit Codegen is needed:**
  - Builds are pure (same input = same output)
  - Clear when files are generated (explicit `npm run codegen`)
  - Pre-commit hook ensures types stay in sync automatically
  - CI catches any missed syncs with clear error messages
  - AUTO-GENERATED headers warn developers not to edit manually
  - Transparent and predictable workflow

**Context from PR #1632:**
These common types need to be copied (not imported across package boundaries) because TypeScript's structural typing requires exact type definitions for complex generics and conditional types to work correctly across `postgrest-js` and `supabase-js`. The codegen approach maintains this requirement while making the process explicit and safe.

### Related Issues

  - Fixes nested `common/common/` directory creation during builds
  - Improves maintainability of type synchronization from PR #1632

### DIFF
--- a/scripts/sync-common-types.js
+++ b/scripts/sync-common-types.js
@@ -13,7 +13,6 @@
 
 const fs = require('fs')
 const path = require('path')
-const crypto = require('crypto')
 
 const SOURCE_DIR = path.join(__dirname, '../packages/core/postgrest-js/src/types/common')
 const DEST_DIR = path.join(__dirname, '../packages/core/supabase-js/src/lib/rest/types/common')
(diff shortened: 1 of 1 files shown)


mandarini:
@avallete let me know if this is ok for what you intended!

avallete:
Totally ! Great work on this, thank's for catching it 👍

grdsdev (approved):

PR #1793, opened 2025-10-21
feat(auth): add OAuth 2.1 authorization consent management API calls

cemalkilic:
## Summary

  Adds user-facing OAuth authorization methods to `@supabase/auth-js` for building consent pages when Supabase Auth acts as an OAuth 2.1 authorization server.

  Implements the **consent flow** where users approve/deny OAuth client authorization requests.

  ## New API

  ### Namespace: `auth.oauth.*`

  Three new methods for managing OAuth authorization consent:

  ```typescript
  // Get authorization details (client info, scopes, user)
  const { data, error } = await supabase.auth.oauth.getAuthorizationDetails(
    authorizationId,
    { skipBrowserRedirect?: boolean }
  )

  // Approve authorization request
  const { data, error } = await supabase.auth.oauth.approveAuthorization(
    authorizationId,
    { skipBrowserRedirect?: boolean }
  )

  // Deny authorization request
  const { data, error } = await supabase.auth.oauth.denyAuthorization(
    authorizationId,
    { skipBrowserRedirect?: boolean }
  )
```

## Use Case

Enables developers to build custom OAuth consent pages for their Supabase projects when acting as an OAuth 2.1 authorization server. Example flow:

  1. User is redirected to consent page with `authorization_id` in URL
  2. App calls `getAuthorizationDetails(authorizationId)` to fetch client and scope info
  3. User reviews and clicks "Approve" or "Deny"
  4. App calls `approveAuthorization()` or `denyAuthorization()`
  5. SDK auto-redirects back to OAuth client with authorization code/error

## What's NOT Included

Third-party OAuth client integration (calling `/oauth/authorize` and `/oauth/token` from external apps) is intentionally not included in this PR.

Rationale:
  - Standard OAuth 2.1 client libraries already handle this well
  - Would require full OAuth client lifecycle (refresh tokens, expiration, storage)
  - Better served by documentation + existing OAuth libraries
  - This PR focuses on the unique Supabase-specific consent management API, utilizing the existing Supabase session

Third-party integration will be addressed through comprehensive documentation showing how to use standard OAuth client libraries with Supabase Auth endpoints.

## Breaking Changes

None. This is a purely additive change with zero breaking changes to existing APIs

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -105,6 +105,9 @@ import type {
   MFAVerifyWebauthnParamFields,
   MFAVerifyWebauthnParams,
   OAuthResponse,
+  AuthOAuthServerApi,
+  AuthOAuthAuthorizationDetailsResponse,
+  AuthOAuthConsentResponse,
   Prettify,
   Provider,
   ResendParams,
@@ -196,6 +199,12 @@ export default class GoTrueClient {
    * Namespace for the MFA methods.
    */
   mfa: GoTrueMFAApi
+  /**
+   * Namespace for the OAuth 2.1 authorization server methods.
+   * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+   * Used to implement the authorization code flow on the consent page.
+   */
+  oauth: AuthOAuthServerApi
   /**
    * The storage key used to identify the values saved in localStorage
    */
@@ -320,6 +329,12 @@ export default class GoTrueClient {
       challengeAndVerify: this._challengeAndVerify.bind(this),
       getAuthenticatorAssuranceLevel: this._getAuthenticatorAssuranceLevel.bind(this),
       webauthn: new WebAuthnApi(this),
+    }
+
+    this.oauth = {
+      getAuthorizationDetails: this._getAuthorizationDetails.bind(this),
+      approveAuthorization: this._approveAuthorization.bind(this),
+      denyAuthorization: this._denyAuthorization.bind(this),
     }
 
     if (this.persistSession) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/core/auth-js/src/lib/types.ts [hunk 1]
```diff
@@ -1590,3 +1590,103 @@ export interface GoTrueAdminOAuthApi {
...
+  client: OAuthAuthorizationClient
+  /** User object associated with the authorization */
+  user: {
+    /** User ID (UUID) */
+    id: string
+    /** User email */
+    email: string
+  }
+  /** Space-separated list of requested scopes */
+  scope: string
```
fadymak, line 1628: Can these values be optional/undefined? I ask because I see them tagged with `omitempty` but I can't seem to find a case they would be

packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3344,6 +3359,165 @@ export default class GoTrueClient {
...
+            headers: this.headers,
+            jwt: session.access_token,
+            xform: (data: any) => {
+              // If the API returns redirect_uri, it means consent was already given
+              if (data.redirect_uri) {
+                // Automatically redirect in browser unless skipBrowserRedirect is true
+                if (isBrowser() && !options?.skipBrowserRedirect) {
+                  window.location.assign(data.redirect_uri)
+                }
+              }
```
fadymak, line 3399: I'm a bit on the fence about this: it feels like the default should be that we return the redirect URL to the caller and let them handle the redirect given that these APIs are slightly lower level (as opposed to having some hosted consent page).

issuedat (approved):

packages/core/auth-js/src/lib/types.ts
cemalkilic: Good point and yeah ideally they should always exists. The context on why we have `omitempty` is: https://github.com/supabase/auth/pull/2107#discussion_r2260911743

packages/core/auth-js/src/GoTrueClient.ts
cemalkilic: Completely get you, wanted to stick with the existing conventions for `skipBrowserRedirect`

https://github.com/supabase/supabase-js/blob/0d7a77a6ac40144695f1a7cc36accabf0b3e1a3e/packages/core/auth-js/src/GoTrueClient.ts#L2455-L2458

fadymak: I guess I would expect the SDK to redirect me when calling a method `signInWithOAuth` but as a user you might find it unexpected to be redirected when calling `getAuthorizationDetails` (personally I'd expect an object to be returned).

cemalkilic: oh that makes sense. addressing in https://github.com/supabase/supabase-js/pull/1811

PR #1795, opened 2025-10-21
chore(ci): integrate Verdaccio for integration tests

mandarini:
## Description

Integrates Verdaccio local npm registry into CI integration tests to fix dependency resolution issues and improve test reliability.

## Problem

Integration tests (Expo, Next, Bun, Deno) were using tarballs with the `file:` protocol, which caused `*` dependencies in package.json to resolve from the public npm registry instead of the workspace. This led to:

1. Tests installing `@supabase/functions-js@2.76.0` from npm (missing `tslib`)
2. Tests not using the actual workspace versions being tested
3. CI failures due to missing transitive dependencies
4. Complex tarball management with manual packing and copying
5. Tests not catching issues with existing code, because they were installing previously `latest` published version

Example of the issue:

```json
  // supabase-js package.json
  "dependencies": {
    "@supabase/functions-js": "*"  // Resolved from npm, not workspace!
  }
```

##  Solution

Replaced tarball-based approach with Verdaccio (local npm registry):

1. Start Verdaccio in each integration test job

```bash
  npx verdaccio --config .verdaccio/config.yml &
  npm set registry http://localhost:4873/
```

2. Publish all workspace packages to Verdaccio

```bash
  for pkg in auth-js functions-js postgrest-js realtime-js storage-js supabase-js; do
    npm publish --registry http://localhost:4873
  done
```

3. Tests install from Verdaccio
```bash
  npm install --registry http://localhost:4873
  # Now `*` resolves to workspace versions!
```

4. Updated test package.json files
    - Changed `"@supabase/supabase-js": "file:supabase-supabase-js-0.0.0-automated.tgz"`
    - To `"@supabase/supabase-js": "*"`

## Benefits

* Correct dependency resolution - * dependencies now resolve to workspace versions
* Realistic testing
* Tests use npm install like real users 
* Would have caught `tslib` issue - All transitive dependencies properly resolved
* No tarball management - Simpler CI workflow
* Consistent approach - All integration tests use same pattern
* Zero impact on releases - Verdaccio only used for testing

##  Security Note

Verdaccio config uses publish: $all (no authentication). This is safe because:
  - Runs on localhost only (not exposed to internet)
  - Ephemeral (starts/stops with each CI job)
  - Temporary storage (deleted after job completes)
  - Never touches real npm registry

### DIFF
--- a/.github/workflows/ci-supabase-js.yml
+++ b/.github/workflows/ci-supabase-js.yml
@@ -186,34 +186,30 @@ jobs:
           npm view @supabase/supabase-js version --registry http://localhost:4873
           npm view @supabase/supabase-js --registry http://localhost:4873 | head -20
 
-      - name: Run Deno Tests in isolation (1.x)
+      - name: Run Deno Tests (1.x)
         if: ${{ matrix.deno == '1.x' }}
         env:
           STORAGE_JS_ENTRY: main
           DENO_NPM_REGISTRY: http://localhost:4873
         run: |
-          TEST_DIR="/tmp/test-deno-1x-${{ github.run_id }}"
-          mkdir -p "$TEST_DIR"
-          cp -r packages/core/supabase-js/test/deno/* "$TEST_DIR/"
-          cd "$TEST_DIR"
+          cd packages/core/supabase-js/test/deno
           echo "registry=http://localhost:4873/" > .npmrc
           npm install
           npm test || npm test
           rm -f .npmrc
+          cd ../../..
 
-      - name: Run Deno Tests in isolation (2.x)
+      - name: Run Deno Tests (2.x)
         if: ${{ matrix.deno == '2.x' }}
         env:
           DENO_NPM_REGISTRY: http://localhost:4873
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-supabase-js.yml
```diff
@@ -41,22 +39,26 @@ jobs:
...
-          path: packages/core/supabase-js/${{ steps.pack.outputs.filename }}
+      - name: Configure npm to use Verdaccio
+        run: npm set registry http://localhost:4873/
+
+      - name: Publish packages to Verdaccio
+        run: |
+          for pkg in auth-js functions-js postgrest-js realtime-js storage-js supabase-js; do
+            echo "Publishing @supabase/$pkg to Verdaccio..."
+            cd packages/core/$pkg
+            npm publish --registry http://localhost:4873 || echo "Failed to publish $pkg"
```
grdsdev: The `|| echo "Failed to publish $pkg"` won't stop the job, right? Should this fail CI publishing didn't work?

mandarini: Yes yes, I wanted to test some things out! :D

grdsdev (approved):

PR #1802, opened 2025-10-24
feat(auth): add TypeScript types for documented JWT claims fields

### LINKED ISSUE #1584: Provide TypeScript types for documented JWT claims fields (email, phone, user_metadata, etc.)
# Refactor Request

## Describe the refactor

The `getClaims()` method returns `JwtPayload` with type `RequiredClaims & { [key: string]: any
}`, which doesn't provide TypeScript types for common Supabase JWT fields that are documented in
the [JWT Claims Reference](https://supabase.com/docs/guides/auth/jwt-fields).

```typescript
const { data } = await supabase.auth.getClaims()
data.claims.email        // No TypeScript support
data.claims.user_metadata // No TypeScript support
```
The documented fields like email, phone, user_metadata, app_metadata, and is_anonymous exist at
runtime but aren't typed, making developers lose autocomplete and type safety. I understand why you did it, but it doesn't make it smooth to work with.

`getUser()` returns a properly typed User object with these fields, but `getClaims()` loses this type
 information despite returning the same data from the JWT.

## Additional context

Would you consider providing TypeScript types for these documented JWT fields, or what's the
recommended approach for type-safe access to these claims when using getClaims()?

SumitKumar-17:
- Added type safety for email, phone, user_metadata, app_metadata, is_anonymous, etc.
- Provides autocomplete and type safety for all documented JWT claims fields
- getClaims() method now returns properly typed JwtPayload with documented fields

Fixes #1584

Closes #(issue_number) 

## 🔄 Breaking changes

## 📋 Checklist

### DIFF
--- a/packages/core/auth-js/src/lib/types.ts
+++ b/packages/core/auth-js/src/lib/types.ts
@@ -1439,21 +1439,31 @@ export type RequiredClaims = {
   session_id: string
 }
 
+/**
+ * JWT Payload containing claims for Supabase authentication tokens.
+ *
+ * Required claims (iss, aud, exp, iat, sub, role, aal, session_id) are inherited from RequiredClaims.
+ * All other claims are optional as they can be customized via Custom Access Token Hooks.
+ *
+ * @see https://supabase.com/docs/guides/auth/jwt-fields
+ */
 export interface JwtPayload extends RequiredClaims {
+  // Standard optional claims (can be customized via custom access token hooks)
   email?: string
   phone?: string
-  user_metadata?: UserMetadata
-  app_metadata?: UserAppMetadata
   is_anonymous?: boolean
-  is_sso_user?: boolean
-  id?: string
-  created_at?: string
-  updated_at?: string
-  confirmed_at?: string
-  email_confirmed_at?: string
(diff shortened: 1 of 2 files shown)


mrbjjackson:
@SumitKumar-17  I'm not sure if this is the correct place to point this out but I'm just looking at your commit - I see you've added an "id" field. I'm not sure where this would come from. It's not mentioned in the docs as far as I can see - rather the claims object uses the "sub" parameter for the user's uuid.

mandarini:
Hi @SumitKumar-17 ! Thank you for this PR. I discussed it why my team and we think this PR has a few incorrect claims added (e.g.: `id`, `created_at`, `updated_at`, `phone_confirmed_at`, etc...). We would recommend you use the following doc as the source of truth for the claims: https://supabase.com/docs/guides/auth/jwt-fields

One caveat here is that a user may use a custom access token hook so there's no guarantee the claims will exist (other than the required ones), so they should all be marked as optional.

Do you think you're able to make these changes? If not, let me know and we can help!

mandarini:
Fixed the types, updated the PR @issuedat

PR #1807, opened 2025-10-27
fix(postgrest): add incoming major 14 support

avallete:
## 🔍 Description

Add type support for incoming postgREST 14 major.

### What changed?

Now we can have either postgREST 12 or 13 or 14

Since we now introspect postgREST version directly, 14.xxx would've make the behavior fallback to 12.xx by default.

~TODO: add runtime tests for new major.~
Tested to update tests snapshots with 14, no differences were found runtime-wise.

### DIFF
--- a/packages/core/postgrest-js/test/db/docker-compose.yml
+++ b/packages/core/postgrest-js/test/db/docker-compose.yml
@@ -1,7 +1,7 @@
 # docker-compose.yml
 services:
-  rest13:
-    image: postgrest/postgrest:v13.0.0
+  rest14:
+    image: postgrest/postgrest:v14.0
     ports:
       - '3001:3000'
     environment:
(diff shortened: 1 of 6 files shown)


packages/core/postgrest-js/src/types/feature-flags.ts
```diff
@@ -0,0 +1,17 @@
...
+    ? true
+    : IsPostgrest14<PostgrestVersion> extends true
+      ? true
+      : false
+
+export type MaxAffectedEnabled<PostgrestVersion extends string | undefined> =
+  IsPostgrestVersionGreaterThan12<PostgrestVersion> extends true ? true : false
+
+export type SpreadOnManyEnabled<PostgrestVersion extends string | undefined> =
+  IsPostgrestVersionGreaterThan12<PostgrestVersion> extends true ? true : false
```
avallete, line 17: **note**

Isolate all "versions flagged features" here for convenience. And account four postgrest 14.

PR #1830, opened 2025-11-04
feat(repo): remove node-fetch dependency, require Node.js 20+

mandarini:
Remove `@supabase/node-fetch` polyfill in favor of native fetch API available in Node.js 20+. This change drops support for Node.js 18, which reached end-of-life on April 30, 2025.

All target environments (Node 20+, browsers, Deno, Bun, React Native) have native fetch support. No API changes for users on supported platforms.

## Changes:
- Remove @supabase/node-fetch dependency from all packages
- Simplify fetch resolution logic to use native fetch
- Add Node.js >= 20 engine constraint to all packages
- Update documentation with migration guide
- Remove Node 18 CI workflow

See: https://github.com/orgs/supabase/discussions/37217

Node.js 18 is no longer supported. Minimum version is now Node.js 20.

## Breaking

This will be a breaking change since we're not supporting Node 18 any more.

### DIFF
--- a/packages/core/auth-js/test/GoTrueClient.browser.test.ts
+++ b/packages/core/auth-js/test/GoTrueClient.browser.test.ts
@@ -80,6 +80,22 @@ describe('GoTrueClient in browser environment', () => {
   })
 
   it('should handle PKCE flow', async () => {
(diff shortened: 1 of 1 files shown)


soedirgo:
Some failures in auth-js tests

soedirgo (approved):

PR #1833, opened 2025-11-05
feat(auth): add OAuth grant listing and revocation endpoints

cemalkilic:
adds OAuth grant listing and revocation endpoints

reference: https://github.com/supabase/auth/pull/2232 and follow-up: https://github.com/supabase/auth/pull/2247

### DIFF
--- a/packages/core/auth-js/src/lib/types.ts
+++ b/packages/core/auth-js/src/lib/types.ts
@@ -1693,10 +1693,7 @@ export type OAuthGrant = {
  * Response type for listing user's OAuth grants.
  * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
  */
-export type AuthOAuthGrantsResponse = RequestResult<{
-  /** Array of OAuth grants authorized by the user */
-  grants: OAuthGrant[]
-}>
+export type AuthOAuthGrantsResponse = RequestResult<OAuthGrant[]>
 
 /**
  * Response type for revoking an OAuth grant.
@@ -1755,7 +1752,7 @@ export interface AuthOAuthServerApi {
    * Lists all OAuth grants that the authenticated user has authorized.
    * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
    *
-   * @returns Array of OAuth grants with client information and granted scopes
+   * @returns Response with array of OAuth grants with client information and granted scopes
    */
   listGrants(): Promise<AuthOAuthGrantsResponse>
 


packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3544,6 +3548,77 @@ export default class GoTrueClient {
...
+
+      throw error
+    }
+  }
+
+  /**
+   * Revokes a user's OAuth grant for a specific client.
+   * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+   */
+  private async _revokeOAuthGrant(clientId: string): Promise<AuthOAuthRevokeGrantResponse> {
```
fadymak: Should we make this accept an object of options? For example:

* No options: `{}` -> revokes all grants for the user
* `{ clientId: '...' }` -> revokes all grants for that client ID
* `{ resourceId: '' }` (future?) -> revokes all grants for the resource server

cemalkilic: Good point! I've updated it to accept an options object for better extensibility. Currently, `clientId` is required since the Auth server requires it for revocation until we implement the other revoke options

packages/core/auth-js/src/lib/types.ts
```diff
@@ -1676,6 +1676,40 @@ export type AuthOAuthConsentResponse = RequestResult<{
...
+ */
+export type OAuthGrant = {
+  /** OAuth client identifier (UUID) */
+  client_id: string
+  /** Human-readable name of the OAuth client */
+  client_name: string
+  /** URI of the OAuth client's website */
+  client_uri: string
+  /** URI of the OAuth client's logo */
+  logo_uri: string
```
fadymak: Small thing that came to mind just now (sorry I didn't think of it when I reviewed the auth server PR) — for these options, I wonder if we should nest them in the response:

```js
{
  client: {
    id: "...",
    name: "...",
    uri: "...",
  },
  scopes: [],
  granted_at: ...
}
```

This would avoid having the `client_` prefix repeated.

Not sure if we use this convention across the board but just wanted to bring it up in case you'd like to consider it (mostly just a stylistic/subjective thing).

issuedat (dismissed):

packages/core/auth-js/src/GoTrueClient.ts
fadymak: Thanks ❤️ Yeah it's always easier to move from required -> optional than the other way around ✅

packages/core/auth-js/src/lib/types.ts
cemalkilic: Sure, totally make sense!! I've updated `OAuthGrant` to use the nested structure, which is consistent with how `OAuthAuthorizationDetails` already uses `OAuthAuthorizationClient`

packages/core/auth-js/src/lib/types.ts
```diff
@@ -1676,6 +1676,34 @@ export type AuthOAuthConsentResponse = RequestResult<{
...
+  granted_at: string
+}
+
+/**
+ * Response type for listing user's OAuth grants.
+ * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+ */
+export type AuthOAuthGrantsResponse = RequestResult<{
+  /** Array of OAuth grants authorized by the user */
+  grants: OAuthGrant[]
```
fadymak: Is there a reason we don't return an array instead of nesting the `grants` property? I feel like it's common to return an array from list endpoints.

issuedat (dismissed):

packages/core/auth-js/src/lib/types.ts
cemalkilic: updated accordingly, thanks!

PR #1838, opened 2025-11-06
fix(storage)!: do not throw error when file does not exist in exists method

### LINKED ISSUE #1363: Storage.from.exists throws an error when the file doesn't exist in the bucket
# Bug report

## Describe the bug
The storage.from(<bucket>).exists(<path>) method returns false **AND** an exception if the file doesn't exist. This doesn't seem like correct behavior. If the cause of the error was only that the file doesn't exist, this api should return `data: false`, `error: null`. The error should only be set if it is a true error. 

## To Reproduce
```
import { createClient } from "@supabase/supabase-js";
const client = createClient(url, key)

// ensure my_bucket exists, but foo.txt does not. 
const r = await client.storage.from("my_bucket").exists("foo.txt")
console.log(JSON.stringify(r, null, 2));
```

## Expected behavior
Result should not contain an error simply if the file doesn't exist. Error should only be set if there is a true error.

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

### LINKED ISSUE #1600: Method storage.from(bucket).exists missing on front-end

mandarini:

Bug fix (breaking).

`storage.from().exists()` returns `{ data: false, error: <StorageError> }` when a file doesn't exist, even though the existence check itself succeeded. Callers have to inspect `error` to decide whether the 404 means "the file isn't there" (a successful answer) or a real failure.

## What is the new behavior?

`exists()` now returns:

- `{ data: true, error: null }` when the file exists
- `{ data: false, error: null }` when the file does not exist
- An error (returned via the result `error` field, or thrown when `.throwOnError()` is set) for any other failure — auth, network, server errors, etc.

Callers should branch on `data === false` to detect "not found"; `error` is now reserved for genuine failures.

## Note on status codes

The Supabase Storage API returns HTTP **400** for HEAD on a missing object (the server intentionally normalizes `NoSuchKey` 404s to 400 in `StorageBackendError`, likely to avoid leaking object existence to unauthorized callers). It also returns **404** in some configurations. `exists()` therefore treats both 400 and 404 as "not found" and matches the prior behavior for that branch — only the `error` field changes from `StorageError` to `null`.

## Tests

- Updated the integration test at `packages/core/storage-js/test/storageFileApi.test.ts` to assert `error` is `null` when the file is missing.
- Added unit tests at `packages/core/storage-js/test/storageFileApiErrorHandling.test.ts` covering the 404 (no-error) and 400 (surfaces error) paths via mocked `fetch`.

## Docs

Updated the TSDoc `@returns` line on `exists()` so the contract is documented at the source.

## Additional context

- Closes #1363
- Targets `develop` for v3
- Supersedes #1745
- Original contributor work: @SumitKumar-17 in supabase/storage-js#256

## Breaking change

BREAKING CHANGE: `storage.from().exists()` now returns `{ data: false, error: null }` when a file does not exist (HTTP 404). Previously it returned `{ data: false, error: StorageError }`. Branch on `data === false` instead of `error` to detect "not found". Additionally, HTTP 400 (malformed request) is no longer treated as "not found" and now surfaces as a real error.

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -941,7 +941,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
    *
    * @category File Buckets
    * @param path The file path, including the file name. For example `folder/image.png`.
-   * @returns `{ data: true, error: null }` when the file exists, `{ data: false, error: null }` when it does not (HTTP 404), or `{ data: false, error: StorageError }` for any other failure.
+   * @returns `{ data: true, error: null }` when the file exists, `{ data: false, error: null }` when it does not (HTTP 400 or 404), or throws/returns a `StorageError` for any other failure.
    *
    * @example Check file existence
    * ```js
@@ -982,7 +982,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
               ? (error.originalError as { status: number })?.status
               : undefined
 
-        if (status === 404) {
+        if (status !== undefined && [400, 404].includes(status)) {
           return { data: false, error: null }
         }
       }
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -941,7 +941,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
    *
    * @category File Buckets
    * @param path The file path, including the file name. For example `folder/image.png`.
-   * @returns Promise with response containing boolean indicating file existence or error
+   * @returns `{ data: true, error: null }` when the file exists, `{ data: false, error: null }` when it does not (HTTP 404), or `{ data: false, error: StorageError }` for any other failure.
    *
    * @example Check file existence
    * ```js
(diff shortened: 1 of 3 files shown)


grdsdev (dismissed):

yahyabedirhan:
hi @mandarini i have a question about the diff

as far as i see, `exists` function results with one of these:

- file exists: `{ data: true, error: null }`
- file doesn't exist: `{ data: false, error: null }`
- or throw error.

what is the reason keeping the `error` in the resulting object, if it'll be null in both cases? is it because of API design, and consistency with other endpoints?

would it be possible to make the return type `{ data: boolean }` to avoid confusion? what do you think?

grdsdev (approved):

PR #1844, opened 2025-11-10
Replace Math.random() with performance.now() in uuid() for pre-render…

BOXNYC:
… builds in Next.js v16

Replace Math.random() with performance.now() in uuid() for pre-rendered caching in Next.js v16 Uses `performance.now()` as the primary entropy source, mixed with a counter and simple hashing. This is a replacement to using Math.random().

Closes #(issue_number) 

## 🔄 Breaking changes

## 📋 Checklist

- [ ] I have read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md)
- [ ] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `<type>(<scope>): <description>`
- [ ] I have run `npx nx format` to ensure consistent code formatting
- [ ] I have added tests for new functionality (if applicable)

### DIFF
--- a/packages/core/auth-js/src/lib/helpers.ts
+++ b/packages/core/auth-js/src/lib/helpers.ts
@@ -9,12 +9,27 @@ export function expiresAt(expiresIn: number) {
   return timeNow + expiresIn
 }
 
-export function uuid() {
-  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
-    const r = (Math.random() * 16) | 0,
-      v = c == 'x' ? r : (r & 0x3) | 0x8
-    return v.toString(16)
-  })
+export function uuid(): string {
+    // Use performance.now() as primary entropy source
+    let counter = 0;
+    let state = 0;
+    
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+        // Use performance timer if available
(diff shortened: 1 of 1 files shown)


mandarini:
Hi @BOXNYC,

Thank you for the quick turnaround on this PR. I understand this is blocking for you with Next.js 16. Unfortunately, I can't merge this PR because using `performance.now()` as an entropy source creates security vulnerabilities:

1. **Predictable UUIDs**: Timestamps are easily guessable, which is dangerous for session/auth tokens
2. **Collision risk**: High-frequency operations can generate duplicate UUIDs
3. **Security regression**: This downgrades security compared to even Math.random()

While I understand that `crypto.getRandomValues()` also fails the Next.js build (thanks for testing that), the root issue is that `createBrowserClient()` shouldn't be generating UUIDs during server pre-rendering at all. We need to fix this architecturally.

## Workaround to Unblock You

Try deferring client initialization to client-side only, or use a client-only wrapper component. This avoids the prerender issue entirely without compromising security. Let me know if this unblocks you while we work on a proper fix.

We'll work with the team to refactor the library to defer UUID generation appropriately.

Thanks again for reporting this!

mandarini:
I am working on a solution here: https://github.com/supabase/supabase-js/pull/1847 but need to discuss with team first

BOXNYC:
The real main issue IMO pertains to createServerClient() which can't be client wrapped. It's the main reason I mention pre-rendering, for us it's an SSR issue. I realize I've been commenting on an issue reporting createBrowserClient()...

PR #1853, opened 2025-11-12
fix(auth): getClaims allow expired

MuzzaiyyanHussain:
### Description
Fixes supabase/auth#2256 — `getClaims(token, { allowExpired: true })` was still throwing an expiration error.

### Change
- Skip `validateExp()` when `allowExpired=true` to properly decode expired JWTs.

### Checklist
- [x] Verified local build
- [x] Fix confirmed via unit tests
- [x] Closes supabase/auth#2256

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -742,9 +742,9 @@ export default class GoTrueClient {
    */
   async signInWithWeb3(credentials: Web3Credentials): Promise<
     | {
-      data: { session: Session; user: User }
-      error: null
-    }
+        data: { session: Session; user: User }
+        error: null
+      }
     | { data: { session: null; user: null }; error: AuthError }
   > {
     const { chain } = credentials
@@ -1021,9 +1021,9 @@ export default class GoTrueClient {
             : []),
           ...(options?.signInWithSolana?.resources?.length
             ? [
-              'Resources',
-              ...options.signInWithSolana.resources.map((resource) => `- ${resource}`),
-            ]
+                'Resources',
+                ...options.signInWithSolana.resources.map((resource) => `- ${resource}`),
+              ]
             : []),
         ].join('\n')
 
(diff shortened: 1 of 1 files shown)


MuzzaiyyanHussain:
@fenos

mandarini:
Hi @MuzzaiyyanHussain,

Thank you for investigating this issue! And thank you very much for your contribution. I've done some testing and found a few important things:

1. Your change is functionally equivalent to the original

The change from `!options?.allowExpired` to `options?.allowExpired !== true` produces identical behavior. In this case, both conditions behave the same for our intended use: `allowExpired` is always a boolean, so the change doesn’t affect the logic. While `!== true` adds a bit of extra type safety, it’s not needed here. The underlying issue seems to be on the server side, not in this check.

2. The real problem is server-side validation

The `allowExpired` option only controls client-side JWT expiration validation. However, when your JWT uses a symmetric algorithm (HS256, which is the default), the code takes a different path:

  1. Line 3657-3663: Checks if it can verify the JWT signature locally
  2. For HS256 tokens → `signingKey` is null (can't verify symmetric keys client-side)
  3. Line 3666-3669: Falls back to calling `await this.getUser(token)`
  4. The server validates the JWT, including expiration, and rejects expired tokens

So even though the client-side check is skipped with `allowExpired: true`, the server still rejects the expired token via the `getUser()` API call.

3. The actual issue remains unfixed

Your PR doesn't address the root cause. The `allowExpired` option is essentially broken for symmetric JWTs because they require server validation.

So, I will be closing this PR, and I transferred the original issue on the auth side!

Again, thank you very very much for your contribution!

MuzzaiyyanHussain:
Thanks for transferring! I'd love to help on the server-side fix if needed, could you point me to where JWT validation happens?

mandarini:
I am already in the process of preparing a PR. I will tag you once I submit it. Thank you @MuzzaiyyanHussain 
We are discussing this internally, and whether we want to support this behaviour in the first place. Thank you for bearing with us.

PR #1854, opened 2025-11-12
feat(postgrest): support INSERT ... RETURNING for insert().select()

### LINKED ISSUE #1832: On an INSERT with a SELECT, use RETURNING (Instead of a separate SELECT call)
### Description

I want `insert({...}).select('id')` to require INSERT, but not require SELECT policy. I want to have a user that can INSERT a record and some corresponding stuff with it. I don't want them to be able to, in general, SELECT a record, but I need to get back at least the id of the thing that was INSERTed so that I can add join things.

### Suggested solution

If RETURNING is used, it should allow `insert({...}).select('id')` to work without having SELECT permissions on the INSERTed row.

### Alternative

The workaround is to create an id in the client, but I don't like that as much as relying on the DB for such things.

### Additional context

_No response_

### Validations

MuzzaiyyanHussain:
### Description
Fixes #1832 — Enables `.insert(...).select('id')` to work without requiring SELECT RLS policy
by using `INSERT ... RETURNING` instead of an additional SELECT request.

### Changes
- Automatically applies `Prefer: return=representation` to insert() requests.
- Ensures backward compatibility for all existing insert flows.

### Impact
- Users with only INSERT permission can now retrieve IDs of inserted rows.
- Reduces insert+select to a single network call.

### Checklist
- [x] Build passes
- [x] Verified correct Prefer header behavior
- [x] Closes #1832

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestQueryBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestQueryBuilder.ts
@@ -332,6 +332,13 @@ export default class PostgrestQueryBuilder<
       }
     }
 
+    // Automatically use RETURNING for inserts so `.select()` works without SELECT policy
+    if (!this.headers.has('Prefer')) {
+      this.headers.append('Prefer', 'return=representation')
+    } else if (![...this.headers.values()].some((h) => h.includes('return='))) {
+      this.headers.append('Prefer', 'return=representation')
+    }
+
     return new PostgrestFilterBuilder({
       method,
       url: this.url,


steve-chavez:
Not sure if this feature is right, I've asked for more input on https://github.com/supabase/supabase-js/issues/1832#issuecomment-3522040596

MuzzaiyyanHussain:
@steve-chavez

MuzzaiyyanHussain:
@mandarini should i close this PR ?

steve-chavez:
@MuzzaiyyanHussain I'd say yes because atm the issue is not well formulated, I've asked for clarification on https://github.com/supabase/supabase-js/issues/1832#issuecomment-3522040596.

mandarini:
I am closing for now.

PR #1862, opened 2025-11-17
fix(realtime): account for null refs when encoding messages

edgurgel:
## 🔍 Description

ref and join_ref can be null/undefined in certain scenarios like when publishing heartbeats.

Currently using JSON to send heartbeats so not an actual issue but in the future this can be changed and we want to be ready.

### What changed?

Serializer handles null/undefined for `join_ref` and `ref`

The `Msg` type is now fixed with optional `join_ref` and `ref`. Heartbeats for example don't send a `join_ref` as they don't have a specific channel to refer to. 

## 🔄 Breaking changes

## 📋 Checklist


edgurgel:
Thanks, team

PR #1900, opened 2025-11-26
fix(postgrest): cross-schema rpc setof type inference

### LINKED ISSUE #1845: RPC fails with select when returning from another schema.
### Describe the bug

Function type inference fails in schema().rpc().select() when setof schema.table and schema.function belong to different schemas.

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

Lets take the following table and function
`CREATE TABLE schema_a.table_a (
  id uuid not null default gen_random_uuid (),
  user_id uuid not null,
)

CREATE OR REPLACE schema_b.function_a()
RETURNS SETOF schema_a.table_a
LANGUAGE plpgsql
AS $$
BEGIN

--SOME OTHER LOGIC, DOESNT MATTER

RETURN QUERY
SELECT t.*
FROM schema_a.table_a;

END;
$$;
`

the supabase-js, since version 2.75, fails when selecting specific rows.

`const {data, error} = await supabase.schema("schema_b").rpc("function_a");`

that works fine and type inference is correct.

but when doing any kind of select:
`const {data, error} = await supabase.schema("schema_b").rpc("function_a").select();`
or
`const {data, error} = await supabase.schema("schema_b").rpc("function_a").select("*");`
or
`const {data, error} = await supabase.schema("schema_b").rpc("function_a").select("id");`

it all fails with `const data: SelectQueryError<"Invalid Relationships cannot infer result type">[] | null`

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 10.70 GB / 31.92 GB
  Binaries:
    Node: 22.20.0 - C:\nvm4w\nodejs\node.EXE
    npm: 10.9.3 - C:\nvm4w\nodejs\npm.CMD
    pnpm: 10.19.0 - C:\nvm4w\nodejs\pnpm.CMD
  Browsers:
    Chrome: 142.0.7444.134
    Edge: Chromium (141.0.3537.85)
  npmPackages:
    @supabase/ssr: catalog: => 0.7.0 
    @supabase/supabase-js: catalog: => 2.75.0 
    supabase: catalog: => 2.54.11
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

7ttp:
## summary  
Fixes a type inference failure when calling `schema().rpc().select()` on functions that return `SETOF` from a different schema.

## changes  
- Fall back to the function’s `Returns` type when `SetofOptions.to` references a table outside the current schema  
- Add a `CrossSchemaError` with a clear message when the return type cannot be inferred  
- Fix `Relationships` lookup to return `null` instead of throwing when a view is missing  
- Add a JSDoc example demonstrating the `overrideTypes` workaround for cross-schema functions

## fixes  
Closes #1845

### DIFF
--- a/packages/core/postgrest-js/src/types/common/rpc.ts
+++ b/packages/core/postgrest-js/src/types/common/rpc.ts
@@ -64,7 +64,7 @@ type RpcFunctionNotFound<FnName> = {
 
 type CrossSchemaError<TableRef extends string> = {
   error: true
-} & `Function returns SETOF from a different schema ('${TableRef}'). Use .rpc<YourReturnType>() to specify the return type explicitly.`
+} & `Function returns SETOF from a different schema ('${TableRef}'). Use .overrideTypes<YourReturnType>() to specify the return type explicitly.`
 
 export type GetRpcFunctionFilterBuilderByArgs<
   Schema extends GenericSchema,
--- a/packages/core/postgrest-js/test/override-types.test-d.ts
+++ b/packages/core/postgrest-js/test/override-types.test-d.ts
@@ -548,3 +548,12 @@ const postgrest = new PostgrestClient<Database>(REST_URL)
     >
   >(true)
 }
+
+// Test cross-schema rpc falls back to Returns type
+{
+  const result = await postgrest.schema('personal').rpc('get_public_users', {})
+  if (result.error) {
+    throw new Error(result.error.message)
+  }
+  expectType<TypeEqual<typeof result.data, { id: string; user_id: string }[]>>(true)
+}
--- a/packages/core/postgrest-js/test/types.override.ts
+++ b/packages/core/postgrest-js/test/types.override.ts
@@ -32,6 +32,18 @@ export type Database = MergeDeep<
           }
         }
       }
+      Functions: {
+        get_public_users: {
+          Args: Record<PropertyKey, never>
+          Returns: { id: string; user_id: string }[]
+          SetofOptions: {
+            from: '*'
+            to: 'public.users'
+            isOneToOne: false
+            isSetofReturn: true
+          }
+        }
+      }
       Views: {}
       Enums: {}
       CompositeTypes: {}
--- a/packages/core/supabase-js/src/lib/rest/types/common/rpc.ts
+++ b/packages/core/supabase-js/src/lib/rest/types/common/rpc.ts
@@ -74,7 +74,7 @@ type RpcFunctionNotFound<FnName> = {
 
 type CrossSchemaError<TableRef extends string> = {
   error: true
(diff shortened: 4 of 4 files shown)


mandarini (changes requested):
Thanks for the PR @7ttp ! This is a good fix for the cross-schema type inference issue. A few things to address:

* The error message mentions wrong API. The CrossSchemaError type says:
```
  Use .rpc<YourReturnType>() to specify the return type explicitly.
```

But `rpc` doesn't accept a return type parameter. The JSDoc example correctly shows `.overrideTypes<>()`. Please update the error message to match:

```
  type CrossSchemaError<TableRef extends string> = {
    error: true
  } & `Function returns SETOF from a different schema ('${TableRef}'). Use 
  .overrideTypes<YourReturnType>() to specify the return type explicitly.`
```

* Can you please add a type test to prevent regressions? There's an existing pattern in  `packages/core/postgrest-js/test/override-types.test-d.ts`. Something like:

```
  // Test cross-schema rpc falls back to Returns type
  {
    const result = await postgrest
      .schema('other_schema')
      .rpc('cross_schema_fn', {})

    expectType<{ id: string; user_id: string }[]>(result.data)
  }
```

Otherwise the approach looks good. Falling back to Returns type when the table isn't in the current schema is the right solution, and fixing the Relationships lookup to handle missing views is a nice catch.

7ttp:
> Thanks for the PR @7ttp ! This is a good fix for the cross-schema type inference issue. A few things to address:
> 
> * The error message mentions wrong API. The CrossSchemaError type says:
> 
> ```
>   Use .rpc<YourReturnType>() to specify the return type explicitly.
> ```
> 
> But `rpc` doesn't accept a return type parameter. The JSDoc example correctly shows `.overrideTypes<>()`. Please update the error message to match:
> 
> ```
>   type CrossSchemaError<TableRef extends string> = {
>     error: true
>   } & `Function returns SETOF from a different schema ('${TableRef}'). Use 
>   .overrideTypes<YourReturnType>() to specify the return type explicitly.`
> ```
> 
> * Can you please add a type test to prevent regressions? There's an existing pattern in  `packages/core/postgrest-js/test/override-types.test-d.ts`. Something like:
> 
> ```
>   // Test cross-schema rpc falls back to Returns type
>   {
>     const result = await postgrest
>       .schema('other_schema')
>       .rpc('cross_schema_fn', {})
> 
>     expectType<{ id: string; user_id: string }[]>(result.data)
>   }
> ```
> 
> Otherwise the approach looks good. Falling back to Returns type when the table isn't in the current schema is the right solution, and fixing the Relationships lookup to handle missing views is a nice catch.

Updated everything as requested,
Please rereview when available!

mandarini (approved):
Thank you! Approving!

PR #1911, opened 2025-12-03
feat(supabase): add failOnNetworkError option to prevent silent fallback to anon key

mandarini:
## Summary

- Adds `failOnNetworkError` option to prevent silent fallback to anon key during network failures
- When enabled, throws instead of silently degrading auth, giving apps clear error handling
- Backward compatible: defaults to `false` (current behavior)

## Problem

During network issues (Cloudflare outages, DNS failures, timeouts), `getSession()` returns `{ session: null, error }`. The client ignores the error and uses the anon key, causing:

- RLS policies with `auth.uid() = user_id` to fail (auth.uid() is NULL)
- Users receive cryptic 406 errors instead of network errors

## Solution

  ```typescript
    auth: { failOnNetworkError: true }
  })
```

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -88,6 +88,7 @@ export default class SupabaseClient<
   protected accessToken?: () => Promise<string | null>
 
   protected headers: Record<string, string>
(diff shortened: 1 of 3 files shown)


mandarini:
Closing after discussion with @hf

PR #1912, opened 2025-12-03
fix(postgrest): set columns parameter for single-object insert/upsert

### LINKED ISSUE #1653: Upsert failing on non-null column, even though value exists in db
# Bug report

## Describe the bug

I want to bulk `update`, but must use `upsert`.  I use `upsert` but receive issues on non-null columns when they are not provided, despite the value existing on the record in the db.

## To Reproduce

Example table:
```
CREATE TABLE test_table (
  id text primary key,
  other_column text null,
  name text not null --<--- The problem
);

INSERT INTO test_table (id, other_column, name)
VALUES ('id1',null,'some_name');
```

Updating a single row (via `upsert`): I know the record exists and the `name` column is populated.  However if I try to `upsert`:
```
await supabase
.from('test_table')
.upsert([
{ id: 'id1', other_column: 'other_column_text' }
])
```
I receive 
```
{
  code: '23502',
  details: 'Failing row contains (id1, other_column_text, null).',
  hint: null,
  message: 'null value in column "name" of relation "test_table" violates not-null constraint'
}

```
## Expected behavior

The row show be "merged", thus it should just update the two columns I provided, and not default the `name` column to null, which in this case is not allowed.

- OS: macOS 
-  "@supabase/ssr": "^0.0.10",
    "@supabase/supabase-js": "^2.39.2",
- Version of Node.js: v20.10.0

mandarini:
## Summary

  - Fix single-object `upsert()` and `insert()` to set the `columns` URL parameter
  - Add `defaultToNull` option to single-object overloads for consistency
  - Previously only array (bulk) operations set `columns`, causing single-object upserts to  attempt updating ALL table columns, setting missing ones to `null`

  Closes #1653

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3341,14 +3341,14 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/@inquirer/external-editor": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/@inquirer/external-editor/-/external-editor-1.0.2.tgz",
-      "integrity": "sha512-yy9cOoBnx58TlsPrIxauKIFQTiyH+0MK4e97y4sV9ERbI+zDxw7i2hxHLCIEGIE/8PPvDxGhgzIOTSOWcs6/MQ==",
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/@inquirer/external-editor/-/external-editor-1.0.3.tgz",
+      "integrity": "sha512-RWbSrDiYmO4LbejWY7ttpxczuwQyZLBUyygsA9Nsv95hpzUWwnNTVQmAq3xuh7vNwCp07UTmE5i11XAEExx4RA==",
       "dev": true,
       "license": "MIT",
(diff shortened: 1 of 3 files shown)


laurenceisla:
The issue https://github.com/supabase/supabase-js/issues/1653 is similar to the one I opened in https://github.com/PostgREST/postgrest/issues/4046, which was closed as not-planned. This is because, for an upsert, it's natural to provide all of the values in the body to insert-or-update a row. So this problem happens because the issue's OP wants to do a bulk update, which is not implemented yet (see https://github.com/PostgREST/postgrest/issues/1959).

They confirm this in the issue (emphasis mine):

> **I want to bulk update**, but must use upsert. I use upsert but receive issues on non-null columns when they are not provided, despite the value existing on the record in the db.

They're forced to use upsert here, but it's not the right tool for this job. I'm afraid the only fix is to make the bulk update work upstream (in PostgREST).

mandarini:
I am closing this PR after discussing with @laurenceisla

PR #1926, opened 2025-12-05
fix(repo): update lock file after dependabot to use npm 11

mandarini:
## Problem

- Every time `dependabot` updated `package-lock.json`, it used `npm@10` which changed the lock file
- Every time `publish.yml` runs, it uses `npm@latest` (now 11) so that we can use trusted publishing. This means that it gets diffs in the lock file.

## Solution

- Add workflow to auto-regenerate lockfile on Dependabot PRs with npm@latest
- Fixes recurring CI failures in publish.yml caused by npm version mismatch (Dependabot uses npm 10, publish uses npm 11)

### DIFF
--- a/.github/workflows/fix-lockfile.yml
+++ b/.github/workflows/fix-lockfile.yml
@@ -1,0 +1,38 @@
+name: Fix Dependabot lockfile
+
+on:
+  pull_request:
+    paths:
+      - 'package-lock.json'
+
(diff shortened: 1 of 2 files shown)


mandarini:
You can specify npm version with renovate I think. But it's ok, I hope this fix fixes the issue. :(

PR #1927, opened 2025-12-05
feat(storage): align analytics from method with { data, error } pattern

mandarini:
## Summary

- Align `StorageAnalyticsClient.from()` catalog operations with SDK's `{ data, error }` pattern
- Bump `iceberg-js` dependency from 0.8.0 to 0.8.1
- Update JSDoc examples and tests to reflect the new API

## Context

An issue showed 500 errors when analytics catalog operations returned 404s. The platform API catch block couldn't distinguish client errors from server errors because the SDK was throwing instead of returning errors in the response.

Most storage operations return `{ data, error }` and don't throw (unless using `throwOnError()`). The analytics catalog was inconsistent with this pattern.

Since analytics buckets are in **public alpha**, we're treating this as a "got it wrong from the beginning" fix rather than a breaking change requiring a major version bump.

## Changes

**Before:**
```js
const catalog = supabase.storage.analytics.from('my-bucket')
const tables = await catalog.listTables({ namespace: ['default'] }) // throws on error
```

**After:**
```
const catalog = supabase.storage.analytics.from('my-bucket')
const { data: tables, error } = await catalog.listTables({ namespace: ['default'] })
if (error?.isNotFound()) {
    console.log('Namespace not found')
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18119,9 +18119,9 @@
       }
     },
     "node_modules/iceberg-js": {
-      "version": "0.8.0",
-      "resolved": "https://registry.npmjs.org/iceberg-js/-/iceberg-js-0.8.0.tgz",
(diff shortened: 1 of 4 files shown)


grdsdev (approved):
Nice!

Issue #1590, opened 2025-09-18
Realtime REST fallback in supabase-js sends empty Authorization header when no session (500 error)


Schlumen:
# Bug report

* [x] I confirm this is a bug with Supabase, not with my own application.
* [x] I confirm I have searched the [[Docs](https://docs.supabase.com/)](https://docs.supabase.com), GitHub [[Discussions](https://github.com/supabase/supabase/discussions)](https://github.com/supabase/supabase/discussions), and [[Discord](https://discord.supabase.com/)](https://discord.supabase.com).

## Describe the bug

When using `@supabase/supabase-js` to broadcast to Realtime **without subscribing to the channel first**, the client falls back to the REST endpoint `/realtime/v1/api/broadcast`. In this REST fallback, if there is **no user session/access token**, the library sends an **empty `Authorization` header** (exactly `Authorization: \r\n`, with **no** value at all). This results in a **500 Internal Server Error** from the Realtime endpoint.

Notes:

* The empty `Authorization` header occurs even if `global.headers.Authorization` is configured on the client.
* Manually calling the REST endpoint with **only** the `apikey` header (and **no** `Authorization` header) works and returns **202 Accepted**.
* Calling `client.realtime.setAuth(<token>)` (server-side) or using WebSocket send after `subscribe()` also avoids the issue.

It looks like the REST fallback builds/forces an `Authorization` header from the current auth session, and if none exists, it still emits an empty header instead of omitting it (or honoring `global.headers`).

## To Reproduce

Minimal server-side repro (no user session, Service Role client):

```ts
import { createClient } from '@supabase/supabase-js'

const supabase = createClient(
  process.env.NEXT_PUBLIC_SUPABASE_URL!,
  process.env.SUPABASE_SERVICE_ROLE_KEY!, // server-side only
  {
    auth: { autoRefreshToken: false, persistSession: false },
    // Setting global headers does not help for the fallback:
    global: {
      headers: {
        apikey: process.env.SUPABASE_SERVICE_ROLE_KEY!,
        Authorization: `Bearer ${process.env.SUPABASE_SERVICE_ROLE_KEY!}`,
      },
    },
  }
)

// No ch.subscribe() here → triggers REST fallback:
const ch = supabase.channel(`channel:${'some-channel-id'}`, {
  config: { private: true },
})

const ok = await ch.send({
  type: 'broadcast',
  event: 'new-message',
  payload: { hello: 'world' },
})

[...]

PR #1937, opened 2025-12-09
fix(realtime): omit authorization header when no access token exists

### LINKED ISSUE #1936: REST Fallback for Realtime Broadcast Sends Empty Authorization Header, Causing 500 Error
# Bug Report

I can confirm that this issue originates in Supabase itself and not in my application code. I have also reviewed the documentation, GitHub Discussions, and Discord prior to submitting this report.

## Summary

When using `@supabase/supabase-js` to send a broadcast message through Realtime **without subscribing to the channel first**, the client switches to the REST fallback endpoint (`/realtime/v1/api/broadcast`).
If no auth session or access token exists, this fallback layer generates an **empty `Authorization` header** (literally `Authorization: \r\n`).
Realtime responds with **HTTP 500**, even though manually calling this endpoint without *any* `Authorization` header works and returns `202`.

### Key observations

* The REST fallback **always emits an `Authorization` header**, even when no token is present.
* If the session is absent, the header is sent with **no value**, instead of being omitted.
* `global.headers.Authorization` is **ignored** by the REST fallback path.
* Sending through WebSocket (after `subscribe()`) or explicitly calling `realtime.setAuth(<token>)` prevents the issue.
* Making the same request manually with only the `apikey` header succeeds.

---

## Steps to Reproduce

Minimal server-side example (Service Role key, no user session):

```ts

  process.env.NEXT_PUBLIC_SUPABASE_URL!,
  process.env.SUPABASE_SERVICE_ROLE_KEY!,
  {
    auth: { autoRefreshToken: false, persistSession: false },
    global: {
      headers: {
        apikey: process.env.SUPABASE_SERVICE_ROLE_KEY!,
        Authorization: `Bearer ${process.env.SUPABASE_SERVICE_ROLE_KEY!}`,
      },
    },
  }
)

const ch = supabase.channel(`channel:${'some-channel-id'}`, {
  config: { private: true },
})

// No subscribe() call → triggers REST fallback
const ok = await ch.send({
  type: 'broadcast',
  event: 'new-message',
  payload: { hello: 'world' },
})

console.log('send ok?', ok)
```

### Actual request produced by the client

```
POST /realtime/v1/api/broadcast
apikey: <service_role>
Authorization:
Content-Type: application/json
```

This results in:

* `HTTP 500 Internal Server Error`
* `channel.send()` returns `{ error: ... }` without any detail about the 500 status.

### Manual request that *does* work

```ts
await fetch(`${process.env.NEXT_PUBLIC_SUPABASE_URL}/realtime/v1/api/broadcast`, {
  method: 'POST',
  headers: {
    apikey: process.env.SUPABASE_SERVICE_ROLE_KEY!,
  },
  body: JSON.stringify({
    messages: [
      {
        topic: `channel:${'some-channel-id'}`,
        event: 'new-message',
        payload: { hello: 'world' },
        private: true,
      },
    ],
  }),
})
```

This returns **202 Accepted**, and clients receive the broadcast.

---

## Effective Workarounds

* Call `subscribe()` and wait for the `SUBSCRIBED` status before calling `.send()`.
* Set the auth token explicitly via `supabase.realtime.setAuth(<jwt>)` server-side.

Both prevent the empty `Authorization` header from being emitted.

---

## Expected Behavior

* The REST fallback should **omit** the `Authorization` header if no token is available, instead of sending a blank one.
* Alternatively, the fallback should respect `global.headers.Authorization` when no auth session exists.
* The Realtime REST endpoint should return **400 or 401** for empty/invalid headers instead of `500`.

---

This scenario is fairly easy to run into, especially when `.send()` is called before `.subscribe()`. The silent `500` makes debugging difficult, since the status is only visible when inspecting raw network traffic.

---

## Possible Fixes

1. In `@supabase/supabase-js`, avoid constructing an `Authorization` header when `access_token` is falsy (or use the value from `realtime.getAuth()` / `global.headers`).
2. In Realtime, return a proper client error instead of a server error when receiving an empty or invalid `Authorization` header.

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

_No response_

### System Info

```shell
* **OS:** Windows 11 Home
* **supabase-js version:** 2.57.4
* **Node.js version:** 22.x
* **Environment:** server-side only, using Service Role key
* `auth.persistSession` is intentionally disabled, so the client has no session.
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

7ttp:
## summary

when using the `realtime rest` fallback without a session, the client sends an empty authorization header which causes a 500 error from the server.

## problem

the rest fallback in `httpSend()` and `send()`always includes an `authorization header`, even when no token exists. this results in `Authorization: ` (empty value) being sent, which the server rejects with 500.

## fix

only add the authorization header when `AccessTokenValue` is truthy. the server accepts requests with just the `apikey` header.

closes #1936 & #1590
supersedes #1561

### DIFF
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -554,21 +554,22 @@ export default class RealtimeChannel {
     payload: any,
     opts: { timeout?: number } = {}
   ): Promise<{ success: true } | { success: false; status: number; error: string }> {
-    const authorization = this.socket.accessTokenValue
-      ? `Bearer ${this.socket.accessTokenValue}`
-      : ''
-
     if (payload === undefined || payload === null) {
       return Promise.reject('Payload is required for httpSend()')
     }
 
+    const headers: Record<string, string> = {
+      apikey: this.socket.apiKey ? this.socket.apiKey : '',
+      'Content-Type': 'application/json',
+    }
+
+    if (this.socket.accessTokenValue) {
+      headers['Authorization'] = `Bearer ${this.socket.accessTokenValue}`
+    }
+
     const options = {
       method: 'POST',
-      headers: {
-        Authorization: authorization,
-        apikey: this.socket.apiKey ? this.socket.apiKey : '',
-        'Content-Type': 'application/json',
-      },
+      headers,
       body: JSON.stringify({
         messages: [
           {
@@ -626,16 +627,18 @@ export default class RealtimeChannel {
       )
 
       const { event, payload: endpoint_payload } = args
-      const authorization = this.socket.accessTokenValue
-        ? `Bearer ${this.socket.accessTokenValue}`
-        : ''
+      const headers: Record<string, string> = {
+        apikey: this.socket.apiKey ? this.socket.apiKey : '',
+        'Content-Type': 'application/json',
+      }
+
+      if (this.socket.accessTokenValue) {
+        headers['Authorization'] = `Bearer ${this.socket.accessTokenValue}`
+      }
+
       const options = {
         method: 'POST',
-        headers: {
-          Authorization: authorization,
-          apikey: this.socket.apiKey ? this.socket.apiKey : '',
-          'Content-Type': 'application/json',
-        },
+        headers,
         body: JSON.stringify({
           messages: [
             {
--- a/packages/core/realtime-js/test/RealtimeChannel.messaging.test.ts
+++ b/packages/core/realtime-js/test/RealtimeChannel.messaging.test.ts
@@ -330,7 +330,7 @@ describe('send', () => {
       {
         description: 'without access token',
         accessToken: undefined,
-        expectedAuth: '',
+        expectedAuth: undefined,
       },
       {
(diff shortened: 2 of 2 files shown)


7ttp:
@mandarini 😅 could you review this as well when you get a chance?

topce:
Hi @7ttp and  @mandarini  
maybe this PR  introduce error
on this line 
type DatabaseWithoutInternals<DB> = Omit<DB, InternalSupabaseKey>;
got error
[ERROR] TS2304: Cannot find name 'InternalSupabaseKey'. 
maybe intention was 
type DatabaseWithoutInternals<DB> = Omit<DB, 'InternalSupabaseKey'>;

If somebody get same  error can disable it with TS config option skipLibCheck : true

2 cents from Cents ;-)

7ttp:
hi @topce thanks for letting us know💚
unfortunately its not related to this pr,
ill look onto it and fix accordingly if its reproducible.

mandarini:
@topce can you please provide a minimum reproduction repository? It would be super helpful.

PR #1939, opened 2025-12-10
fix(postgrest): preserve nullability when using JSON path

### LINKED ISSUE #1635: Null should be included in return type when using json path (->)

## Describe the bug

When I use a json path in `select()` the type inference produces non nullable types even when the return type might be null.

## To Reproduce

1. Create a table with a column `json_col` that can be null and override the database type to be `{ a: string } | null`
2. Create a query with `.select('json_col->a')`
3. Return type will be { a: string }

## Expected behavior

This query will actually return `{ a: null } when the column is null. This should be reflected in the type.

This should be the case when the column is nullable or when one of the properties along the path is nullable.

I currently only have this type safety when I do

```
const { data } = ...select('json_col');

data!.json_col.a; // Object is possibly 'null'.
```

dev-hari-prasad:
Description:
**What changed?**  
Fixed the type inference for JSON path selections (`->`) by adding `| null` when the underlying JSON column can be null.

When selecting a JSON path from a nullable JSON column, the generated TypeScript types incorrectly assumed the value could never be null.  
In reality, the runtime returns `null` when the whole column is null, so the types were unsafe.  
This patch brings the inferred type in line with actual behavior.

Closes/Fixes #1635

Screenshots/Examples:
N/A since it is a type-only change.

Breaking changes:
This PR contains **no breaking changes**.

📋 Checklist: 
- [x] I have read the Contributing Guidelines  
- [x] My PR title follows the conventional commit format: `<type>(<scope>): <description>`  

Additional notes:
The change is intentionally minimal and only affects the nullable branch of `JsonPathToType`.  
It preserves existing behavior for non-nullable JSON columns while correctly reflecting actual return values for nullable ones.

CC @avallete & @mandarini


mandarini:
Thanks for the fix @dev-hari-prasad ! This correctly addresses the nullability issue in #1635. However, since this changes type inference in a way that could cause compilation errors for existing users (nullable JSON paths will now correctly return `T | null` instead of `T`), I will be treating this as a breaking change.

Before merging, could you please:
  1. Add type tests that verify the new nullable behavior
  2. Update any existing type tests that may be affected

I'll tag this for v3 (next major release). And we can merge it when the time comes!

dev-hari-prasad:
> Before merging, could you please:
> 
> 1. Add type tests that verify the new nullable behavior
> 2. Update any existing type tests that may be affected

Sure

dev-hari-prasad:
@mandarini Tests are now added, let me know if anything else is required

packages/core/postgrest-js/test/index.test-d.ts
```diff
@@ -312,7 +312,7 @@ const postgrestWithOptions = new PostgrestClient<DatabaseWithOptions>(REST_URL)
   expectType<
     {
       baz: string
-      en: 'ONE' | 'TWO' | 'THREE'
+      en: string
```
avallete: issue
That could be an issue, type-wise this shouldn't be broaden out into "any string" when it's actually a specific union. I wonder why the inference is loosing the proper type here.

packages/core/postgrest-js/test/index.test-d.ts
```diff
@@ -283,11 +283,11 @@ const postgrestWithOptions = new PostgrestClient<DatabaseWithOptions>(REST_URL)
...
       typeof result.data,
       {
-        baz: number
-        en: 'ONE' | 'TWO' | 'THREE'
+        baz: number | null
+        en: 'ONE' | 'TWO' | 'THREE' | null
         bar: {
           baz: number
-        }
+        } | null
```
avallete, line 292: **question**

Since `bar` is nullable because of the single arrow operator (`->`) shouldn't the `baz` field also be `{ baz: number | null }` ?

Could be useful to add a runtime test where `->>` operator is used over a json field that is `null`.

mandarini:
Thanks @dev-hari-prasad ! We'll wait for v3 now!

packages/core/postgrest-js/test/relationships.test.ts
avallete: **note**

If the field is nullable, we should have a runtime case where the actual result from postgrest is null.
Goal of those tests is to quickly be able to see  at a glance that postgrest runtime value and type inference matches 1-1.

packages/core/postgrest-js/test/types.override.ts
```diff
@@ -4,7 +4,7 @@ import { MergeDeep } from 'type-fest'
 import { z } from 'zod'
 
 export const CustomUserDataTypeSchema = z.object({
-  foo: z.string(),
+  foo: z.object({ bar: z.record(z.string(), z.unknown()) }),
```
avallete: **note**

Would rather keep this field a string, and put this `record` into another field. So we can reduce the changes in the tests but rather have addition to check this new `record` behaviour.

dev-hari-prasad:
Happy to update or rewrite the affected tests as part of this pr. Just let me know if you’d like me to take that on, or if you prefer to handle it internally.

@mandarini @avallete

avallete:
> Happy to update or rewrite the affected tests as part of this pr. Just let me know if you’d like me to take that on, or if you prefer to handle it internally.
> 
> @mandarini @avallete

Hey @dev-hari-prasad first of all, thanks a lot for the contribution!

You’re very welcome to address the comments I left, that would be much appreciated and should help speed up getting this merged and released.

Otherwise, I can look into adding or updating the tests myself once I have the bandwidth.

packages/core/postgrest-js/test/index.test-d.ts
```diff
@@ -311,9 +313,9 @@ const postgrestWithOptions = new PostgrestClient<DatabaseWithOptions>(REST_URL)
   }
   expectType<
     {
-      baz: string
-      en: 'ONE' | 'TWO' | 'THREE'
-      bar: string
+      baz: string | null
+      en: 'ONE' | 'TWO' | 'THREE' | null
+      bar: string | null
```
avallete: **note**

This is the part why this must be treated as a breaking change. Even though this is more correct, this new "option" on defined custom types might break people typecheck on upgrade.

avallete (approved):

avallete:
I don't expect this to really be a BC anymore. But as @mandarini mentioned, some users might rely over the "non null" behavior so that might break their typecheck (even if that's more correct).

mandarini:
hmmm i don't know what happened, i did not close this. give me a moment

mandarini:
i merged the other PR so base branch got deleted. give me a moment to fix this

mandarini:
Yeah, something weird happen, and we cannot reopen the PR and target the master branch, this is what github copilot says:

## What happened to this PR

This PR was automatically closed when its base branch (`chore/add-supabase-cli-as-devdeps-package`) was merged to `master` and subsequently deleted.

### Timeline of events: 

1. **PR created**: `fix/Null-should-be-included-in-return-type` → `chore/add-supabase-cli-as-devdeps-package`
2. **Base branch merged**:  `chore/add-supabase-cli-as-devdeps-package` was merged into `master` and deleted
3. **PR auto-closed**: GitHub closed this PR instead of retargeting it to `master`

### Why GitHub closed it instead of retargeting:

The head branch (`fix/Null-should-be-included-in-return-type`) was based on `chore/add-supabase-cli-as-devdeps-package` and contained all of its commits plus additional changes. When the base branch was merged to `master`, those commits became part of `master`, creating an ambiguous situation where many commits in this PR were now already in the target branch.

Rather than auto-retarget and show duplicate commits, GitHub closed the PR. 

### Next steps:

I created a new PR (https://github.com/supabase/supabase-js/pull/2011) that: 
- Targets `master` directly
- Contains only the unique commits from this branch (rebased onto `master`)
- Preserves the fix for preserving nullability when using JSON path

---

Thanks @dev-hari-prasad for the contribution!  The work continues in the new PR.

Issue #1677, opened 2025-04-30
getAuthenticatorAssuranceLevel() dont work on edge function


Dramelac:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

I'm aware the problematic of this issue is the same and the source of the problem also comes from the session but the subject and the implementation is not the same so I didn't consider it a duplicate of https://github.com/supabase/auth-js/issues/910

## Describe the bug

Hello everyone

I'm creating this issue because I'm trying to check the authentication level (aal) on the edge function side and impose aal2 if the user has enrolled at least one factor.
When I try to retrieve `await supabaseClient.auth.mfa.getAuthenticatorAssuranceLevel()` I get this:
```
{
  currentLevel: null,
  nextLevel: null,
  currentAuthenticationMethods: []
}
```
(Even though the MFA is properly configured for the current user)

## To Reproduce

Here is my code from a supabase edge function. Run locally `Using supabase-edge-runtime-1.67.2 (compatible with Deno v1.45.2)`:
```
const supabaseClient = createClient<Database, app_schema>(
        Deno.env.get("SUPABASE_URL") ?? "",
        Deno.env.get("SUPABASE_ANON_KEY") ?? "",
        {
            global: {headers: {Authorization: authHeader}},
            // I tried different auth options but nothing fix the getAuthenticatorAssuranceLevel response
            auth: {
                autoRefreshToken: false,
                persistSession: false,
                detectSessionInUrl: false,
                storage: false
            }
        }
    );
// getUser works, I can verify the session and get the list of verified factors but not the current aal
const current_user = await supabaseClient.auth.getUser();

// But not MFA policy check
const { data , error } = await supabaseClient.auth.mfa.getAuthenticatorAssuranceLevel()
console.log("Current MFA:", data)

// RLS query works too
[..]
```
The `createClient` code is from the official documentation exemple: 
https://github.com/supabase/supabase/blob/14562e9c8abe793bd428eba3a1da196bf03f3e41/examples/edge-functions/supabase/functions/select-from-table-with-auth-rls/index.ts#L18-L30

What I tried to do as workaround:
- Differents option on `auth{}` option like `persistSession: false`, `storage: false` to put the session in memory instead of local storage.
[...]

PR #1940, opened 2025-12-11
feat(auth): add optional jwt parameter to getAuthenticatorAssuranceLevel

7ttp:
## Summary

Adds an optional `jwt` parameter to `mfa.getAuthenticatorAssuranceLevel()`, enabling it to work in server side environments like Edge Functions where no session is stored.

## Problem

In Edge Functions, `getAuthenticatorAssuranceLevel()` always returns `{ currentLevel: null, nextLevel: null, currentAuthenticationMethods: [] }` because it relies on `getSession()` which reads from storage. Edge Functions don't have a stored session, only the JWT from the Authorization header.

## Solution

Added an optional `jwt` parameter following the same pattern as `getUser(jwt?)`:

closes https://github.com/supabase/supabase-js/issues/1677

### DIFF
--- a/packages/core/auth-js/test/GoTrueClient.test.ts
+++ b/packages/core/auth-js/test/GoTrueClient.test.ts
@@ -1422,6 +1422,135 @@ describe('MFA', () => {
     expect(aalData!.currentAuthenticationMethods).toBeDefined()
   })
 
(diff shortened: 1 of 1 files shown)


mandarini:
@7ttp can you please rebase your PR with latest `master`? We made some updates in the workflows, and it will be blocked if you don't rebase

PR #1948, opened 2025-12-14
feat(auth): add suppressedPaths option for non-supabase oauth flows

7ttp:
adds suppressedPaths option to skip implicit grant detection on specific paths.

useful when using third party oauth (like facebook login) that returns tokens in the url, prevents supabase from intercepting tokens not meant for it and logging users out.


7ttp:
hey @mandarini 😥
wanted to check if this approach makes sense for handling non supabase oauth flows.
basically the idea is letting users specify paths where supabase skips token detection so third party callbacks (fb, insta, etc) don't get intercepted.
happy to change/continue things up 
if you have a different approach in mind!

Issue #1883, opened 2025-11-21
Support for setting the Content-Encoding header when uploading files to Storage


miwelc:
### Description

I've searched in the documentation but I haven't found a way to set the Content-Encoding header when uploading files to Storage.
In our use case, we are storing in Supabase Storage big files that would compress nicely with gzip (compressed files are just ~10% of uncompressed size).
We upload these files from an Edge function and consume them in the browser.
If we could set the Content-Encoding we could save bandwidth and storage space by storing them compressed and then letting the browser ungzip the file when fetching it.

### Suggested solution

```javascript
const file = await Deno.open(localPath, { read: true });
const compressedStream = file.readable.pipeThrough(new CompressionStream('gzip'));
await supabase.storage
	.from('my_bucket')
	.upload(remotePath, compressedStream, {
		contentType: 'application/octet-stream',
		contentEncoding: 'gzip'
	});
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://supabase.com/docs/reference/javascript/introduction).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

yoshifumi-kondo:
## Investigation Results

I investigated adding `contentEncoding` support to the storage-js client and found that **the backend (storage-api) does not currently support Content-Encoding for uploads**.

### Findings

1. **PR #169** ([supabase/storage-api#169](https://github.com/supabase/storage/pull/169)) proposed adding `contentEncoding`, `contentLanguage`, and `contentDisposition` support, but **it was closed without being merged**.

2. **Current backend state** (verified in latest storage-api code):
   - The `FileUpload` interface in `uploader.ts` does not include `contentEncoding`
   - The `fileUploadFromRequest()` function does not extract `contentEncoding` from FormData or headers
   - The `Uploader.upload()` method only passes `mimeType`, `cacheControl`, and `userMetadata` to the storage backend
   - Even the S3 protocol routes extract `ContentEncoding` from requests but then ignore it in `putObject()`

3. **Code evidence** (from `src/storage/uploader.ts`):
   ```typescript
   interface FileUpload {
     body: Readable
     mimeType: string
     cacheControl: string
     isTruncated: () => boolean
     xRobotsTag?: string
[...]

PR #1949, opened 2025-12-15
feat(storage): add content-encoding support for uploads

Athaxv:
Add content-encoding support for uploads

## 🔍 Description

- Add `contentEncoding` to `FileOptions` and apply the `Content-Encoding` header for uploads and signed uploads when provided.
- Ensure streaming / non-FormData uploads can declare gzip (or other) encodings.
- Add unit coverage verifying the header is sent.

### What changed?

- Extended `FileOptions` with an optional `contentEncoding` field.
- Updated `StorageFileApi` upload and `uploadToSignedUrl` to set the `Content-Encoding` header when `contentEncoding` is passed.
- Added tests in `storageFileApi.test.ts` to assert the correct headers are sent for uploads using streams/buffers.

- Storage currently doesn’t expose a way to set `Content-Encoding` on uploads.
- This is required to support use cases where files are uploaded already compressed (e.g. gzip) from Edge Functions and should be served correctly in browsers.
- Implements the feature requested in #1883.

- No breaking changes. The new `contentEncoding` option is optional and only affects behavior when explicitly provided.

### DIFF
--- a/packages/core/storage-js/test/storageFileApi.test.ts
+++ b/packages/core/storage-js/test/storageFileApi.test.ts
@@ -35,8 +35,6 @@ const findOrCreateBucket = async (name: string, isPublic = true) => {
 
 const uploadFilePath = (fileName: string) => path.resolve(__dirname, 'fixtures', 'upload', fileName)
 
-
-
 describe('Object API', () => {
   let bucketName: string
   let file: Buffer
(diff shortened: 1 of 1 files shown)


mandarini:
Thanks for the PR @Athaxv! This is a great feature request and the client-side implementation looks good.

However, the storage server's REST API (`/object/*` routes) doesn't currently handle the Content-Encoding header. The client will send it, but the server will ignore it.

Specifically, in the storage server:
  - `fileUploadFromRequest()` only extracts `content-type`, `cache-control`, `x-robots-tag`, and `x-metadata` headers
  - The `uploadObject()` backend method doesn't have a `contentEncoding` parameter
  - The S3 Upload params don't include `ContentEncoding`

The S3-compatible API routes (/s3/*) already support `Content-Encoding`, it's just the REST API that's missing this.

You can open a feature request on the [supabase/storage](https://github.com/supabase/storage) repo to add `Content-Encoding` support to the REST API routes. Once that's implemented and released, we can merge this PR and the feature will work end-to-end.

You can reference this PR and issue #1883 in that request. The server-side changes would involve:
  1. Extracting `content-encoding` header in `fileUploadFromRequest()`
  2. Adding `contentEncoding` parameter to the `uploadObject()` method
  3. Passing `ContentEncoding` to the S3 Upload params

I'll keep this PR open and we can revisit once the server-side support lands.

PR #1956, opened 2025-12-16
fix(postgrest): add explicit response type for maybeSingle()

MuzzaiyyanHussain:
Adds a dedicated PostgrestMaybeSingleResponse type to better represent the
existing runtime behavior of maybeSingle().

This is a type-only change and does not modify any runtime logic.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -2,7 +2,7 @@ import type {
   PostgrestSingleResponse,
   PostgrestResponseSuccess,
   CheckMatchingArrayTypes,
-  PostgrestMaybeSingleResponse, 
+  PostgrestMaybeSingleResponse,
   MergePartialResult,
   IsValidResultOverride,
 } from './types/types'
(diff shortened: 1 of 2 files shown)


mandarini:
@MuzzaiyyanHussain thanks for your PR. I'll review more in a bit. Can you please use `npm i` instead of `pnpm i`? And please delete `pnpm-lock.yml`.

mandarini:
Hey @MuzzaiyyanHussain, thanks for the contribution, but I don't think this PR provides a meaningful change. Let me explain:

**The Types Are Semantically Identical**

The current definition:

```ts
  export type PostgrestMaybeSingleResponse<T> = PostgrestSingleResponse<T | null>
```

Expands to (since PostgrestSingleResponse<T> = PostgrestResponseSuccess<T> | PostgrestResponseFailure):  PostgrestResponseSuccess<T | null> | PostgrestResponseFailure

Your proposed change:

```ts
  export type PostgrestMaybeSingleResponse<T> =
    | PostgrestResponseSuccess<T | null>
    | PostgrestResponseFailure
```
  
These are exactly the same type. TypeScript will treat them identically - there's no difference for consumers of the library.

**Unused Import**

The PR adds an import of PostgrestMaybeSingleResponse in PostgrestBuilder.ts, but it's not used anywhere. This would likely fail linting.

I'm going to close this PR since it doesn't change any behavior or types. However, if you're experiencing an actual type issue with maybeSingle(), I'd encourage you to:

  1. Open an issue first describing the problem you encountered
  2. Include a code example showing the type error or unexpected behavior
  3. Then we can discuss the right solution

If there's a real typing problem with maybeSingle(), we definitely want to fix it - we just need to understand what issue you're trying to solve. Thanks for your interest in contributing to Supabase! 💚

PR #1958, opened 2025-12-16
feat(auth): allow custom predicate for detectSessionInUrl option

mandarini:
## Summary

  - Extends `detectSessionInUrl` option to accept a custom predicate function
  - Allows users to exclude specific URLs from Supabase's implicit grant detection
  - Fixes issue where non-Supabase OAuth flows (e.g., Facebook Login) were incorrectly intercepted

## Problem

When users have other OAuth providers that return `access_token` in the URL fragment (like Facebook Login for connecting Instagram accounts), Supabase Auth intercepts these tokens and attempts to process them, causing unintended session logouts.

## Solution

The `detectSessionInUrl` option now accepts either:
  - `boolean` - existing behavior (default: `true`)
  - `function(url: URL, params: object) => boolean` - custom predicate

Example usage:

```ts
    auth: {
      detectSessionInUrl: (url, params) => {
        // Ignore Facebook OAuth redirects
        if (url.pathname === '/facebook/redirect') return false
        // Default behavior for Supabase callbacks
        return Boolean(params.access_token || params.error_description)
      }
    }
  })
```

Related: 
* https://github.com/supabase/supabase-js/issues/1697
* https://github.com/supabase/auth/pull/2299

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -2109,12 +2109,7 @@ export default class GoTrueClient {
    */
   private _isImplicitGrantCallback(params: { [parameter: string]: string }): boolean {
     if (typeof this.detectSessionInUrl === 'function') {
-      try {
-        return this.detectSessionInUrl(new URL(window.location.href), params)
-      } catch (e) {
-        this._debug('_isImplicitGrantCallback', 'error in detectSessionInUrl function', e)
-        return false
-      }
+      return this.detectSessionInUrl(new URL(window.location.href), params)
     }
     return Boolean(params.access_token || params.error_description)
   }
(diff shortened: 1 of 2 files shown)


packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -2100,8 +2102,20 @@ export default class GoTrueClient {
...
+   * if the URL should be processed as a Supabase auth callback. This allows users to exclude
+   * URLs from other OAuth providers (e.g., Facebook Login) that also return access_token in the fragment.
    */
   private _isImplicitGrantCallback(params: { [parameter: string]: string }): boolean {
+    if (typeof this.detectSessionInUrl === 'function') {
+      try {
+        return this.detectSessionInUrl(new URL(window.location.href), params)
+      } catch (e) {
+        this._debug('_isImplicitGrantCallback', 'error in detectSessionInUrl function', e)
+        return false
```
hf: To me it feels like this should crash instead of return false. Make the developer aware there's a bug in their implementation.

PR #1962, opened 2025-12-16
fix(auth): add configurable lock acquisition timeout to prevent deadlocks

### LINKED ISSUE #1594: [Auth] GoTrueClient uses infinite timeouts for lock acquisition causing deadlocks in production
# Bug report

## Describe the bug

I encountered a situation where Supabase Auth never finish initializing.

- The impact on my app is that **it never finishes initializing** and **any auth operation hangs**.

- I already checked that `onAuthStateChange` is not async and doesn't call any Supabase API, so #762 doesn't apply.

- The `GoTrueClient` initialization hangs indefinitely on some devices (in my case, Chrome Android) due to a lack of timeouts in the Web Locks API implementation.

- I investigated this by digging into the source code and enabling the undocumented `debug` option. On the affected device, the `#_acquireLock` call never resolved

  | On a working device, the debug log is normal | On a problematic device, it got stuck at `#_acquireLock` |
  | --- | --- |
  | Image | Image |

  Additionally, on a problematic device, checking `supabase.auth.initializePromise` returns a never-resolving promise.

- It seems that a previous lock wasn't properly released… when I run `await navigator.locks.query()` I see that some locks are being held indefinitely.

  | Held locks |
  | ---------- |
  | Image |

- **Is it really a good idea to wait for a lock indefinitely??** Sorry if I sound frustrated, but I kinda am – I just paid for Supabase Pro and didn't expect this from a professional auth solution. I kinda spent an hour scratching my head figuring out what's wrong with my app. That said, I really appreciate the overall quality of Supabase and believe this is just an edge case to be fixed. I hope this investgation will be useful.

## To Reproduce

This issue is not easily reproducible as it appears to be race condition or environment-dependent. However, it occurs in production on real devices where users experience complete authentication failure.

The **fundamental issue** is that the current implementation holds Web Locks indefinitely without a timeout, which violates basic concurrency principles. Any scenario that prevents proper lock cleanup can trigger this deadlock.

## Expected behavior

Supabase Auth should be able to recover from a situation where the lock is not properly released. The lock mechanism should have a timeout and fallback to ensure the client doesn't hang indefinitely.

Currently, the there is no `acquireTimeout` (-1 means no timeout):

https://github.com/supabase/auth-js/blob/aadf02e63179746a06451f4247a370dfd05740ea/src/GoTrueClient.ts#L358-L362

Maybe it should be set to some sane value, like maybe 1 minute.

## Screenshots

| The app doesn't launch as Auth never finished initializing, and User sees a white page. |
| ------ |
| Image |

## System information

- OS: Android
- Browser: Chrome 
- Version of supabase-js: 2.71.1
- Affected code: `@supabase/auth-js` GoTrueClient

### Current workaround

Until a proper lease is implemented, I am currently working around this issue by specifying a `noOpLock` when initializing Supabase. However I don’t know what might go wrong if we skip the locks.

```javascript
const noOpLock = async (name: string, acquireTimeout: number, fn: () => Promise<any>) => {
  return await fn()
}

  auth: { lock: noOpLock }
})
```

### Debug session and debug log

**Checking the locks in JS Console**
```javascript
await navigator.locks.query()

{
  held: [
    {clientId: '0ccb43ce-27c2-423e-8cbe-63a9c2da37b4', mode: 'exclusive', name: 'lock:sb-nmzhiwafofgwvqucnbds-auth-token'}
  ],
  pending: [
    {clientId: '0ccb43ce-27c2-423e-8cbe-63a9c2da37b4', mode: 'exclusive', name: 'lock:sb-nmzhiwafofgwvqucnbds-auth-token'},
    {clientId: 'e6ee664e-e2d9-4f1d-8a9a-0ab0b5e83ee6', mode: 'exclusive', name: 'lock:sb-nmzhiwafofgwvqucnbds-auth-token'},
    {clientId: 'e6ee664e-e2d9-4f1d-8a9a-0ab0b5e83ee6', mode: 'exclusive', name: 'lock:sb-nmzhiwafofgwvqucnbds-auth-token'},
    {clientId: '9d09c2ce-ac43-4783-b356-e988cae6bc26', mode: 'exclusive', name: 'lock:sb-nmzhiwafofgwvqucnbds-auth-token'}
  ]
}
```

**Debug logs showing the hang:**
```
GoTrueClient@0 (2.71.1) 2025-09-13T19:24:55.355Z #_acquireLock begin -1
GoTrueClient@0 (2.71.1) 2025-09-13T19:24:55.355Z #onAuthStateChange() registered callback with id a5571c4b-9f69-4111-b314-119f2e0ecb87
(hangs here - never proceeds to lock acquisition)
```

yoshifumi-kondo:
## Summary

- Add `lockAcquireTimeout` option to `GoTrueClientOptions` (default: 60000ms)
- Replace infinite lock timeout (-1) with configurable timeout to prevent production deadlocks
- Users can customize timeout or set negative for infinite wait (not recommended)

Fixes #1594

## Test plan

- [x] Build passes (`npx nx build auth-js`)
- [x] Existing tests pass
- [x] Manual testing with concurrent browser tabs

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -174,7 +174,7 @@ const DEFAULT_OPTIONS: Omit<
   debug: false,
   hasCustomAuthorizationHeader: false,
   throwOnError: false,
-  lockAcquireTimeout: 60000, // 1 minute
+  lockAcquireTimeout: 10000, // 10 seconds
 }
 
 async function lockNoOp<R>(name: string, acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
--- a/packages/core/auth-js/src/lib/locks.ts
+++ b/packages/core/auth-js/src/lib/locks.ts
@@ -227,6 +227,12 @@ export async function processLock<R>(
       acquireTimeout >= 0
         ? new Promise((_, reject) => {
             setTimeout(() => {
+              console.warn(
+                `@supabase/gotrue-js: Lock "${name}" acquisition timed out after ${acquireTimeout}ms. ` +
+                  'This may be caused by another operation holding the lock. ' +
+                  'Consider increasing lockAcquireTimeout or checking for stuck operations.'
+              )
+
               reject(
                 new ProcessLockAcquireTimeoutError(
                   `Acquiring process lock with name "${name}" timed out`
--- a/packages/core/auth-js/src/lib/types.ts
+++ b/packages/core/auth-js/src/lib/types.ts
@@ -113,9 +113,39 @@ export type GoTrueClientOptions = {
    */
   throwOnError?: boolean
   /**
-   * The maximum time in milliseconds to wait for a lock to be acquired.
-   * If the lock cannot be acquired within this time, an error with `isAcquireTimeout` set to true is thrown.
-   * Defaults to 60000 (1 minute). Set to a negative value to wait indefinitely (not recommended).
+   * The maximum time in milliseconds to wait for acquiring a cross-tab synchronization lock.
+   *
+   * When multiple browser tabs or windows use the auth client simultaneously, they coordinate
+   * via the Web Locks API to prevent race conditions during session refresh and other operations.
+   * This timeout controls how long to wait for the lock before failing.
+   *
+   * If the lock cannot be acquired within this time, a `LockAcquireTimeoutError` is thrown.
+   * You can catch this by checking `error.isAcquireTimeout === true`.
+   *
+   * - **Positive value**: Wait up to this many milliseconds before timing out
+   * - **Zero (0)**: Fail immediately if the lock is unavailable
+   * - **Negative value**: Wait indefinitely (not recommended - can cause deadlocks)
+   *
+   * @default 10000
+   *
+   * @example
+   * ```ts
+   * const client = createClient(url, key, {
+   *   auth: {
+   *     lockAcquireTimeout: 10000, // 10 seconds
+   *   },
+   * })
+   *
+   * try {
+   *   await client.auth.getSession()
+   * } catch (error) {
+   *   if (error.isAcquireTimeout) {
+   *     // Lock held by another tab/instance, or a previous operation is stuck.
+   *     // Consider: closing other tabs, increasing timeout, or restarting the browser.
+   *     console.error('Could not acquire lock within timeout period.')
+   *   }
+   * }
+   * ```
    */
   lockAcquireTimeout?: number
 }
--- a/packages/core/auth-js/test/GoTrueClient.test.ts
+++ b/packages/core/auth-js/test/GoTrueClient.test.ts
@@ -3232,6 +3232,80 @@ describe('Lock functionality', () => {
     await expect(client._acquireLock(1000, mockFn)).rejects.toThrow('Lock acquisition timeout')
     expect(mockFn).not.toHaveBeenCalled()
   })
+
+  test('should use custom lockAcquireTimeout when provided', async () => {
+    const capturedTimeouts: number[] = []
+    const mockLock = jest
+      .fn()
+      .mockImplementation(async (name: string, timeout: number, fn: () => Promise<unknown>) => {
+        capturedTimeouts.push(timeout)
+        return fn()
+      })
+
+    const customTimeout = 20000 // 20 seconds (different from default 10s)
+
+    const client = new GoTrueClient({
+      url: GOTRUE_URL_SIGNUP_ENABLED_AUTO_CONFIRM_ON,
+      lock: mockLock,
+      autoRefreshToken: false,
+      persistSession: false,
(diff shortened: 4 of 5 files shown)


yoshifumi-kondo:
@mandarini
Thank you for the thoughtful review and clear feedback! I've addressed all the requested changes.

## 1. Enhanced JSDoc Documentation 

Added comprehensive JSDoc to `lockAcquireTimeout` option in `types.ts`:
```typescript
/**
 * The maximum time in milliseconds to wait for acquiring a cross-tab synchronization lock.
 *
 * When multiple browser tabs or windows use the auth client simultaneously, they coordinate
 * via the Web Locks API to prevent race conditions during session refresh and other operations.
 * This timeout controls how long to wait for the lock before failing.
 *
 * If the lock cannot be acquired within this time, a `LockAcquireTimeoutError` is thrown.
 * You can catch this by checking `error.isAcquireTimeout === true`.
 *
 * - **Positive value**: Wait up to this many milliseconds before timing out
 * - **Zero (0)**: Fail immediately if the lock is unavailable
 * - **Negative value**: Wait indefinitely (not recommended - can cause deadlocks)
 *
 * @default 10000
 *
 * @example
 * try {
 *   await client.auth.getSession()
 * } catch (error) {
 *   if (error.isAcquireTimeout) {
 *     // Lock held by another tab/instance, or a previous operation is stuck.
 *     // Consider: closing other tabs, increasing timeout, or restarting the browser.
 *     console.error('Could not acquire lock within timeout period.')
 *   }
 * }
 */
```

## 2. Add Tests for Timeout Behavior 

Added tests in two locations:

**locks.test.ts** (actual timeout behavior):
- `should throw LockAcquireTimeoutError when lock acquisition times out` — verifies real timeout with `isAcquireTimeout: true`
- `should fail immediately when acquireTimeout is 0 and lock is held`

**GoTrueClient.test.ts** (configuration integration):
- `should use custom lockAcquireTimeout when provided`
- `should use default lockAcquireTimeout (10000ms) when not provided`
- `should pass negative timeout to lock for indefinite wait`

## 3. Document User Recovery 

Added recovery guidance in:
- JSDoc example shows how to catch and handle the error
- `console.warn` message in `processLock` includes actionable guidance:
```
Lock "{name}" acquisition timed out after {timeout}ms.
This may be caused by another operation holding the lock.
Consider increasing lockAcquireTimeout or checking for stuck operations.
```

## 4. Answers to Questions

**Q: Is 60 seconds too long as default?**

I initially chose 60 seconds to account for edge cases like slow networks or heavily loaded devices where lock operations might take longer than usual.

However, your point about user experience is valid — in actual deadlock situations, 60 seconds is far too long for users to wait on a white screen. Since lock operations typically complete in milliseconds, a shorter timeout still provides adequate buffer for edge cases while giving users faster feedback.

Changed to 10 seconds (10000ms) as a balance between these considerations.

**Q: Should we add console logging?**

Added `console.warn` to `processLock` when timeout occurs. Includes lock name and timeout value for debugging context.

**Q: Should we add a retry mechanism?**

For now, users can implement custom retry logic by catching `error.isAcquireTimeout === true`.

I believe automatic retry should be considered in a follow-up PR because:
1. This PR's scope is deadlock prevention via configurable timeout (Issue #1594)
2. Retry mechanism requires careful design decisions (retry count, backoff strategy, configurability)
3. Different use cases have different retry requirements

If there's interest, I'm happy to open a follow-up issue to discuss retry mechanism design.

mandarini (approved):

Owez:
Thanks for this :) any idea what release this'll be in? Right now we're using the noop workaround

tadeumaia:
My React Native app is exploding with timeout errors now. What is the best practice for RN?  should i hunt all that is causing deadlocks or the previous behavior good enough for apps?

mandarini:
@tadeumaia what version are you using? Do you think this PR caused your issue?

tadeumaia:
@mandarini Yeah, when I updated to latest I started getting a bunch of warning for @supabase/gotrue-js: Lock "lock:sb-tvjkvpkckfxgxdhyggiy-auth-token" acquisition timed out after 10000ms. This may be caused by another operation holding the lock. Consider increasing lockAcquireTimeout or checking for stuck operations. 

I'm on "@supabase/supabase-js": "^2.90.1" was on "^2.87.1", before. 

Just no idea what the best practice and if the lock is necessary for RN or only for browser (multi tab stuff) so I should disable it or hunt for places where I might be calling locks too much.

Owez:
I've been trying out the new canary version in prod and it seems to have mainly fixed the main issue of deadlocks, but I get the feeling like theres something odd going on internally for it to be causing it. I still seem to be getting the occasional total lockout still, but its performing better than before (say 1 per day vs ~5). Could be a fault on my end, though it doesn't feel seemless.

Total vibes tell me its multiple open sessions at the same time, but I'm assuming getSession is locally JWT checking so that shouldn't matter

Though I'm just an end user and probs can't reproduce on a public repo 😔

Issue #1640, opened 2025-02-23
Unexpected behavior for multiple operations on the same table


rkistner:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Multiple operations on the same "table" (PostgrestQueryBuilder instance) share state unexpectedly. If this is by design, it should be documented.

## To Reproduce

Run the following code:

```js
const table = supabase.from('mytable');
await table.update({a: 1}).eq('id', 1);
await table.update({a: 2}).eq('id', 2);
```

I'd expect this to do two separate operations: `update mytable set a = 1 where id = 1`, then `update mytable set a = 2 where id = 2`. What actually happens is the second operation is the equivalent of `update mytable set a = 2 where id = 1 and id = 2`, resulting in nothing being updated. Request logs show the search string for the PATCH request as `?id=eq.1&id=eq.2`.

## Expected behavior

The two operations on `table` should be independent, and not affect each other.

## Screenshots

N/A

## System information

- OS: N/A
- Browser: Tested on Chrome
- Version of supabase-js: 2.48.1 / postgrest-js 1.18.1
- Version of Node.js: N/A

## Additional context

There is a simple workaround - refactor the code as follows:

```js
await supabase.from('mytable').update({a: 1}).eq('id', 1);
await supabase.from('mytable').update({a: 2}).eq('id', 2);
```

It's difficult to tell whether this behavior is by design or a bug. But it's definitely not expected, and leads to difficult-to-debug issues.

I also tested the Dart [postgrest package](https://pub.dev/packages/postgrest), and there the same code _does_ work as expected (two independent operations).

I tested with `update` specifically, but I assume other operations such as `select` will likely have the same issue.

PR #1978, opened 2025-12-19
fix(postgrest): prevent shared state between query builder operations

7ttp:
when reusing a `PostgrestQueryBuilder` instance returned by `.from()`, 
url search params and headers were being shared across multiple operations, 
causing filters to accumulate unexpectedly.
this fix ensures each operation receives its own cloned state.

fixes https://github.com/supabase/supabase-js/issues/1640


7ttp:
Added test for shared state isolation between query builder operations. 
Ensures filters don't leak when reusing the same table reference.

PR #1982, opened 2025-12-23
feat(realtime): expose heartbeat latency on heartbeat callback

edgurgel:
## 🔍 Description

Expose latency for heartbeats when we get a reply. Only available for status `ok` and `error` otherwise `undefined`. It's backwards compatible because the second argument is `number | undefined` so a callback with just one argument will stil work fine. 

I've also changed so that it matches the heartbeatRef when the message arrives. This avoids any potential race condition where the message for a previous heartbeat arrives later but we had already given up on that previous heartbeat.

### What changed?

An extra `latency` can be passed to the heartbeat callback when heartbeat gets a reply from the server. RealtimeClient keeps track of the last time the heartbeat was sent and provide the diff.

A feature that can help people react to high latency users like people on slow mobile internet. 

## 🔄 Breaking changes

## 📋 Checklist

### DIFF
--- a/packages/core/realtime-js/src/RealtimeClient.ts
+++ b/packages/core/realtime-js/src/RealtimeClient.ts
@@ -75,7 +75,7 @@ export type RealtimeClientOptions = {
   transport?: WebSocketLikeConstructor
   timeout?: number
   heartbeatIntervalMs?: number
-  heartbeatCallback?: (status: HeartbeatStatus) => void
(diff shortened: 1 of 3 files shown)


edgurgel:
I've updated multiplayer.dev to use that instead of a custom room for the latency display:

https://github.com/supabase/multiplayer.dev/pull/10

https://realtime-demo-ka5nlzyra-supabase.vercel.app/

mandarini (approved):

PR #1983, opened 2025-12-23
test(postgrest): use supabase cli for tests and type gen

mandarini:
Switch to using Supabase CLI instead of docker compose for postgrest-js tests.

Related: https://github.com/supabase/supabase-js/pull/1981

### DIFF
--- a/packages/core/postgrest-js/project.json
+++ b/packages/core/postgrest-js/project.json
@@ -19,7 +19,7 @@
     "test:suite": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "jest --runInBand --coverage",
+        "command": "jest --runInBand --coverage --testPathIgnorePatterns='test/v12/'",
         "cwd": "{projectRoot}"
       },
       "outputs": ["{projectRoot}/coverage"],


packages/core/postgrest-js/test/supabase/.branches/_current_branch
```diff
@@ -0,0 +1 @@
+main
```
grdsdev: this should be ignored per gitignore

packages/core/postgrest-js/test/supabase/.temp/cli-latest
```diff
@@ -0,0 +1 @@
+v2.67.1
```
grdsdev: this should be ignored per gitignore

packages/core/postgrest-js/test/relationships-spread-operations.test.ts
```diff
@@ -326,42 +322,8 @@ test('select with spread on nested relation', async () => {
   ExpectedSchema.parse(res.data)
 })
 
-test('select spread on many relation postgrest14', async () => {
```
grdsdev, line 329: Are pg14 tests not needed anymore?

packages/core/postgrest-js/test/relationships-join-operations.test.ts
```diff
@@ -1063,23 +1063,31 @@ test('inner join on many relation', async () => {
...
           },
         ],
       },
       "error": null,
       "status": 200,
       "statusText": "OK",
     }
-  `)
+  `
+  )
```
grdsdev: I think matching the exact response is good here, does the id changes between runs? Otherwise, I think sticking with matching whole object is better.

packages/core/postgrest-js/test/relationships-error-handling.test.ts
```diff
@@ -261,33 +261,6 @@ test('typecasting and aggregate', async () => {
   expectType<TypeEqual<typeof result, typeof expected>>(true)
 })
 
-test('select spread on many relation', async () => {
```
grdsdev, line 264: Is this being tested somewhere else?

grdsdev (dismissed):

packages/core/postgrest-js/test/relationships-spread-operations.test.ts
mandarini: The test wasn't removed, it was consolidated. Before, the tests ran two separate PostgREST Docker container (postgrest v12.2.0, postgrest v14.0). There were duplicate tests for the same functionality:
  - 'select spread on many relation postgrest14' - using postgrest14 client
  - 'select spread on many relation postgrest14FromDatabaseTypes'

The tests now use supabase cli which bundles a postgrest version 14. The test was renamed to just 'select spread on many relation' and uses the single postgrest client pointing to `http://localhost:54321/rest/v1`. The test functionality is same query. Supabase cli uses postgrest 14+. We no longer need separate "pg12" vs "pg14" tests since the old v12 container was removed entirely.

packages/core/postgrest-js/test/relationships-error-handling.test.ts
mandarini: The "error" test was removed because postgrest v14 made spreading many relations a valid feature instead of an error. The feature is still tested but in the success tests ([here](https://github.com/supabase/supabase-js/blob/test/supabase-cli-postgrest/packages/core/postgrest-js/test/relationships-spread-operations.test.ts#L325)), not the error tests. Postgrest behavior changed between versions. Postgrest 12 gave error "spread not possible", Postgrest 14 returns arrays, no error. The old docker-compose ran both v12 and v14, so it needed both tests. Now with Supabase cli (which uses postrest v14+), the error test was removed because this is no longer an error.

avallete: Just a comment on here, that was intended as we can't control that the user have a 1-1 matching version between their `postgrest-js` and their actual `supabase`/Postgrest infra (user could run on Postgrest12 while having the latest `postgrest-js` version as a deps). So those kind of tests on two versions were mostly there to ensure no breaking changes and retro-compatibility of the types with older versions, mostly because the "major release" was something we avoided.

If that's not the case anymore and we do release breaking changes into majors more often, then I think they can go away. But might wanna add the compatibles versions between the client lib and the Postgrest version somewhere to the docs / changelogs, so users having diffs can identify that it comes from not having the latest `client libs`/`postgrest version` combo.

packages/core/postgrest-js/test/relationships-join-operations.test.ts
mandarini: Reverted! Good point!

packages/core/postgrest-js/test/relationships-error-handling.test.ts
mandarini: Oh I see. So in this case there would be no way to only use the Supabase CLI to test this right? I would still need to keep the docker compose configuration for the v12? @avallete

mandarini: I can restore some v12 tests, and add plan to remove these in v3. what do you think @avallete @grdsdev ?

avallete: I think we can still approach this in phases, since I do see the benefit of unifying the run:

1. Start by keeping the typing part of this test as a .d.ts for backward compatibility, to ensure we don’t break this behavior in the future. Historically, we merged runtime and type tests together because it makes discrepancies between runtime results and type inference much easier to spot.
2. Drop v12 support in the next major version? We’d need input from the PostgREST team to confirm whether this is feasible, as I believe it’s still used by some projects on the platform.

Also, one alternative could be to leverage `supabase/.temp/postgrest-version` to switch locally between 14 and 12 if needed. So I think we can totally go ahead with those changes.

mandarini: Ok, I added back some compatibility tests for v12!

avallete (dismissed):

matthias-luft (approved):

PR #1988, opened 2025-12-24
fix(functions): auto-stringify object body when custom Content-Type header is provided

### LINKED ISSUE #1587: FunctionsClient.invoke not carrying body content
# Bug report

## Describe the bug

The js call to invoke an edge function didn't send body in the request. This has caused me a lot of time to debug. I found this link https://github.com/orgs/supabase/discussions/19148 has reported the issue and it seems not being fixed. Used raw fetching did the work.

Evidence: the content length is 0.

Image

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

```
const headers = await this.getAuthHeaders();
            
const { data, error } = await supabase.functions.invoke(functionName, {
    body,
    headers
});
```

The body was somehow not placed into the request.

## Expected behavior

The body should carry the json object to the server.

## Screenshots

See above. Tested both on Chrome and Edge.

7ttp:
Extends #1758 to also stringify object bodies when a custom Content-Type header is provided. Without this, passing a plain object results in [object Object] being sent instead of the actual JSON.

Closes #1587

### DIFF
--- a/packages/core/functions-js/test/spec/params.spec.ts
+++ b/packages/core/functions-js/test/spec/params.spec.ts
@@ -491,3 +491,56 @@ describe('params reached to function', () => {
     expect(data).toMatchObject(expected)
   })
 })
+
+describe('body stringify with custom headers', () => {
+  test('should stringify object body when custom Content-Type is provided', async () => {
+    const mockFetch = jest.fn().mockResolvedValue(new Response('ok'))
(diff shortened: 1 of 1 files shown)


7ttp:
Added tests for object body stringify with custom headers, nested objects, and string body passthrough.

PR #1994, opened 2025-12-29
fix(postgrest): use post with return minimal for rpc head requests with object args

7ttp:
fixes object args being serialized as `[object Object]` when using `rpc()` with `head: true`

when `head: true` is used with object type arguments, the client now uses POST with `Prefer: return=minimal` header instead of HEAD, since HEAD requests cannot have a body and objects cannot be properly serialized to URL params.

closes https://github.com/supabase/supabase/issues/41641


mandarini (changes requested):
Updated.

PR #1997, opened 2026-01-01
chore(supabase): add unified SupabaseError type

### LINKED ISSUE #1414: Standardize supabase method response types
# Chore
add error handling callback in `createClient`, or have a standardized type that supabase calls adhere to. (ex. they should all return `{ data: T, error: SupabaseError }`

## Describe the chore

I'm writing a util that handles supabase errors, so instead of doing `const { data, error} = await suapabase.callSomething()` I can handle the errors automatically. But there are AuthErrors, PostgressErrors, and now i'm hitting OAuthErrors which are different somehow even though it conforms to the same response type as the other errors? 

It would be really nice if there was just one error type that users could specify if needed but was general enough that we can just grab the errors and expect data to come back without errors (unless we re-threw them) 

## Additional context

Add any other context or screenshots that help clarify the task.

Here is the function I have that works so far for most calls: 

```
export function handleSupabaseErrorAllowNull<T>(
  result: {
    data?: T
    error: AuthError | PostgrestError | null
  },
  description?: string,
): T | null {
  const { data, error } = result

  if (error) {
    if (description) {
      // do something
    }

    throw error
  }

  return data || null
}
```

passing an oauth function in here throws a type error though because of.... I'm guessing I need a new result type, but it would be easier if I could just pass all supabase calls in here and not worry about it

7ttp:
adds a unified SupabaseError type that combines AuthError, PostgrestError, FunctionsError, and StorageError into a single union type for easier generic error handling across all supabase operations.

closes #1414

### DIFF
--- a/packages/core/supabase-js/src/index.ts
+++ b/packages/core/supabase-js/src/index.ts
@@ -25,6 +25,7 @@ export type {
   QueryData,
   QueryError,
   DatabaseWithoutInternals,
+  SupabaseError,
 } from './lib/types'
 
 /**
--- a/packages/core/supabase-js/src/lib/types.ts
+++ b/packages/core/supabase-js/src/lib/types.ts
@@ -1,7 +1,8 @@
-import { GoTrueClientOptions } from '@supabase/auth-js'
+import { GoTrueClientOptions, AuthError } from '@supabase/auth-js'
 import { RealtimeClientOptions } from '@supabase/realtime-js'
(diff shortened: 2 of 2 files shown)


mandarini:
@7ttp I will not be merging this right now. It's planned to create a shared library for errors, and I will be working on that soon.

7ttp:
> @7ttp I will not be merging this right now. It's planned to create a shared library for errors, and I will be working on that soon.

Aight, Glad to know!💚

PR #1999, opened 2026-01-02
fix(realtime): validate table filter in postgres_changes event dispatch

7ttp:
when subscribing to multiple tables, events could be dispatched to the wrong callback. adds table validation to ensure payloads only reach their intended subscription.

closes https://github.com/supabase/supabase-js/issues/1721

### DIFF
--- a/packages/core/realtime-js/test/RealtimeChannel.postgres.test.ts
+++ b/packages/core/realtime-js/test/RealtimeChannel.postgres.test.ts
@@ -570,6 +570,48 @@ describe('Postgres Changes Trigger Tests', () => {
   })
 })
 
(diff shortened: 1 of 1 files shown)


mandarini:
@7ttp thank you for this PR! I like this fix. Before I merge it can we please add a test case that reproduces the bug by sending `ids: ['conv-id', 'msg-id']` with `data.table: 'messages'` and verifying only the messages callback fires? Thanks again!

mandarini (changes requested):
Requested test case

7ttp:
Added test to verify only the matching table's callback fires when multiple postgres_changes bindings share the same ids array.

PR #2004, opened 2026-01-07
chore(repo): do not fail pr label on forks

mandarini:
Cannot add label on PR from forks (needs github app setup, but maybe it's overkill so skipping for now). Handle error gracefully instead.


7ttp (approved):
😅 such a needed fix

PR #2011, opened 2026-01-08
fix(postgrest): preserve nullability when using JSON path

### LINKED ISSUE #1635: Null should be included in return type when using json path (->)

mandarini:
Original Author: @dev-hari-prasad
Moved from: https://github.com/supabase/supabase-js/pull/1939

Description:
**What changed?**  
Fixed the type inference for JSON path selections (`->`) by adding `| null` when the underlying JSON column can be null.

When selecting a JSON path from a nullable JSON column, the generated TypeScript types incorrectly assumed the value could never be null.  
In reality, the runtime returns `null` when the whole column is null, so the types were unsafe.  
This patch brings the inferred type in line with actual behavior.

Closes/Fixes #1635


avallete:
Just a feedback here, testing this out on some codebase result in some BC/bugs where things get infered as "null" only. Will work on making a MRE that can be added to the tests for it once I get the bandwidth for it.

mandarini:
@avallete what is the plan here?

avallete:
> @avallete what is the plan here?

Didn't had the bandwidth to work on it. Definitely hit a regression when canary testing over the `platform` repo at the time.<br><br>I need to get this regression reproduced in a MRE before passing this out (and making sure it's fixed).

mandarini:
Ok ok! No rush! I was just doing some triaging and ran into this!

dev-hari-prasad:
Hey @mandarini, so sorry this almost slipped my mind, I was on a break. I'll look into this immediately now that I'm free and do the needful.

dev-hari-prasad:
@mandarini, is there an expected date when this gets merged?

mandarini:
waiting updates from @avallete

PR #2014, opened 2026-01-10
fix(auth): defer subscriber notification in exchangeCodeForSession to prevent deadlock

### LINKED ISSUE #1566: exchangeCodeForSession() doesn't update Authorization headers on iOS/mobile platforms
# Bug: `exchangeCodeForSession()` doesn't update Authorization headers on iOS/mobile platforms

## Summary
After successfully calling `exchangeCodeForSession()` during OAuth flow on iOS native (Capacitor), the Supabase client fails to update its internal HTTP Authorization headers, causing all subsequent database queries to fail with authentication errors. The session is stored correctly and can be retrieved via `getSession()`, but the client's REST API headers remain unset.

## Environment
- **@supabase/supabase-js**: v2.39.7
- **Platform**: iOS native (Capacitor)
- **Framework**: React 18.3.1 with Vite
- **Capacitor**: v6.0.0
- **iOS Version**: 16.4+ (tested on iPhone 15 Pro simulator and physical devices)
- **OAuth Providers**: Google and Apple (both affected)

## Steps to Reproduce

1. Set up OAuth in a Capacitor iOS app:
```typescript
// OAuth setup for native iOS
async function startOAuthFlow(provider: 'google' | 'apple') {
  const { data, error } = await supabase.auth.signInWithOAuth({
    provider,
    options: {
      redirectTo: 'yourapp://auth/callback',
      skipBrowserRedirect: true,
    },
  });
  
  // Open in-app browser
  await Browser.open({ url: data.url });
}
```

2. Handle OAuth callback:
```typescript
// Handle deep link callback
async function handleOAuthCallback(url: string) {
  const code = new URL(url).searchParams.get('code');
  
  // Exchange code for session
  const { error } = await supabase.auth.exchangeCodeForSession(code);
  // Success! Session is stored
}
```

3. Attempt database query immediately after:
```typescript
// This query will fail even though auth succeeded
  .from('profiles')
  .select('*')
  .eq('id', userId)
  .single();
  
// Error: User not authenticated / JWT expired
```

## Expected Behavior
After `exchangeCodeForSession()` succeeds:
- The session should be stored in storage (this works)
- The Supabase client's HTTP headers should include `Authorization: Bearer [TOKEN]`
- Database queries should work immediately with authenticated session

## Actual Behavior
After `exchangeCodeForSession()` succeeds:
- The session IS stored correctly in storage (confirmed)
- The Supabase client's HTTP headers are NOT updated (problem)
- All database queries fail with authentication errors
- The client only works after a full page/app reload

## Console Logs Demonstrating the Issue

```log
TO JS {"url":"yourapp://auth/callback?code=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX#"}
[log] - [AuthStore] onAuthStateChange: event='SIGNED_IN', session exists= true
[log] - [AuthStore] onAuthStateChange: User signed in via google
[log] - [AuthStore] Supabase client headers AFTER OAuth: {
  "hasAuthHeader": false,
  "authHeaderPrefix": "undefined...",
  "sessionToken": "[REDACTED_TOKEN]...",
  "tokensMatch": false
}
[log] - [ProfileService] Current session state: {
  "hasSession": true,
  "userId": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
  "hasAccessToken": true,
  "tokenPrefix": "[REDACTED_TOKEN]...",
  "provider": "google"
}
[log] - [ProfileService] Supabase client headers: {
  "hasAuthHeader": false,
  "authHeaderPrefix": "undefined...",
  "apiKey": true
}
[error] - Error fetching user profile: {"code":"PGRST301","message":"JWT expired"}
```

## Debug Code to Verify Issue

```javascript
// Add this after exchangeCodeForSession() to see the problem
const { data: { session } } = await supabase.auth.getSession();
const currentHeaders = (supabase as any).rest?.headers;

console.log('Auth state after OAuth:', {
  hasValidSession: !!session,
  sessionToken: '[REDACTED_TOKEN]...',
  hasAuthHeader: !!currentHeaders?.Authorization,
  headerValue: currentHeaders?.Authorization?.substring(0, 30) + '...',
  tokensMatch: false // Would check if header contains the session token
});

// Output shows:
// hasValidSession: true (works)
// sessionToken: "[REDACTED_TOKEN]..." (exists)
// hasAuthHeader: false (PROBLEM)
// headerValue: "undefined..." (missing)
// tokensMatch: false (mismatch)
```

## Root Cause Analysis

The Supabase JavaScript client updates its internal HTTP headers during:
- Initial client creation (when a session exists in storage)
- Page reloads (which reinitialize the client from storage)
- Some auth operations (login with password, etc.)

However, it does NOT update headers after:
- `exchangeCodeForSession()` on mobile/native platforms

This works on web because OAuth redirects cause a natural page reload, which reinitializes the client with headers from the stored session.

On native platforms without page reload, the client maintains stale headers from before authentication.

## Current Workaround

Force a page reload immediately after OAuth on native platforms:

```javascript
// In onAuthStateChange listener
if (isNativeIOS() && event === 'SIGNED_IN' && session.user.app_metadata?.provider !== 'email') {
  // Force reload to reinitialize Supabase client with proper headers
  window.location.reload();
  return;
}
```

This workaround is not ideal as it:
- Disrupts user experience with a visible reload
- Adds latency to the authentication flow
- Is a hack for what should be handled internally

## Suggested Fix

The `exchangeCodeForSession()` method should internally update the client's Authorization headers after successfully setting the session:

## Impact

This bug affects:
- **All Capacitor/native mobile apps** using OAuth authentication with Supabase
- **User experience**: Users cannot access authenticated features without manual app restart
- **Onboarding flows**: New users get stuck after OAuth sign-up
- **Production readiness**: Makes OAuth unusable in production mobile apps

## Reproducibility
- **Consistency**: 100% reproducible
- **Platforms**: iOS (confirmed), likely affects Android native as well
- **Providers**: Both Google and Apple OAuth affected
- **Versions**: Confirmed on latest @supabase/supabase-js (2.39.7)

This is a critical bug that makes OAuth authentication unusable in production Capacitor apps without workarounds.

## Test Repository
I can provide a minimal reproduction repository if needed. The issue is consistent and easily reproducible with any Capacitor + Supabase OAuth setup.

---

**Please let me know if you need any additional information or clarification. This is blocking OAuth authentication in our production iOS app.**

7ttp:
## Summary

Defers `_notifyAllSubscribers` call in `_exchangeCodeForSession` using `setTimeout` to prevent deadlock on iOS/Capacitor PKCE OAuth flow.

## Problem

On iOS/Capacitor with PKCE flow, `exchangeCodeForSession()` would hang indefinitely when a subscriber callback (in `onAuthStateChange`) tried to call `getSession()` or make authenticated requests.

The deadlock occurred because:
1. `exchangeCodeForSession()` acquires a lock via `_acquireLock()`
2. Inside the lock, it calls `await _notifyAllSubscribers()` which awaits all subscriber callbacks
3. If a subscriber callback calls `getSession()`, it tries to acquire the same lock
4. The lock queues the `getSession()` call to wait for the current operation
5. But the current operation is waiting for the subscriber callback to complete
6. Deadlock: subscriber waits for `getSession()` which waits for lock which waits for subscriber

## Solution

Defer `_notifyAllSubscribers` using `setTimeout(..., 0)` so it runs after the lock is released. 
This matches the existing pattern used here: https://github.com/supabase/supabase-js/blob/master/packages/core/auth-js/src/GoTrueClient.ts#L521

## Related

Closes #1566

Thanks @simplysparsh

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -1146,7 +1146,9 @@ export default class GoTrueClient {
       }
       if (data.session) {
         await this._saveSession(data.session)
-        await this._notifyAllSubscribers('SIGNED_IN', data.session)
+        setTimeout(async () => {
+          await this._notifyAllSubscribers('SIGNED_IN', data.session)
+        }, 0)
       }
       return this._returnResult({ data: { ...data, redirectType: redirectType ?? null }, error })
     } catch (error) {


7ttp:
@mandarini @grdsdev

mandarini (approved):

PR #2015, opened 2026-01-11
fix(auth): prevent deadlock in updateUser

### LINKED ISSUE #1441: supabase.auth.setSession and supabase.auth.updateUser do not resolve unless there is an error
**Summary**
When using the Supabase client in an async function, both supabase.auth.setSession and supabase.auth.updateUser fail to resolve or return unless an error occurs. This blocks further execution in the code despite successful authentication or updates.

```
// Scenario 1: setSession
const { data, error } = await supabase.auth.setSession({
  access_token,
  refresh_token,
});
console.log('Setting session with token:', data, error); // Only logs if error occurs
console.log(data?.user); // Never reached if no error
```
```
// Scenario 2: updateUser
const { error: updateError } = await supabase.auth.updateUser({
  password: newPassword,
});
console.log('Update password error:', updateError); // Not called unless there's an error
```

Expected Behavior
Both methods should resolve regardless of whether there is an error, returning the appropriate data or null. Logging should be possible after the calls, and the behavior should be consistent with other auth methods like getSession.

Actual Behavior
setSession and updateUser appear to hang or do not return unless there's an error.

Causes the rest of the async function to not proceed, giving the illusion of being stuck.

Environment
Supabase JS SDK version: @supabase/supabase-js@2.9.9

Runtime: React Native / Expo (also affects Web SDK)

Additional Context
This issue is critical in flows like password recovery and token-based login where setSession or updateUser are required. It breaks user flows and leads to confusion.

ForestLinSen:
### What changed?

Modified the `_updateUser` method in `src/GoTrueClient.ts` to execute the `_notifyAllSubscribers` call asynchronously.

This fixes a deadlock issue where `updateUser` would hang indefinitely despite a successful server response.

I've got the same issue as issue #1441 . After digging into it, I found that the issue is because `updateUser` acquires a lock, and after success, it calls and `awaits _notifyAllSubscribers`. However, the lock is unavailable for the subscribers. So the subscriber waits for the lock, and updateUser waits for the subscriber => Deadlock

Closes #1441 

## 🔄 Breaking changes

## 📋 Checklist

- [ ] I have added tests for new functionality (if applicable)

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -1830,7 +1830,7 @@ export default class GoTrueClient {
         }
         session.user = data.user as User
         await this._saveSession(session)
-        await this._notifyAllSubscribers('USER_UPDATED', session)
+        void this._notifyAllSubscribers('USER_UPDATED', session).catch(console.error)
         return this._returnResult({ data: { user: session.user }, error: null })
       })
     } catch (error) {


mandarini:
@ForestLinSen thank you for the contribution! I went with this solution instead (https://github.com/supabase/supabase-js/pull/2014) because it matches existing patterns!

Once again, thank you for taking the time to contribute to Supabase! You're helping us make the tool better! 💚

PR #2019, opened 2026-01-14
fix(auth): prevent deadlock and unhandled rejections in processLock

mayur9210:
## Summary
Fixes unhandled promise rejections and potential deadlocks in `processLock` by replacing the `Promise.race` timeout pattern with a flag-based approach.

## Problem
The `processLock` function would crash Node.js with unhandled promise rejections when running tests with rapid successive lock operations that have mixed timeouts.

The crash occurred because:

1. `processLock` uses `Promise.race([lockOperation(), timeoutPromise])` to implement acquire timeout
2. When `lockOperation()` wins the race (lock acquired before timeout), the timeout promise still fires later
3. The timeout promise rejection goes unhandled, crashing Node.js
4. Additionally, the `PROCESS_LOCKS[name]` promise chain would re-throw errors, causing subsequent operations waiting on it to receive unhandled rejections or potentially deadlock

## Solution
- Replace `Promise.race` with a flag-based approach: set a `timeoutError` flag in the setTimeout callback, check it after acquiring the lock
- Clear the timeout immediately when the lock is acquired to prevent unnecessary timeout firing
- Ensure `PROCESS_LOCKS[name]` always resolves (never throws) so subsequent operations can proceed without deadlocking
- Fix test expectations: operation 0 acquires the lock immediately (no previous operation to wait on), so it completes successfully rather than timing out

## Related
Fixes the `ProcessLockAcquireTimeoutError: Acquiring process lock with name "rapid-test" timed out` crash in `@supabase/auth-js:test`

### DIFF
--- a/packages/core/auth-js/src/lib/locks.ts
+++ b/packages/core/auth-js/src/lib/locks.ts
@@ -196,20 +196,11 @@ const PROCESS_LOCKS: { [name: string]: Promise<any> } = {}
  * Useful for environments like React Native or other non-browser
  * single-process (i.e. no concept of "tabs") environments.
  *
- * Use {@link #navigatorLock} in browser environments.
- *
  * @param name Name of the lock to be acquired.
  * @param acquireTimeout If negative, no timeout. If 0 an error is thrown if
  *                       the lock can't be acquired without waiting. If positive, the lock acquire
- *                       will time out after so many milliseconds. An error is
- *                       a timeout if it has `isAcquireTimeout` set to true.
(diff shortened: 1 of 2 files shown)


mandarini:
We're working on another solution for a lockless sdk. Thank you very much for your contribution. I am going to close this PR, but keeping it in mind and in our list of potential fixes for our problematic lock mechanism, which will hopefully go away.

Please do not be discouraged from contritbuting again in the future! Contributions like yours are what make Supabase what it is today! 💚

PR #2021, opened 2026-01-14
fix(auth): add dispose method to cleanup BroadcastChannel

### LINKED ISSUE #2020: bug: BroadcastChannel event listener never removed causing memory leak
### Describe the bug

In `GoTrueClient`, a `BroadcastChannel` is created and an event listener is attached in the constructor, but there's no mechanism to remove the listener or close the channel. This causes memory leaks as the listener holds a reference to the client instance via closure.

### Expected behavior
The class should provide a dispose() method to properly clean up:

Remove the BroadcastChannel event listener
Close the BroadcastChannel
Stop auto refresh timers
Remove visibility change callback
Clear state change emitters
Note: The class already implements _removeVisibilityChangedCallback() for similar cleanup, but no equivalent exists for BroadcastChannel.

### Impact
Memory leaks in SPAs where clients are created/destroyed
Potential duplicate event handling with multiple client instances
Resources held unnecessarily after client is no longer needed

7ttp:
## Summary

Adds a `dispose()` method to `GoTrueClient` to properly clean up the `BroadcastChannel` event listener and close the channel.

## Problem

In `GoTrueClient`, a `BroadcastChannel` is created with an event listener in the constructor, but there's no way to remove the listener or close the channel. The cleanup pattern exists for `visibilityChangedCallback` (via `_removeVisibilityChangedCallback()`) but was missed for `BroadcastChannel`.

**Location:** `packages/core/auth-js/src/GoTrueClient.ts` (~line 398)

## Solution

- Store callback reference in `broadcastChannelCallback` property
- Add private `_removeBroadcastChannel()` method (matches existing pattern)
- Add public `dispose()` method for full resource cleanup

## Changes

- `GoTrueClient.ts`: Added `broadcastChannelCallback` property, `_removeBroadcastChannel()`, and `dispose()` methods
- `GoTrueClient.browser.test.ts`: Added test for dispose cleanup

## Related

Closes #2020

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -277,6 +277,7 @@ export default class GoTrueClient {
    * Used to broadcast state change events to other tabs listening.
    */
   protected broadcastChannel: BroadcastChannel | null = null
+  protected broadcastChannelCallback: ((event: MessageEvent) => Promise<void>) | null = null
 
   protected logDebugMessages: boolean
   protected logger: (message: string, ...args: any[]) => void = console.log
@@ -395,7 +396,7 @@ export default class GoTrueClient {
         )
       }
 
-      this.broadcastChannel?.addEventListener('message', async (event) => {
+      this.broadcastChannelCallback = async (event) => {
         this._debug('received broadcast notification from other tab or client', event)
 
         try {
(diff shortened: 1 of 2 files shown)


7ttp:
@mandarini

mandarini:
Hi @7ttp ! Thank you for the PR! I discussed it with the auth team and we decided that we will not be merging this PR. We do not see the reason why this needs cleaning. The general opinion is that it doesn’t solve a real problem, but in fact it may create more because someone will try to use it properly, fail, and then be “why isn’t X working”.

Thank you for contributing though! 💚

PR #2022, opened 2026-01-14
fix(storage): add no-store cache control to download

Akkii88:
**fix(storage): add no-store cache control to download**

**Description:**
This Pull Request addresses an issue where the `storage.download()` method would frequently return stale or cached data instead of the most recent version of a file. By explicitly adding the `Cache-Control: no-store header` to the download request, we ensure that the client always fetches the latest file content directly from the server, bypassing any intermediate caches.

**What changed?** 
I modified the `StorageFileApi.ts` file within the storage-js package. Specifically, I updated the download
method to include the `Cache-Control` header with a value of `no-store` in the options object passed to the underlying 
`get` request function.

This change was necessary to resolve Issue #41841. Users reported that after updating a file in Supabase Storage, immediate subsequent calls to _`_download()_`_ would return the old file content. This behavior suggests that browsers or CDNs were serving a cached response. Adding no-store is a standard HTTP mechanism to instruct all caching layers to avoid storing the response, thereby guaranteeing that the application receives the most up-to-date data.

Breaking changes This Pull Request contains no breaking changes. It only modifies internal header logic for the download request and does not alter the public API signature or return types.

## Summary by CodeRabbit

* **Bug Fixes**
  * File downloads now disable browser caching to ensure fresh content is retrieved on each request.


mandarini:
related

-  https://github.com/supabase/supabase/issues/41841
- https://github.com/supabase/supabase/pull/41855

Thanks for the PR @Akkii88 . I am not sure I will approve, need to discuss/think about it. In the meantime, can you please remove pnpm lock file from git? We use npm!

Thank you again very much for contributing to Supabase! You're helping us make the tool better! 💚

mandarini:
Hi @Akkii88!

Thank you for identifying this issue and proposing a fix. The problem you've identified is real, `download()` doesn't provide a way to control caching, which causes stale data issues in Edge Functions.

However, we can't merge this PR as-is because hardcoding `Cache-Control: no-store` would:

1. **Force all users to bypass caching** - including browsers and Node.js environments where caching improves performance
2. **No opt-out mechanism** - removes user control over caching behavior
3. **Performance impact** - increases server load, bandwidth costs, and latency for all users
4. **Breaking change** - alters default behavior that existing applications may rely on

## Better Approach

I've opened [a new PR](https://github.com/supabase/supabase-js/pull/2090) that takes a different approach: **exposing fetch parameters** rather than hardcoding headers.

This allows Edge Function users to opt-in when needed:

```js
// Edge Functions - bypass cache
const { data } = await supabase.storage
  .from('avatars')
  .download('file.png', {}, { cache: 'no-store' })

// Default behavior (unchanged)
const { data } = await supabase.storage
  .from('avatars')
  .download('file.png')
```

Thank you SO much for contributing to Supabase! Your PR actually helped me identify a problem and come up with a solution. Please do not be discouraged to contribute again in the future. Contributions like yours are what makes Supabase better! 💚

Issue #2023, opened 2026-01-14
return type of `getAuthorizationDetails` is wrong


Kroppeb:
### Describe the bug

I kept getting errors while implementing oauth and couldn't figure out why. `getAuthorizationDetails` is typed as returning `AuthOAuthAuthorizationDetailsResponse`. This however is not correct. It's possible for the `data` to only contain a `redirect_url` when the user has already approved access to that client.

so it's a `AuthOAuthConsentResponse` instead of a `AuthOAuthAuthorizationDetailsResponse`

### Library affected

supabase-js

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-12650H
    Memory: 2.61 GB / 23.73 GB
  Binaries:
    Node: 20.11.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (131.0.2903.112)
    Firefox: 146.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.7309
```

### Used Package Manager

npm

### Logs

I used the following snippet to figure out what was going on:
```js
// Fetch authorization details from Supabase
  const { data: authDetails, error: authError } = await supabase.auth.oauth.getAuthorizationDetails(authorizationId)
  console.log('Auth Details:', authDetails, 'Auth Error:', authError)
```
and got the following log:
```
Auth Details: {
  redirect_url: 'http://localhost:8080/oidc/callback?code=REDACTED&state=REDACTED'
} Auth Error: null
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://supabase.com/docs/reference/javascript/introduction).
- [x] Check that there isn't [already an issue](https://github.com/supabase/supabase-js/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a Supabase JS Library issue and not an issue with the Supabase platform. If it's a Supabase platform related bug, it should likely be reported to [supabase/supabase](https://github.com/supabase/supabase) instead.
[...]

PR #2024, opened 2026-01-15
fix(auth): correct getAuthorizationDetails return type for already consented OAuth responses

7ttp:
## Summary

Make `OAuthAuthorizationDetails` fields optional to support already-consented OAuth responses.

## Problem

`getAuthorizationDetails` is typed to return `AuthOAuthAuthorizationDetailsResponse` with required fields (`authorization_id`, `client`, `user`, `scope`).
 However, when a user has already approved access to an OAuth client, the API returns only `{ redirect_url: string }` without the other fields, causing TypeScript type mismatches.

## Solution

Made `authorization_id`, `client`, `user`, and `scope` optional in `OAuthAuthorizationDetails` type. 
Also fixed a typo in the JSDoc (`redirect_uri` → `redirect_url`).

## Related

- Closes https://github.com/supabase/supabase-js/issues/2023

### DIFF
--- a/packages/core/auth-js/src/lib/types.ts
+++ b/packages/core/auth-js/src/lib/types.ts
@@ -1782,10 +1782,21 @@ export type OAuthAuthorizationDetails = {
 }
 
 /**
+ * OAuth authorization details when user has already consented.
+ * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
+ */
+export type OAuthAlreadyConsentedDetails = {
+  /** URL to redirect the user back to the OAuth client */
+  redirect_url: string
+}
+
+/**
  * Response type for getting OAuth authorization details.
  * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
  */
-export type AuthOAuthAuthorizationDetailsResponse = RequestResult<OAuthAuthorizationDetails>
+export type AuthOAuthAuthorizationDetailsResponse = RequestResult<
+  OAuthAuthorizationDetails | OAuthAlreadyConsentedDetails
+>
 
 /**
  * Response type for OAuth consent decision (approve/deny).
@@ -1834,7 +1845,7 @@ export interface AuthOAuthServerApi {
    * Only relevant when the OAuth 2.1 server is enabled in Supabase Auth.
    *
    * This method returns authorization details including client info, scopes, and user information.
-   * If the response includes a redirect_uri, it means consent was already given - the caller
+   * If the response includes a redirect_url, it means consent was already given - the caller
    * should handle the redirect manually if needed.
    *
    * @param authorizationId - The authorization ID from the authorization request


Kroppeb:
I'd prefer a union type of the 2 possible options.

7ttp:
> I'd prefer a union type of the 2 possible options.

Lets see what @mandarini suggests, will adjust accordingly if needed! 💚🙌

mandarini:
Thank you for identifying and fixing this issue! The type mismatch you've found is indeed a real bug, the current types don't match the actual API behavior. However, I'd like to suggest an alternative approach that preserves type safety while fixing the issue.                                                                                                                                                                         

While making all fields optional does technically solve the TypeScript error, it loses important type information:                                                                              

```ts
  // With optional fields, TypeScript can't help us:                                                                                                                                              
  const { data } = await getAuthorizationDetails(id)                                                                                                                                              
  console.log(data.client.name) // TypeScript says this is fine, but might crash at runtime                                                                                                       
```
                                                                                                                                                                                         
The API actually returns two distinct response shapes depending on the authorization state:                                                                                                     
                                                                                                                                                                                                  
1. First-time authorization (user needs to consent):                                                                                                                                            

```json
  {                                                                                                                                                                                               
    authorization_id: string,                                                                                                                                                                     
    client: OAuthAuthorizationClient,                                                                                                                                                             
    user: { id: string, email: string },                                                                                                                                                          
    scope: string,                                                                                                                                                                                
    redirect_url?: string  // optional                                                                                                                                                            
  }                                                                                                                                                                                               
```

2. Already consented (immediate redirect):                                                                                                                                                      

```json
  {                                                                                                                                                                                               
    redirect_url: string                                                                                                                                                                          
  }                                                                                                                                                                                               
```
                                                                                                                                                                                                  
I think we should use a union type like @Kroppeb suggested to represent both cases accurately. The codebase already has `AuthOAuthConsentResponse` for the already-consented case (line 1794).                 
                                                                                                                                                                                                  
 Here's the proposed change:                                                                                                                                                                     

```ts                                                                                                                                                                                             
  // Keep OAuthAuthorizationDetails as-is (required fields for first-time auth)                                                                                                                   
  export type OAuthAuthorizationDetails = {                                                                                                                                                       
    /** The authorization ID */                                                                                                                                                                   
    authorization_id: string                                                                                                                                                                      
    /** Redirect URL - present if user already consented (can be used to trigger immediate redirect) */                                                                                           
    redirect_url?: string                                                                                                                                                                         
    /** OAuth client requesting authorization */                                                                                                                                                  
    client: OAuthAuthorizationClient                                                                                                                                                              
    /** User object associated with the authorization */                                                                                                                                          
    user: {                                                                                                                                                                                       
      /** User ID (UUID) */                                                                                                                                                                       
      id: string                                                                                                                                                                                  
      /** User email */                                                                                                                                                                           
      email: string                                                                                                                                                                               
    }                                                                                                                                                                                             
    /** Space-separated list of requested scopes */                                                                                                                                               
    scope: string                                                                                                                                                                                 
  }                                                                                                                                                                                               
                                                                                                                                                                                                  
  // Add a new type for the already-consented response                                                                                                                                            
  export type OAuthAlreadyConsentedDetails = {                                                                                                                                                    
    /** URL to redirect the user back to the OAuth client */                                                                                                                                      
    redirect_url: string                                                                                                                                                                          
  }                                                                                                                                                                                               
                                                                                                                                                                                                  
  // Update the response type to be a union                                                                                                                                                       
  export type AuthOAuthAuthorizationDetailsResponse = RequestResult<                                                                                                                              
    | OAuthAuthorizationDetails                                                                                                                                                                   
    | OAuthAlreadyConsentedDetails                                                                                                                                                                
  >                                                                                                                                                                                               
```
                                                                                                                                                                                                  
With this approach, TypeScript forces developers to handle both scenarios:                                                                                                                      

```ts                                                                                                                                                                                                  
  const { data, error } = await supabase.auth.oauth.getAuthorizationDetails(authorizationId)                                                                                                      
                                                                                                                                                                                                  
  if (error || !data) {                                                                                                                                                                           
    // Handle error                                                                                                                                                                               
    return                                                                                                                                                                                        
  }                                                                                                                                                                                               
                                                                                                                                                                                                  
  // Type narrowing using discriminator                                                                                                                                                           
  if ('authorization_id' in data) {                                                                                                                                                               
    // TypeScript knows all fields are present here                                                                                                                                               
    console.log('Client:', data.client.name)                                                                                                                                                      
    console.log('Scopes:', data.scope)                                                                                                                                                            
    console.log('User:', data.user.email)                                                                                                                                                         
                                                                                                                                                                                                  
    // Show consent UI...                                                                                                                                                                         
  } else {                                                                                                                                                                                        
    // TypeScript knows only redirect_url is present                                                                                                                                              
    console.log('Already consented, redirecting to:', data.redirect_url)                                                                                                                          
                                                                                                                                                                                                  
    // Handle redirect...                                                                                                                                                                         
  }                                                                                                                                                                                               
```
                                                                                                                                                                                         
The current types claim all fields are required, but the API sometimes returns only `redirect_url`. So the existing types are incorrect and misleading. Fixing them to match reality is a bug fix, not a breaking change. Existing code that incorrectly assumes all fields are always present already has runtime bugs that just haven't surfaced yet. The union type approach will surface these bugs at compile time, which is exactly what TypeScript is for.                                                                                                                                                        

P.S. The JSDoc typo fix (redirect_uri → redirect_url) is perfect - that should definitely stay regardless of which approach we take!

7ttp:
@mandarini 
can u have another look on this?
also please lmk if any more changes are required!  😁💚

mandarini:
Hey @7ttp! Unfortunately, I think I'm going to close this PR in favor of a more comprehensive fix. The problem is on us. 

The issue with this PR: While you added the union type for the already-consented case, you missed the underlying bug in OAuthAuthorizationDetails itself. The API actually returns redirect_uri (not redirect_url) when consent is needed. This was incorrectly changed in PR #1891.                         
                                                                                                           
What we're doing instead:                                                                                
  - Fix the core bug: redirect_url → redirect_uri in OAuthAuthorizationDetails (matches actual API)        
  - Add union type (as you did), but with a more generic OAuthRedirect type                                
  - Consolidate duplicate types (reuse OAuthRedirect in consent response)                                  
  - Fix all JSDoc to clarify which field names appear in which cases                                       
                                                                                                           
Your PR would have fixed the already-consented case but left the consent-needed case still broken with the wrong field name. Thanks again for reporting this! 💚

mandarini:
Fixed here: https://github.com/supabase/supabase-js/pull/2088

PR #2028, opened 2026-01-17
Fix/postgrest not null narrowing

### LINKED ISSUE #1360: Inferred types for nullable columns after not null checks should be non-nullable
# Bug report

## Describe the bug

For nullable columns, even if the query includes a not null check (e.g. `.not("column_name", "is", null)`), the inferred return type for that column is still nullable.

## To Reproduce

1.  Create a `blog_posts` table with a nullable column, `published_at`.

```sql
create table blog_posts(
  id           uuid        primary key default uuid_generate_v4(),
  title        text        not null,
  created_at   timestamptz not null default now(),
  published_at timestamptz
);
```

2. Query this table to retrieve all published blog posts.

```ts
  .from("blog_posts")
  .select("*")
  .not("published_at", "is", null);
```

3. Observe that the `published_at` field in `data` is still typed as nullable.

```ts
const data: {
  id: string;
  title: string;
  created_at: string;
  published_at: string | null; // <-- why is this nullable?
}[] | null
```

## Expected behavior

Inferred type for `published_at` should not be nullable.

## System information

- Version of supabase-js: 2.48.0

DevanshTyagi04:
Improves TypeScript type inference for PostgREST filters by correctly narrowing column types after explicit non-null checks.

### What changed?

- Added a specialized overload for `not(column, 'is', null)` in `PostgrestFilterBuilder`
- Ensured that when a column is filtered to exclude `NULL`, the resulting row type marks that column as **non-nullable**
- Preserved existing behavior for all other `not()` filter usages

Previously, filtering out `NULL` values using `not(column, 'is', null)` did not update the inferred row type, leaving the column typed as nullable.  
This caused a mismatch between PostgREST semantics and TypeScript inference.

This change aligns the type system with runtime behavior and improves developer ergonomics.

Closes #1360

## 📸 Screenshots/Examples

```ts
// Before
  .from('users')
  .select()
  .not('email', 'is', null)

// email: string | null ❌

// After
// email: string ✅

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -76,6 +76,10 @@ type ResolveFilterRelationshipValue<
 
 export type InvalidMethodError<S extends string> = { Error: S }
 
+type NonNullableColumn<Row, Column extends keyof Row> = Omit<Row, Column> & {
+  [K in Column]-?: NonNullable<Row[K]>
+}
+
 export default class PostgrestFilterBuilder<
   ClientOptions extends ClientServerOptions,
   Schema extends GenericSchema,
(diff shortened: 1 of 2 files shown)


mandarini:
I am closing this PR in favor of https://github.com/supabase/supabase-js/pull/2264.

I think there are a number of issues and missing pieces in the current solution, but thank you VERY much regardless, @DevanshTyagi04, because you motivated me to address that long standing issue.

Do not get discouraged by this, and thank you very much for contributing to Supabase!

PR #2029, opened 2026-01-17
fix(realtime): wait for postgres_changes system ready before emitting SUBSCRIBED

### LINKED ISSUE #1599: Premature 'SUBSCRIBED' status for postgres_changes signals incorrect readiness
# Bug report

## Describe the bug

There is a subtle but critical race condition in the Realtime subscription process for `postgres_changes`. The client libraries (e.g., `supabase-js`) emit a `'SUBSCRIBED'` status prematurely, before the backend's PostgreSQL logical replication listener is fully initialized and ready to stream changes.

This leads to a state where the client believes it is successfully subscribed, but any database writes performed within a brief window (approx. 1-3 seconds) immediately following this confirmation will not trigger a `postgres_changes` event on the client. The event is silently missed.

This issue primarily affects time-sensitive applications that need to perform a database write as a direct result of a successful subscription confirmation.

## To Reproduce

The behavior is reproducible in any client-side environment.

1.  Set up a basic client application and a database table (e.g., `messages`).
2.  In the client code, subscribe to `postgres_changes` for the `messages` table.
3.  In the `.subscribe()` callback, listen for the `status === 'SUBSCRIBED'` event. When this event is received, immediately enable a button that inserts a new row into the `messages` table.
4.  In the same component, set up an `.on('postgres_changes', ...)` listener that logs the payload to the console.
5.  Load the application and, **the instant the button becomes enabled**, click it.
6.  **Observe the error:** The database `insert` will succeed, but the `.on('postgres_changes', ...)` listener will not fire. No payload will be logged to the console. If you wait a few seconds after the button is enabled before clicking, the event is received correctly.

## Expected behavior

The `'SUBSCRIBED'` status should be a reliable signal that the end-to-end connection is established and the client is ready to receive all subsequent `postgres_changes` events.

When a client performs a database write immediately after receiving the `'SUBSCRIBED'` status, the corresponding `postgres_changes` event should be reliably delivered to the client's listener. There should be no missed events due to an internal initialization delay.

## System information

- OS: All
- Browser (if applies): All
- Version of supabase-js: Affects all `v2.x.x` versions.
- Version of Node.js: N/A (client-side issue)

This is a platform-level issue and likely affects all client libraries that use the Realtime service in the same way.

## Additional context

This issue stems from a discrepancy between the public API contract and the internal workings of the Realtime server.

**The Interplay Between Realtime Core and Client Libraries:**

1.  **Two-Stage Backend Process:** An analysis of the Realtime Elixir source code reveals a complex, multi-stage, asynchronous process for establishing a `postgres_changes` subscription.
    *   **Stage 1 (Fast):** The client joins a Phoenix WebSocket channel. This is a very quick operation.
    *   **Stage 2 (Slow):** The Realtime server then initiates a separate, slower workflow managed by `Realtime.Tenants.Connect` and `Realtime.Tenants.ReplicationConnection`. This involves connecting to the project's database, checking for/creating a publication, creating a temporary replication slot, and finally starting the logical replication stream. This process can take several seconds.

2.  **Misleading API Contract:**
    *   The **Realtime backend** does not emit a single, high-level event that signals the completion of Stage 2. The final "ready" state can only be inferred by observing a second, low-level `system` message in the WebSocket traffic.
    *   The **client libraries** (like `supabase-js`) are designed to listen for the confirmation of Stage 1 (the Phoenix channel join). They correctly interpret the `phx_reply` as a successful subscription and translate this into the public `'SUBSCRIBED'` status for the developer.

This abstraction, while simple, is misleading. The client library is functioning as designed, but the event it is designed to listen for does not accurately represent the readiness of the `postgres_changes` stream. The bug is in the API contract between the core service and its clients.

A robust solution would involve either the backend sending a definitive "replication ready" event, or the client libraries becoming smarter by waiting for the internal `system` message before emitting the public `'SUBSCRIBED'` status.

7ttp:
## Summary

Fixes a race condition where `SUBSCRIBED` status is emitted before PostgreSQL logical replication is ready,
 causing early database writes to miss `postgres_changes` events.

## Problem

The Realtime subscription has a two-stage backend process:
1. **Stage 1 (Fast)**: Phoenix WebSocket channel join completes
2. **Stage 2 (Slow)**: PostgreSQL replication slot creation and stream initialization

Previously, the client emitted `SUBSCRIBED` after Stage 1, but the replication stream wasn't ready until Stage 2 completed. Any database writes in this window (~1-3 seconds) would not trigger `postgres_changes` events.

## Solution

Wait for the server's system message `{ extension: 'postgres_changes', status: 'ok' }` before emitting `SUBSCRIBED` when postgres_changes bindings are present. This message is sent by the server after `Realtime.Tenants.ReplicationConnection` confirms the replication stream is active.

For channels without postgres_changes (broadcast/presence only), behavior is unchanged.

## Changes

- Store subscribe callback and track pending system confirmations
- Add `_maybeEmitSubscribed()` that only fires when all confirmations received
- Handle system message in `_trigger()` to clear pending state
- Updated existing lifecycle tests to include system message trigger.

---
## Related

- Closes: https://github.com/supabase/supabase-js/issues/1599

### DIFF
--- a/packages/core/realtime-js/test/RealtimeChannel.lifecycle.test.ts
+++ b/packages/core/realtime-js/test/RealtimeChannel.lifecycle.test.ts
@@ -349,6 +349,8 @@ describe('Channel Lifecycle Management', () => {
           },
         })
 
+        channel._trigger('system', { extension: 'postgres_changes', status: 'ok' })
+
         expect(spy).toHaveBeenCalledTimes(1)
         expect(spy).toHaveBeenCalledWith({
           topic: 'realtime:topic',
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -355,8 +355,12 @@ export default class RealtimeChannel {
 
             this.bindings.postgres_changes = newPostgresBindings
 
-            this._subscribeCallback = callback || null
-            this._pendingSystemReady.add('postgres_changes')
+            if (bindingsLen > 0) {
+              this._subscribeCallback = callback || null
+              this._pendingSystemReady.add('postgres_changes')
+            } else {
+              callback?.(REALTIME_SUBSCRIBE_STATES.SUBSCRIBED)
+            }
             return
           }
         })
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -180,6 +180,9 @@ export default class RealtimeChannel {
   broadcastEndpointURL: string
   subTopic: string
   private: boolean
+  private _subscribeCallback: ((status: REALTIME_SUBSCRIBE_STATES, err?: Error) => void) | null =
+    null
+  private _pendingSystemReady: Set<string> = new Set()
 
   /**
    * Creates a channel that can broadcast messages, sync presence, and listen to Postgres changes.
@@ -352,7 +355,8 @@ export default class RealtimeChannel {
 
             this.bindings.postgres_changes = newPostgresBindings
 
-            callback && callback(REALTIME_SUBSCRIBE_STATES.SUBSCRIBED)
+            this._subscribeCallback = callback || null
+            this._pendingSystemReady.add('postgres_changes')
             return
           }
         })
@@ -820,6 +824,14 @@ export default class RealtimeChannel {
   }
 
   /** @internal */
+  _maybeEmitSubscribed(): void {
+    if (this._pendingSystemReady.size === 0 && this._subscribeCallback) {
+      this._subscribeCallback(REALTIME_SUBSCRIBE_STATES.SUBSCRIBED)
+      this._subscribeCallback = null
+    }
+  }
+
+  /** @internal */
   _trigger(type: string, payload?: any, ref?: string) {
     const typeLower = type.toLocaleLowerCase()
     const { close, error, leave, join } = CHANNEL_EVENTS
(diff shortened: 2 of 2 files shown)


mandarini:
@7ttp I know you and @edgurgel discussed this internally, so I am putting a "do not merge" label

mandarini:
@7ttp @edgurgel  what is the status here? Should we keep open / should we close?

edgurgel:
Hey @mandarini, funnily enough we were discussing this issue last week and I think we found a good way forward here that is good enough:

We think that we could change so that IF a channel has ONLY postgres changes (so no broadcast and presence `on` callbacks) we could then only consider SUBSCRIBED then the system ready message arrives. 

Thoughts? @7ttp

7ttp:
> Hey @mandarini, funnily enough we were discussing this issue last week and I think we found a good way forward here that is good enough:
> 
> We think that we could change so that IF a channel has ONLY postgres changes (so no broadcast and presence `on` callbacks) we could then only consider SUBSCRIBED then the system ready message arrives.
> 
> Thoughts? @7ttp

looks good!!
should I put up a pr for that? :)

edgurgel:
> looks good!! should I put up a pr for that? :)

Yes! That would be great ❤️

mandarini:
@7ttp should we close this PR, and you will open new one?

packages/core/realtime-js/test/RealtimeChannel.lifecycle.test.ts
```diff
@@ -349,6 +349,8 @@ describe('Channel Lifecycle Management', () => {
           },
         })
 
+        channel._trigger('system', { extension: 'postgres_changes', status: 'ok' })
+
```
7ttp, line 353: moved to: 
- https://github.com/supabase/supabase-js/pull/2438

Issue #2037, opened 2026-01-20
Breaking behavior change: `exchangeCodeForSession` defers `SIGNED_IN` event in v2.91.0, impacting OAuth cookie writes in SSR/serverless


dalkommatt:
## Summary
After upgrading `@supabase/supabase-js` from `2.90.1` to `2.91.0`, OAuth login stopped working (both localhost and Vercel). OTP auth continues to work.

Both OAuth providers tested (Google + Discord) are affected.

The regression appears to be caused by the v2.91.0 change **“auth: defer subscriber notification in exchangeCodeForSession to prevent deadlock”** (PR #2014).

In Next.js Route Handlers / serverless runtimes using `@supabase/ssr`, OAuth callbacks commonly rely on the `SIGNED_IN` event emitted during `auth.exchangeCodeForSession()` to write auth cookies via the SSR cookie adapter. In v2.91.0, `SIGNED_IN` is now notified via `setTimeout(..., 0)`, which may not run before the request completes, resulting in **no auth cookies being set**.

This may be an intentional tradeoff to avoid a deadlock in some environments, but it is a breaking behavior change for serverless/SSR callback handlers that expect `exchangeCodeForSession` to complete cookie persistence before the response returns.

## Affected flow
- `auth.signInWithOAuth()` redirects the user to the provider
- Provider redirects back to `/auth/callback?code=...`
- Server route calls `auth.exchangeCodeForSession(code)`
- **Expected:** auth cookies are set and user is signed in
- **Actual (v2.91.0):** callback redirects, but cookies are not set; user remains logged out

In apps that protect pages via Next.js middleware (e.g. calling `supabase.auth.getClaims()` / `getUser()` on each request), the next navigation to a protected page (like `/dashboard`) immediately redirects back to `/login` because no session is present.

## Regression
- ✅ Works: `@supabase/supabase-js@2.90.1`
- ❌ Broken: `@supabase/supabase-js@2.91.0`

### Observed result
- Callback handler executes and redirects.
- `exchangeCodeForSession` does **not** return an error.
- No Supabase auth cookies are persisted (e.g. `sb-...` cookies).
- Subsequent server/client calls return no session; protected routes redirect to `/login`.

### Expected result
- `exchangeCodeForSession` causes the SSR client to persist auth cookies during the callback request.

## Suspected cause
v2.91.0 includes:
- Release note: https://github.com/supabase/supabase-js/releases/tag/v2.91.0
[...]

PR #2016, opened 2026-01-12
fix(auth): make _notifyAllSubscribers non-blocking to prevent callback deadlocks

7ttp:
## Summary

Makes `_notifyAllSubscribers` non-blocking by removing `await` on subscriber callbacks. This prevents deadlocks when callbacks perform async operations that require the auth lock.

## Problem

When a subscriber callback in `onAuthStateChange` performs async operations like `getUser()` or `getSession()`, a deadlock occurs:

1. Auth method (e.g., `signInWithPassword`) acquires lock via `_acquireLock()`
2. Inside the lock, it calls `await _notifyAllSubscribers()` which awaits all callbacks
3. Subscriber callback calls `getUser()` which tries to acquire the same lock
4. Lock queues `getUser()` to wait for current operation
5. Current operation waits for callback to complete
6. **Deadlock**: callback waits for `getUser()` → waits for lock → waits for callback

## Solution

Remove `async/await` from `_notifyAllSubscribers`. Callbacks fire synchronously, async callbacks run in background without blocking. Errors still caught and logged.

## Related

- Extends #2014 - that PR added `setTimeout` wrapper at one call site (`exchangeCodeForSession`). This fix makes `_notifyAllSubscribers` itself non-blocking, fixing all 20+ call sites.
- closes https://github.com/supabase/supabase-js/issues/2013
- Related https://github.com/supabase/supabase/issues/41968

Issue #2013, opened 2026-01-10
Supabase Auth-JS Deadlock Issue


datrine:
### Describe the bug

# Supabase Auth-JS Deadlock Issue

## Summary
`supabase.auth.signInWithPassword()` and `supabase.auth.setSession()` hang indefinitely after receiving a successful 200 OK response from the Supabase auth API. The deadlock occurs in the `_acquireLock` mechanism during session storage operations.

## Environment
- **Package**: `@supabase/auth-js`
- **Version**: Latest (as of January 2026)
- **Browser**: Chrome/Edge (Chromium-based)
- **Platform**: Windows
- **Framework**: React + Vite

### Observed Behavior
1. Call `supabase.auth.signInWithPassword({ email, password })`
2. Network tab shows successful 200 OK response from `/auth/v1/token?grant_type=password`
3. Response body contains valid session data (access_token, refresh_token, user)
4. **Function never resolves or rejects** - hangs indefinitely
5. Browser tab becomes unresponsive
6. Same issue occurs with `supabase.auth.setSession()`

### Expected Behavior
- Function should resolve with session data after successful authentication
- Session should be saved to storage
- User should be signed in

### Primary Issue: Deadlock in `_acquireLock` Mechanism

**File**: `node_modules/@supabase/auth-js/src/GoTrueClient.ts`

**Line 1485-1549**: `_acquireLock` implementation
```typescript
private async _acquireLock<R>(acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
  // ...
  return await this.lock(`lock:${this.storageKey}`, acquireTimeout, async () => {
    this.lockAcquired = true
    const result = fn()
    // ...
    await result
    // ...
  })
}
```

**Line 672-713**: `signInWithPassword` calls `_acquireLock`
```typescript
async signInWithPassword(credentials: SignInWithPasswordCredentials): Promise<AuthResponse> {
  // ...
  return this._acquireLock(-1, async () => {
    return await this._signInWithPassword(credentials)
  })
}
```

**Line 1854-1856**: `setSession` also calls `_acquireLock`
```typescript
async setSession(currentSession: { access_token: string; refresh_token: string }): Promise<AuthResponse> {
  await this.initializePromise
  return await this._acquireLock(-1, async () => {
    return await this._setSession(currentSession)
  })
}
```

**Line 2769-2816**: `_saveSession` attempts storage operations while lock is held
```typescript
protected async _saveSession(session: Session) {
  // ...
[...]

PR #2039, opened 2026-01-20
fix(auth): restore SSR OAuth functionality broken in v2.91.0

7ttp:
# Summary

Reverts #2014 which broke OAuth authentication in SSR/serverless environments.

# Problem

PR #2014 added `setTimeout(..., 0)` to defer `SIGNED_IN` event notifications in `exchangeCodeForSession`. While this prevented a deadlock in some scenarios, it introduced a critical regression:

**In serverless/SSR environments (Next.js, Vercel, etc.):**
- OAuth callbacks rely on `exchangeCodeForSession` to trigger auth state change listeners
- These listeners (via `@supabase/ssr`) write auth cookies during the request
- `setTimeout(..., 0)` defers execution to the next event loop tick
- The serverless function completes and returns the response **before** the deferred callback runs
- Result: **No auth cookies are set, OAuth login fails**

**User Impact:**
- OAuth login (Google, etc.) stopped working after upgrading from v2.90.1 to v2.91.0
- Users are redirected but remain logged out
- OTP auth continues to work (unaffected)

# Solution

The deadlock issue that #2014 attempted to fix is properly addressed by #2016, which makes `_notifyAllSubscribers` non-blocking by removing `await` on subscriber callbacks. This approach:

✅ Prevents deadlocks when callbacks perform async operations like `getUser()` or `getSession()`  
✅ Keeps notifications synchronous so SSR cookie adapters work correctly  
✅ Fixes the root cause instead of working around it with `setTimeout`

# Related

- Closes https://github.com/supabase/supabase-js/issues/2037
- Supersedes https://github.com/supabase/supabase-js/pull/2038
- Superseded by #2016 (proper deadlock fix)
- Original deadlock issue: #2013

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -1146,9 +1146,7 @@ export default class GoTrueClient {
       }
       if (data.session) {
         await this._saveSession(data.session)
-        setTimeout(async () => {
-          await this._notifyAllSubscribers('SIGNED_IN', data.session)
-        }, 0)
+        await this._notifyAllSubscribers('SIGNED_IN', data.session)
       }
       return this._returnResult({ data: { ...data, redirectType: redirectType ?? null }, error })
     } catch (error) {


7ttp:
@mandarini
 can you please have a look at #2016  before reviewing this! 💚

PR #2043, opened 2026-01-21
fix(realtime): fix tests to show bugs with resubscribe

GuzekAlan:
## 🔍 Description

Fixes tests to show that there are bugs with resubscribe

## 📸 Screenshots/Examples

image


mandarini:
Hi @GuzekAlan , thanks for investigating this! Typically, PRs that only make tests fail aren't mergeable on their own since they'd break CI. Would you be able to either:

- Include the actual fix for the resubscribe bug in this PR, or
- Open a GitHub issue describing the bug with these test changes as evidence?

That way the bug is documented and we have a path to resolution.

Issue #1654, opened 2024-01-20
Remove signature to allow any table name for from


olee:
I think the 3rd signature for the `from` method should be removed which currently allows using any table name inside the postgrest client:
https://github.com/supabase/postgrest-js/blob/f730aa793fbc8650de8bf914c816e943eba42844/src/PostgrestClient.ts#L59-L66
Currently this makes it quite hard to catch errors when for example removing or renaming tables.

------
If anyone would like to allow calling from on unknown tables, they could still do it imho by adding an index signature to the schema eg.
```ts
type Database = {
    public: {
        Tables: {
            /* standard signatures go here */

            // Allow any table name in from etc
            [key: string]: any;
        }
    }
}
```

PR #2058, opened 2026-01-26
fix(postgrest): enforce type safety for table and view names in from() method

7ttp:
## Problem
The `from()` method's implementation signature accepted any string, 
defeating TypeScript's type safety and allowing typos or invalid table/view names to pass compile time checks.

## Solution
Constrained the implementation signature parameter type to
 `(string & keyof Schema['Tables']) | (string & keyof Schema['Views'])`, 
ensuring only valid schema-defined table and view names are accepted

## Related
- Closes https://github.com/supabase/supabase-js/issues/1654

## Summary by CodeRabbit

* **Breaking Changes**
  * Type-level validation for query targets is stricter: references must match valid tables or views from your schema.

* **Improvements**
  * Invalid table/view references are detected at development time rather than runtime, improving reliability and developer experience.

### DIFF
--- a/packages/core/postgrest-js/test/index.test-d.ts
+++ b/packages/core/postgrest-js/test/index.test-d.ts
@@ -9,12 +9,14 @@ const REST_URL = 'http://localhost:54321/rest/v1'
 const postgrest = new PostgrestClient<Database>(REST_URL)
 const postgrestWithOptions = new PostgrestClient<DatabaseWithOptions>(REST_URL)
 
-// table invalid type
+// table and view name type safety
 {
   // @ts-expect-error Argument of type '42' is not assignable to parameter of type
   postgrest.from(42)
   // @ts-expect-error Argument of type '"nonexistent_table"' is not assignable to parameter of type
   postgrest.from('nonexistent_table')
+  // @ts-expect-error Argument of type '"nonexistent_view"' is not assignable to parameter of type
+  postgrest.from('nonexistent_view')
 }
 
 // `null` can't be used with `.eq()`


packages/core/postgrest-js/src/PostgrestClient.ts
```diff
@@ -87,7 +87,9 @@ export default class PostgrestClient<
    *
    * @param relation - The table or view name to query
    */
-  from(relation: string): PostgrestQueryBuilder<ClientOptions, Schema, any, any> {
+  from(
+    relation: (string & keyof Schema['Tables']) | (string & keyof Schema['Views'])
+  ): PostgrestQueryBuilder<ClientOptions, Schema, any, any> {
```
steve-chavez, line 92: Code LGTM but could we add tests for this?

cc @avallete

7ttp:
The type tests were already there,
 just added a line for invalid view names to match the existing table validation 😁💚

packages/core/postgrest-js/test/index.test-d.ts
```diff
@@ -9,12 +9,14 @@ const REST_URL = 'http://localhost:54321/rest/v1'
 const postgrest = new PostgrestClient<Database>(REST_URL)
 const postgrestWithOptions = new PostgrestClient<DatabaseWithOptions>(REST_URL)
 
-// table invalid type
+// table and view name type safety
 {
   // @ts-expect-error Argument of type '42' is not assignable to parameter of type
   postgrest.from(42)
   // @ts-expect-error Argument of type '"nonexistent_table"' is not assignable to parameter of type
   postgrest.from('nonexistent_table')
```
avallete, line 17: **question**

If that was just a generic string, shouldn't this test had fail (because no error being raised) ?

7ttp: > **question**
> 
> If that was just a generic string, shouldn't this test had fail (because no error being raised) ?

Umm the tests for `42` and `nonexistent_table` were already there, they work because the constrained signature rejects those invalid inputs. 
I just added `nonexistent_view` to make sure views are covered too.

7ttp:
> Flagging here that I wonder if that will break for clients that don't use Database introspected types. It might require a BC release.

For users with generated Database types, this adds compile time safety.
 For users without types, Database defaults to `any`, and since keyof any resolves to `string | number | symbol`, 
any string still works fine. 
So it's backward compatible ig for untyped users while adding safety for typed users 

thanks for the review @steve-chavez & @avallete 😁💚

7ttp:
@mandarini

PR #2072, opened 2026-01-29
feat(postgrest): add automatic retries for transient errors

grdsdev:
### What changed?

Adds automatic retry logic to `@supabase/postgrest-js` for two classes of transient errors:

- **520 errors** (Cloudflare timeout/connection errors under load) — ephemeral and safe to retry
- **503 errors** (PostgREST `PGRST002` schema cache loading on a replica) — PostgREST explicitly signals retryability; when a `Retry-After` header is present, its value is used as the delay

**Implementation details:**

- Retries are enabled by default for GET/HEAD/OPTIONS requests (idempotent methods only)
- Up to 3 retry attempts with exponential backoff (1s, 2s, 4s, capped at 30s)
- Network errors are also retried for idempotent methods
- Response body is drained before each retry to prevent stream leaks
- `AbortError` is rethrown immediately without retrying
- Backoff sleeps respect the `AbortSignal` — aborting mid-wait resolves immediately
- `X-Retry-Count` request header is added on retry attempts (requested by PostgREST author for observability)
- Simple boolean API to control retries globally or per-request

**New `retry` option on `PostgrestClient`:**

```typescript
// Retries enabled by default - no configuration needed
const { data, error } = await supabase.from('users').select()

// Disable retries globally
const client = new PostgrestClient('https://api.example.com', {
  retry: false,
})

// Disable retries per-request
  .from('users')
  .select()
  .retry(false)

// Re-enable per-request when globally disabled
  .from('users')
  .select()
  .retry(true)
```

**Key design decisions:**

| Decision | Rationale |
|----------|-----------|
| Enabled by default | Gives perception of more reliable service without opt-in friction |
| Only GET/HEAD/OPTIONS | Non-idempotent methods (POST/PATCH/DELETE) could cause data duplication |
| 520 + 503 | 520 = Cloudflare transient; 503 = PostgREST schema cache loading |
| Respects `Retry-After` header | PostgREST sends `Retry-After: 0` for PGRST002 — no unnecessary delay |
| `X-Retry-Count` header | Requested by PostgREST author for server-side observability of retry storms |
| AbortError not retried | User explicitly cancelled; retrying would violate intent |
| Abort respected during backoff | Aborting mid-sleep resolves immediately instead of waiting the full delay |
| Body drained before retry | Avoids stream leaks on retried responses |

Customers using load-balanced Supabase URLs and Cloudflare-fronted deployments experience intermittent failures that are safe to retry automatically. Without retries, these transient errors surface as real failures, making PostgREST appear unreliable even though the underlying request would succeed on a second attempt.

## Screenshots/Examples

N/A

## Breaking changes

## Checklist

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -18,9 +18,24 @@ import { ContainsNull } from './select-query-parser/types'
 
 /**
  * Sleep for a given number of milliseconds.
+ * If an AbortSignal is provided, the sleep resolves early when the signal is aborted.
  */
-function sleep(ms: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms))
+function sleep(ms: number, signal?: AbortSignal): Promise<void> {
+  return new Promise((resolve) => {
+    if (signal?.aborted) {
+      resolve()
+      return
+    }
+    const id = setTimeout(() => {
+      signal?.removeEventListener('abort', onAbort)
+      resolve()
+    }, ms)
+    function onAbort() {
+      clearTimeout(id)
+      resolve()
+    }
+    signal?.addEventListener('abort', onAbort)
+  })
 }
 
 /**
@@ -159,9 +174,9 @@ export default abstract class PostgrestBuilder<
    *
    * Configure retry behavior for this request.
    *
-   * By default, retries are enabled for GET/HEAD requests that fail with
-   * 520 status codes (Cloudflare timeout/connection errors). Retries use
-   * exponential backoff (1s, 2s, 4s) with a maximum of 3 attempts.
+   * By default, retries are enabled for idempotent requests (GET, HEAD, OPTIONS)
+   * that fail with network errors or specific HTTP status codes (503, 520).
+   * Retries use exponential backoff (1s, 2s, 4s) with a maximum of 3 attempts.
    *
    * @param enabled - Whether to enable retries for this request
    *
@@ -251,7 +266,7 @@ export default abstract class PostgrestBuilder<
           if (this.retryEnabled && attemptCount < DEFAULT_MAX_RETRIES) {
             const delay = getRetryDelay(attemptCount)
             attemptCount++
-            await sleep(delay)
+            await sleep(delay, this.signal)
             continue
           }
 
@@ -268,7 +283,7 @@ export default abstract class PostgrestBuilder<
               : getRetryDelay(attemptCount)
           await res.text()
           attemptCount++
-          await sleep(delay)
+          await sleep(delay, this.signal)
           continue
         }
 
--- a/packages/core/postgrest-js/src/PostgrestClient.ts
+++ b/packages/core/postgrest-js/src/PostgrestClient.ts
@@ -55,9 +55,9 @@ export default class PostgrestClient<
    * @param options.timeout - Optional timeout in milliseconds for all requests. When set, requests will automatically abort after this duration to prevent indefinite hangs.
    * @param options.urlLengthLimit - Maximum URL length in characters before warnings/errors are triggered. Defaults to 8000.
    * @param options.retry - Enable or disable automatic retries for transient errors.
-   *   When enabled, GET/HEAD requests that fail with 520 errors (Cloudflare timeouts)
-   *   will be automatically retried up to 3 times with exponential backoff.
-   *   Defaults to `true`.
+   *   When enabled, idempotent requests (GET, HEAD, OPTIONS) that fail with network
+   *   errors or HTTP 503/520 responses will be automatically retried up to 3 times
+   *   with exponential backoff (1s, 2s, 4s). Defaults to `true`.
    * @example
    * ```ts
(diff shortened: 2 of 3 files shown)


packages/core/postgrest-js/src/PostgrestBuilder.ts
```diff
@@ -119,108 +189,64 @@ export default abstract class PostgrestBuilder<
...
+            attemptCount++
+            await sleep(delay)
+            continue
+          }
 
-        if (error && this.shouldThrowOnError) {
-          throw new PostgrestError(error)
+          // Exhausted retries or retries disabled, throw the last error
+          throw fetchError
         }
```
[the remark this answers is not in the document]
mandarini: I think this is outdated

steve-chavez:
> Note: The X-Retry-Count diagnostic header was intentionally omitted from this PR to avoid CORS preflight issues with static Access-Control-Allow-Headers configs. It will be introduced in a future PR alongside an update to cors.ts.

The lack of retry header is concerning, without a retry header it will be too hard to debug at the origin server. Can that be added in this PR?

mandarini:
@steve-chavez I was worried about breaking changes. Since we now have https://supabase.com/docs/guides/functions/cors we can rely on this and hope users will be using it. I will add back the headers. Any thoughts about the 5xx errors?

packages/core/supabase-js/src/lib/rest/types/common/common.ts
grdsdev, line 1: This is duplicate of `packages/core/postgrest-js/src/types/common/common.ts`? Why?

packages/core/postgrest-js/src/PostgrestBuilder.ts
```diff
@@ -141,101 +210,73 @@ export default abstract class PostgrestBuilder<
...
+          })
+
+          // Check if we should retry this response
+          if (shouldRetry(this.method, res.status, attemptCount, this.retryEnabled)) {
+            const retryAfterHeader = res.headers.get('Retry-After')
+            const delay =
+              retryAfterHeader !== null
+                ? Math.max(0, parseInt(retryAfterHeader, 10) || 0) * 1000
+                : getRetryDelay(attemptCount)
+            await res.text()
```
o-santi: Why? If we are retrying, why try reading the response and not using the result afterwards?

packages/core/postgrest-js/src/PostgrestBuilder.ts
```diff
@@ -141,101 +210,73 @@ export default abstract class PostgrestBuilder<
...
 
-        if (error && this.shouldThrowOnError) {
-          throw new PostgrestError(error)
+          // Check if we should retry network errors
+          if (this.retryEnabled && attemptCount < DEFAULT_MAX_RETRIES) {
+            const delay = getRetryDelay(attemptCount)
+            attemptCount++
+            await sleep(delay)
+            continue
+          }
```
o-santi, line 271: When would this happen? Does `fetch` throw on 520? What is this part covering inside the `catch`? It seems to me that this is bigger reaching that intended.

o-santi: It is also not checking for `shouldRetry`. Is this inlining the `shouldRetry` function body inside the `catch`?  If so, they are not doing the same thing.

packages/core/supabase-js/src/lib/rest/types/common/common.ts
mandarini: The `Generic*` types are needed by both `postgrest-js` (internally, for type inference) and `supabase-js` (publicly, as part of its API surface). Rather than making them part of postgrest-js's public API, [this PR](https://github.com/supabase/supabase-js/pull/1632) chose to keep them as a shared definition that lives in both packages, with `postgrest-js` as the source of truth and a codegen script (which [I added later](https://github.com/supabase/supabase-js/pull/1778)) keeping `supabase-js` in sync. This is a deliberate ownership boundary: postgrest-js owns the type logic, supabase-js owns the public API.

packages/core/postgrest-js/src/PostgrestBuilder.ts
mandarini: We need to drain the response body before retrying to avoid connection/stream leaks. on HTTP/1.1 the connection can't be reused until the body is fully consumed. The result is intentionally discarded since it's just the error body of the transient failure we're retrying past.

packages/core/postgrest-js/src/PostgrestBuilder.ts
mandarini: re: fetch on 520:  Correct! fetch doesn't throw on 520, it resolves with a response. The catch was only meant to cover true network errors (connection resets, DNS failures), but because `processResponse` was also inside the try, it was accidentally catching `PostgrestError` throws too. I fixed it: the `try/catch` now wraps only the `_fetch()` call, so `processResponse` errors can never trigger a retry.

mandarini: re: shouldRetry: Correct again 😅 , they weren't equivalent. I restructured it and the two paths are now separate: shouldRetry handles HTTP response retries (outside the try), and the catch handles only thrown fetch errors.

PR #2078, opened 2026-01-29
feat(postgrest): add URL length validation and timeout protection

mandarini:
## Problem
When users create queries with very long field lists (10k+ characters), requests can hang indefinitely instead of returning an error. This occurs because:

1. PostgREST echoes query parameters back in `Content-Location` response headers
2. Large response headers exceed Node's undici parser limits (default 8KB)
3. Errors during response header parsing may not propagate correctly, causing hangs
4. No timeout protection exists by default

## Solution
This PR implements three defensive changes to prevent indefinite hangs and improve developer experience:

### 1. URL Length Warning (PostgrestQueryBuilder.ts)
- Console warning when `select` parameter exceeds 8000 characters
- Helps developers identify potential issues before they occur
- Non-breaking: warning only, doesn't change behavior

```typescript
// When select string > 8KB, logs:
// "postgrest-js: select parameter is 9889 characters. Very long field lists
//  may cause issues due to HTTP header limits. Consider using views or
//  selecting fewer fields."
```

### 2. Timeout Option (PostgrestClient.ts)
- New optional `timeout` parameter in constructor
- Automatically aborts requests after specified duration
- Properly merges with existing AbortSignals
- Non-breaking: opt-in via constructor option

```typescript
const client = new PostgrestClient('https://example.com/rest/v1', {
  timeout: 30000 // 30 second timeout
})
```

### 3. Enhanced Error Messages (PostgrestBuilder.ts)
- AbortError responses now include helpful hints
- Links timeout to potential root cause when URL is very long
- Adds `PGRST_TIMEOUT` error code for easier error handling

```typescript
// Example error response for timeout with long URL:
{
  error: {
    message: "AbortError: The operation was aborted",
    details: "...",
    hint: "Request was aborted (timeout or manual cancellation). Note: Your request URL is 9889 characters, which may exceed server limits. Consider using views or selecting fewer fields.",
    code: "PGRST_TIMEOUT"
  },
  data: null,
  count: null,
  status: 0,
  statusText: ""
}
```

## Breaking Changes
**None.** All features are:
- Opt-in (timeout requires explicit configuration)
- Additive only (new fields in error responses, console warnings)
- Preserve existing behavior when not configured

## Usage

```typescript
// Before (still works exactly the same)
const client = new PostgrestClient('https://example.com/rest/v1')

// After (with timeout protection)
const client = new PostgrestClient('https://example.com/rest/v1', {
  timeout: 30000 // optional, prevents indefinite hangs
})
```

## Summary by CodeRabbit

* **New Features**
  * Configurable request timeout and configurable URL-length limit (defaults applied) for database requests; client surfaces runtime warnings when URLs are very long.
* **Documentation**
  * Constructor/options docs updated to document timeout and URL-length limit settings.
* **Tests**
  * New tests covering timeout behavior, long-URL warnings, header-overflow scenarios, and enhanced error messages (unit and integration).

### DIFF
--- a/packages/core/postgrest-js/tsconfig.json
+++ b/packages/core/postgrest-js/tsconfig.json
@@ -10,7 +10,6 @@
     "rootDir": "src",
     "sourceMap": true,
     "target": "ES2017",
-    "lib": ["ES2022"],
 
     "strict": true,
 


steve-chavez:
>  URL Length Warning (PostgrestQueryBuilder.ts)
> Console warning when select parameter exceeds 8000 characters

Could we instead have the above protection for the whole URL? Because this can also happen when filters are too large: 
- https://github.com/supabase/postgrest-js/issues/393
- https://github.com/supabase/postgrest-js/issues/292

steve-chavez:
And when filters are too large, we can give a hint to solve the error by using RPC to send the values.

steve-chavez:
> 3. Enhanced Error Messages (PostgrestBuilder.ts)
> AbortError responses now include helpful hints
> Links timeout to potential root cause when URL is very long
> Adds PGRST_TIMEOUT error code for easier error handling

Could we leave `code` empty or perhaps have another prefix? The problem is that postgREST uses the `PGRST` prefix for its own errors ([ref](https://docs.postgrest.org/en/v14/references/errors.html#postgrest-error-codes)) so `PGRST_TIMEOUT` could be confused to be a backend error.

I think leaving it empty should be fine as it's a pure client-side error.

packages/core/postgrest-js/src/PostgrestBuilder.ts
```diff
@@ -247,12 +249,24 @@ export default abstract class PostgrestBuilder<
...
         }
 
+        // Handle AbortError specially with helpful hints
+        if (fetchError?.name === 'AbortError' || fetchError?.code === 'ABORT_ERR') {
+          code = 'PGRST_TIMEOUT'
+          hint = 'Request was aborted (timeout or manual cancellation)'
+
+          // If URL is very long, add helpful hint about potential cause
+          const urlLength = this.url.toString().length
+          if (urlLength > 8000) {
```
steve-chavez: Do we know what's the maximum URL length for Supabase services?

IMO perhaps it would be better to make this configurable on the postgrest-js side. Then on supabase-js set the max length for Supabase. This so `postgrest-js` stays generic.

packages/core/postgrest-js/src/PostgrestBuilder.ts
```diff
@@ -247,12 +249,24 @@ export default abstract class PostgrestBuilder<
...
+            hint += `. Note: Your request URL is ${urlLength} characters, which may exceed server limits. Consider using views or selecting fewer fields.`
```
steve-chavez: This message would be misleading in the cases mentioned on the [previous comment](https://github.com/supabase/supabase-js/pull/2078#issuecomment-3818447934) -- when values of the filters are too big.

(like a big `col=eq.<laaarge>` or a big `col=in.(1,2,3,4,.......,1000000)` for both cases changing to an RPC solves it)

Maybe we can fit both hints here or perhaps just leave the second part empty.

"If selecting too many fields consider using views or if sending large filters consider using an RPC instead"

steve-chavez (approved):

PR #2079, opened 2026-01-30
feat(storage): add setHeader method to BaseApiClient

dabarrell:
### What changed?

Added a `setHeader(name, value)` method to `BaseApiClient` in storage-js. This allows setting per-request HTTP headers on any storage operation, matching the existing pattern in postgrest-js which was added in https://github.com/supabase/postgrest-js/pull/550.

The method creates a shallow copy of the headers object to avoid mutating shared state between instances (e.g. between `StorageClient` and `StorageFileApi` returned by `from()`).

postgrest-js supports per-request header overrides via `setHeader`, but storage-js has no equivalent. Setting headers for storage requests currently requires creating a new Supabase client, setting the header using `options.global.headers` – not preferable when headers need to be set dynamically per request.

This PR enables use cases like:

```ts
supabase.storage.from('bucket').setHeader('x-custom', 'value').upload(...)
```

## 📸 Screenshots/Examples

```ts
// Set a single header
const { data, error } = await supabase.storage
  .from('avatars')
  .setHeader('custom-header', 'value')
  .upload('path/to/file.txt', fileBody)

// Chain multiple headers
const { data, error } = await supabase.storage
  .from('avatars')
  .setHeader('x-header-a', '1')
  .setHeader('x-header-b', '2')
  .download('path/to/file.txt')
```

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

storage-js currently uses an object to store headers, whereas postgrest-js transitioned from an object to the standard `Headers` API ~8mo ago in https://github.com/supabase/postgrest-js/pull/619. I chose to keep the change minimal, but I'd be happy to migrate to `Headers` in this PR if you'd prefer.

## Summary by CodeRabbit

* **New Features**
  * Added method to set custom headers on API clients with support for method chaining
  * Headers are applied immutably without affecting original configuration

* **Tests**
  * Added comprehensive test coverage for header-setting functionality

### DIFF
--- a/packages/core/storage-js/src/lib/common/BaseApiClient.ts
+++ b/packages/core/storage-js/src/lib/common/BaseApiClient.ts
@@ -47,6 +47,19 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
   }
 
   /**
+   * Set an HTTP header for the request.
+   * Creates a shallow copy of headers to avoid mutating shared state.
+   *
+   * @param name - Header name
+   * @param value - Header value
+   * @returns this - For method chaining
+   */
+  public setHeader(name: string, value: string): this {
+    this.headers = { ...this.headers, [name]: value }
+    return this
+  }
+
+  /**
    * Handles API operation with standardized error handling
    * Eliminates repetitive try-catch blocks across all API methods
    *
--- a/packages/core/storage-js/test/common/BaseApiClient.test.ts
+++ b/packages/core/storage-js/test/common/BaseApiClient.test.ts
@@ -1,5 +1,6 @@
 import BaseApiClient from '../../src/lib/common/BaseApiClient'
 import { StorageError, StorageApiError } from '../../src/lib/common/errors'
+import { StorageClient } from '../../src/index'
 
 // Test implementation of BaseApiClient
 class TestApiClient extends BaseApiClient {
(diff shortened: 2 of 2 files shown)


mandarini:
Hi @dabarrell ! Thank you for the amazing PR! :D I gave this some thought, and I decided to take you up on your offer. Yes please! Since you're already adding setHeader() to match `postgrest-js`, let's go all the way and **migrate to the Headers API** too. It'll keep `storage-js` consistent with `postgrest-js` and we won't have to do this migration separately later. Plus it's backward compatible since HeadersInit accepts plain objects. :D Wdyt?

dabarrell:
@mandarini Happy to. I've started implementing it, but I've realised that auth-js, functions-js, supabase-js and realtime-js are similar to storage-js in that they don't use `Headers` – it's only postgrest-js that does. Given the `Headers` API is pretty simple and requires some (simple) workarounds for e.g. merging headers I found myself making some design choices and therefore wanting to create some helper functions to standardise, but AFAICT there isn't an approach in this repo for sharing common helpers between packages.

I'm happy to continue, but given it might need some back and forth, would you be okay merging this PR, and I'll follow up with a separate one to convert `storage-js` to Headers? We can figure out on that PR what patterns we want to use, then I can follow on to apply them to the other packages and eventually get them all onto Headers

mandarini (approved):

PR #2081, opened 2026-01-30
feat(supabase): add bypassAuthSession option for service-role clients

### LINKED ISSUE #2045: Secret Key + security_invoker view not returning results
### Describe the bug

Hi supabase team,

I'm unable to access a view from a js client with a secret key.

I've created a view that joins a `profiles` table to `auth.last_sign_in_at`.

I can query this fine when its public (and thus unsafe), but when i add:
```
alter view profiles_plus_auth
set (security_invoker = true);
```
the view returns `null` from the js client.

I would have expected that with the secret key being used, it would bypass any RLS issues and return the data from the table, while not allowing public or authenticated users to see it.

I know that my secret key works for other tables and commands like ` const data = await supabase.auth.admin.listUsers();`, just not for the view with auth involved.

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

1. Create a js client using the secret key.
2. Create the tables
```sql
create view public.profiles_plus_auth as
select
  p.id,
  p.name,
  au.last_sign_in_at
from
  profiles p
  left join auth.users au on p.id = au.id;

alter view profiles_plus_auth
set (security_invoker = true);
```
3. Try to query this table.

### System Info

```shell
"@supabase/supabase-js": "2.91.0"

 System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 132.66 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.8.0 - /Users/schoko/.nvm/versions/node/v24.8.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 11.6.0 - /Users/schoko/.nvm/versions/node/v24.8.0/bin/npm
    bun: 1.2.19 - /opt/homebrew/bin/bun
    Deno: 2.4.5 - /opt/homebrew/bin/deno
    Watchman: 2025.08.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 142.0.7444.176
    Firefox: 121.0.1
    Safari: 17.5
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

saakshigupta2002:
## 🔍 Description

This PR adds a new `global.bypassAuthSession` option to `SupabaseClient` to fix an issue where service-role clients would incorrectly use user session tokens instead of the service-role key when querying `security_invoker` views.

### What changed?

- Added `bypassAuthSession?: boolean` option to `SupabaseClientOptions.global`
- Modified `_getAccessToken()` method in `SupabaseClient` to check `bypassAuthSession` before retrieving session tokens
- Added comprehensive test coverage with 4 new unit tests validating the new behavior
- Updated TypeScript type definitions with detailed JSDoc documentation

When using a service-role key with `security_invoker = true` views or functions, PostgREST runs queries with the caller's permissions. The issue occurred because `_getAccessToken()` would prioritize the user's session token over the service-role key when both existed.

This caused permission errors when:
1. A client was initialized with a service-role key
2. A user session existed (from previous authentication)
3. Queries were made to `security_invoker` views that accessed restricted schemas (like `auth.users`)

The requests would run with the `authenticated` role (from the session token) instead of the `service_role` (from the API key), resulting in "permission denied" errors even though the client was configured with full admin privileges.

Closes #2045

### Test Coverage 

All 4 new tests pass successfully:

✅ Uses supabaseKey when bypassAuthSession: true even with active session
✅ Uses session token when bypassAuthSession: false (default behavior)
✅ Passes correct token to fetchWithAuth wrapper
✅ Stores bypassAuthSession value correctly

### Alternative approaches considered:

- Always using service_role key for admin clients - Rejected because it would break existing functionality where admins might want to perform user-level operations
- Automatic detection based on key type - Rejected because it's not always possible to distinguish service_role keys from anon keys
- Per-query option - Rejected in favor of client-level configuration for consistency
- This approach provides explicit control while maintaining backward compatibility.

## Summary by CodeRabbit

* **New Features**
  * Added bypassAuthSession configuration option. When enabled, the client uses the API key for Authorization on all requests, bypassing active user session tokens.

* **Tests**
  * Added test suite validating bypassAuthSession behavior, token selection, and header usage across authentication scenarios.

### DIFF
--- a/packages/core/supabase-js/src/lib/types.ts
+++ b/packages/core/supabase-js/src/lib/types.ts
@@ -151,9 +151,15 @@ export type SupabaseClientOptions<SchemaName> = {
      *
      * @default false
      *
+     * @security **WARNING: Only use this option in server-side code** (Node.js,
+     * Edge Functions, serverless functions, etc.). Never expose service role keys
+     * in client-side/browser applications as they bypass all Row Level Security
+     * (RLS) policies and grant full unrestricted database access. For admin
+     * operations in browsers, create server-side API endpoints instead.
+     *
      * @example
      * ```ts
-     * // Create an admin client that always uses the service-role key
+     * // Server-side only: Create an admin client that always uses the service-role key
      * const adminClient = createClient(url, serviceRoleKey, {
      *   global: { bypassAuthSession: true }
      * })


mandarini:
@saakshigupta2002 thank you for your PR. This sparked interesting conversations within the team. We decided not to merge it though. I personally think it will normalize an antipattern. The team agreed that we are going to make the experience better in the next major version. Right now, a workaround exists to get the behaviour one expects.

Thank you for taking the time to contribute. Contributions like yours are what makes our tools better. Thank you for doing this for Supabase! 💚

PR #2083, opened 2026-02-02
feat(postgrest): based select() API to postgrest-js

MildTomato:
### What changed?

Added array-based `select()` API to postgrest-js that provides IDE autocomplete while maintaining full backward compatibility with string-based queries.

- New `select-builder.ts` with typed specs (`FieldSpec`, `RelationSpec`, `SpreadSpec`, `CountSpec`) and `serializeSelectSpec()`
- Updated `PostgrestQueryBuilder.select()` and `PostgrestTransformBuilder.select()` to accept arrays
- Exported new types from package index
- claude code plan [can be viewed here](https://github.com/supabase/supabase-js/pull/2083/changes#diff-b923cfc47fff3e556d167a2632af6688d76b798865931d61a91e87145df03c3a) to help with review

The string-based select API doesn't provide IDE autocomplete or compile-time validation. This array-based alternative enables better developer experience while keeping full backward compatibility.

```typescript
// String (still works)
.select('id, display_name:username, posts!inner(id)')

// Array (new - with autocomplete)
.select([
  'id',
  { column: 'username', as: 'display_name' },
  { relation: 'posts', inner: true, select: ['id'] }
])
```

## 🔄 Breaking changes

## 📋 Checklist

- [x] I have run `nx format` to ensure consistent code formatting

### DIFF
--- a/packages/core/postgrest-js/docs/SELECT_BUILDER_PLAN.md
+++ b/packages/core/postgrest-js/docs/SELECT_BUILDER_PLAN.md
@@ -370,16 +370,13 @@ type SelectSpec = string | SelectItem[] // string for backward compat only
 ### New Files
 
 1. `packages/core/postgrest-js/src/select-query-parser/select-builder.ts`
-
    - Type definitions for SelectSpec, FieldSpec, RelationSpec, etc.
    - `serializeSelect()` function to convert array to string
 
 2. `packages/core/postgrest-js/src/select-query-parser/result-from-spec.ts`
-
    - Type inference for array-based specs (parallel to result.ts)
 
 3. `packages/core/postgrest-js/test/select-builder.test.ts`
-
    - Unit tests for serialization
 
 4. `packages/core/postgrest-js/test/select-builder.test-d.ts`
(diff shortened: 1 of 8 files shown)


packages/core/postgrest-js/docs/SELECT_BUILDER_PLAN.md
```diff
@@ -0,0 +1,111 @@
+# Plan: Add Array-Based Type-Safe select() API
```
mandarini, line 1: Let's remove this file

mandarini (commented):
I like this, I know we have discussed it a lot. Changes and implementation look good to me.

mandarini:
@MildTomato  Getting back to this after forever 😅 . So, I think I'd push back on shipping in this shape:

1. **No autocomplete.** `column`/`relation` are just `string`, so typos compile. The reason to add an object API is type-safe field names, which is not implemented.

2. **Return type regresses.** `Query` defaults to `'*'` on the array overload, so `.select(['id', 'name'])` returns the full row, not `Pick<Row, 'id' | 'name'>`. The string form is narrower today.

3. **Two APIs forever** without a story for collapsing them. Smaller: plan doc shouldn't ship inside the package; the "integration" tests only assert URL strings (no real PostgREST round-trip); `!left` is a documented no-op, so let's not expose it.

Ideal path before merge: `column: keyof Row`, `relation` constrained against relationships, and result-type narrowing from array contents.

So, given these, do you think it's worth pursuing? If yes, I can do the remaining work.

Issue #1394, opened 2025-04-21
createClient pattern lacks of extensibility with multi-schema usage (singleton gotrue warning)


softmarshmallow:
I am constantly getting warning from supabase.

```
Multiple GoTrueClient instances detected in the same browser context. It is not an error, but this should be avoided as it may produce undefined behavior when used concurrently under the same storage key. 
```

<img width="824" alt="Image" src="https://github.com/user-attachments/assets/c99303e0-22a3-43d9-9e09-348aef471476" />

I am using multiple clients with `singleton: false` since I am using multiple schema, and client instanciation with singleton does not allow it.

e.g.

```ts
import type { Database } from "@/database.types";
import { createBrowserClient as _createBrowserClient } from "@supabase/ssr";

const __create_browser_client = <
  SchemaName extends string & keyof Database = "public" extends keyof Database
    ? "public"
    : string & keyof Database,
>(
  schema: SchemaName
) =>
  _createBrowserClient<Database, SchemaName>(
    process.env.NEXT_PUBLIC_SUPABASE_URL!,
    process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!,
    {
      db: {
        schema: schema,
      },
      isSingleton: false,
    }
  );

export const createBrowserClient = () =>
  __create_browser_client<"public">("public");

export const createBrowserFormsClient = () =>
  __create_browser_client<"grida_forms">("grida_forms");

export const createBrowserCommerceClient = () =>
  __create_browser_client<"grida_commerce">("grida_commerce");

export const createBrowserCanvasClient = () =>
  __create_browser_client<"grida_canvas">("grida_canvas");

export const createBrowserWWWClient = () =>
  __create_browser_client<"grida_www">("grida_www");

export const createBrowserWestReferralClient = () =>
  __create_browser_client<"grida_west_referral">("grida_west_referral");

```

(this is not a `@supabase/ssr` problem, but a `supabase-js` problem)

What is the best practice when using multiple schema? do you have plans for supporting switching schema on-the-go with a singleton instance?

e.g.

```ts
// switch schema with builder pattern
const client = createClient()
client.schema("other_schema").from("...")


// or.. (multi singleton by group)

const client = createClient({singleton: true, singletone_group: "public"})
const otherClient = createClient({singleton: true, singletone_group: "other_schema"})


// core client - consumer client pattern
const base = createCoreClient()
const client = createClient(base)
[...]

PR #2086, opened 2026-02-04
fix(auth) : use schema specific storage keys for multi schema auth isolation

7ttp:
## Summary

Fixes auth state collision when multiple Supabase clients use different schemas but share the same storage key.

## Problem
When creating multiple clients with different schemas (`public` and `private`), 
they shared the same storage key, causing auth state to overwrite each other. 
This broke multi tenant applications using schema-based isolation and triggered the "Multiple GoTrueClient instances detected" warning even with `isSingleton: false`

## Solution
Modified storage key generation to include the schema name, 
ensuring each schema gets its own isolated storage namespace. 
BC cause clients using the default `public` schema continue using the same storage key format.

## Related
- Closes https://github.com/supabase/supabase-js/issues/1394

## Release Notes

* **New Features**
  * Authentication tokens are now isolated by database schema, preventing conflicts when using multiple schemas in the same project.

* **Tests**
  * Added unit tests to verify correct token storage behavior across different database schemas.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -124,7 +124,9 @@ export default class SupabaseClient<
     this.functionsUrl = new URL('functions/v1', baseUrl)
 
     // default storage key uses the supabase project ref as a namespace
-    const defaultStorageKey = `sb-${baseUrl.hostname.split('.')[0]}-auth-token`
+    const schemaName = options?.db?.schema
+    const schemaSuffix = schemaName && schemaName !== 'public' ? `-${schemaName}` : ''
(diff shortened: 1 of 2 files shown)


mandarini:
Thank you @7ttp . If we release this, customers with non-public schemas (`db: { schema: 'other_schema' }`) will be logged out. I am working with the team to see if we consider this ok or not for now.

mandarini:
After team discussion, we've identified that this change would be **breaking** for existing users: anyone using non-public schemas would have their users logged out and need to sign in again after this update is released.

For now, our official recommendation is to use different `storageKey` values for each schema-specific client, which provides the same isolation without the breaking change. You can see the full guidance in [our response on the original issue](https://github.com/supabase/supabase-js/issues/1394).

We're keeping track of multi-schema UX improvements for v3 where we can introduce these kinds of changes more safely with proper migration paths.

Thanks again for the contribution! 💚

PR #2090, opened 2026-02-04
fix(storage): expose fetch parameters in download method

mandarini:
## Problem

Issue [#41841](https://github.com/supabase/supabase/issues/41841) reports that `storage.download()` returns stale/cached data after file updates, particularly in Supabase Edge Functions (Deno runtime).

The `download()` method currently doesn't expose a way to pass fetch options like `cache: 'no-store'` to the underlying HTTP request, even though the infrastructure exists in the `get()` function.

## Solution

Add an optional `parameters?: FetchParameters` argument to the `download()` method, following the same pattern already used by the `list()` method.

This allows users to opt into cache control when needed:

```js
// Edge Functions - bypass cache
const { data } = await supabase.storage
  .from('avatars')
  .download('file.png', {}, { cache: 'no-store' })

// Abort signal support
const { data } = await supabase.storage
  .from('avatars')
  .download('file.png', {}, { signal: controller.signal })

// Default behavior (unchanged)
const { data } = await supabase.storage
  .from('avatars')
  .download('file.png')
```

Inspired by https://github.com/supabase/supabase-js/pull/2022 and @Akkii88, thank you!

## Summary by CodeRabbit

* **New Features**
  * Downloads now accept optional fetch parameters (including cache directives) for finer control over HTTP caching and other fetch behaviors.
  * Download calls can accept an abort signal and remain backward-compatible.

* **Documentation**
  * Added usage examples showing cache control and abort scenarios.

* **Tests**
  * Added tests covering cache options, abort behavior, transform usage with fetch parameters, and backward-compatibility.

### DIFF
--- a/packages/core/storage-js/test/common/fetch.test.ts
+++ b/packages/core/storage-js/test/common/fetch.test.ts
@@ -323,7 +323,14 @@ describe('Common Fetch', () => {
       })
 
       it('should support all cache values', async () => {
-        const cacheValues = ['default', 'no-store', 'reload', 'no-cache', 'force-cache', 'only-if-cached'] as const
+        const cacheValues = [
+          'default',
+          'no-store',
+          'reload',
+          'no-cache',
+          'force-cache',
+          'only-if-cached',
+        ] as const
 
         for (const cacheValue of cacheValues) {
           const responseData = { result: 'success' }


packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -718,21 +719,46 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
    *     }
    *   })
    * ```
+   *
+   * @example Download with cache control (useful in Edge Functions)
+   * ```js
+   * const { data, error } = await supabase
+   *   .storage
+   *   .from('avatars')
+   *   .download('folder/avatar1.png', {}, { cache: 'no-store' })
```
ferhatelmas, line 728: is `cache` included into the type?

packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -718,21 +719,46 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
...
+   *
+   * @example Download with abort signal
+   * ```js
+   * const controller = new AbortController()
+   * setTimeout(() => controller.abort(), 5000)
+   *
+   * const { data, error } = await supabase
+   *   .storage
+   *   .from('avatars')
+   *   .download('folder/avatar1.png', {}, { signal: controller.signal })
```
ferhatelmas, line 739: This is a bit ugly, using a single options object would be more desirable

packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -718,21 +719,46 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
...
+   * ```
    */
   download<Options extends { transform?: TransformOptions }>(
     path: string,
-    options?: Options
+    options?: Options,
+    parameters?: FetchParameters
```
ferhatelmas, line 745: even if simple, having a test that exercises this new parameter would be awesome

packages/core/storage-js/src/packages/StorageFileApi.ts
mandarini: I think we should keep the two-parameter pattern, since it maintains consistency with existing `list()` and `listV2()` methods (which accept options and fetchparams). I think it shows the separation of concerns between transform options and fetch parameters.

packages/core/storage-js/src/packages/StorageFileApi.ts
mandarini: Ooops not included!

packages/core/storage-js/src/packages/StorageFileApi.ts
mandarini: Adding!

packages/core/storage-js/src/lib/types.ts
```diff
@@ -202,6 +202,17 @@ export interface FetchParameters {
...
+
+  /**
+   * Controls how the request interacts with the browser's HTTP cache.
+   * - 'default': Use standard cache behavior
+   * - 'no-store': Bypass cache entirely (useful in Edge Functions)
+   * - 'reload': Bypass cache but update it with response
+   * - 'no-cache': Validate with server before using cached response
+   * - 'force-cache': Use cache even if stale
+   * - 'only-if-cached': Only use cache, fail if not cached
+   */
```
ferhatelmas, line 214: Awesome to have inline definitions

packages/core/storage-js/src/lib/types.ts
ferhatelmas, line 200: It seems there is a duplication with `FetchParameters` under common lib (`lib/common/fetch.ts`). Not a hard blocker but would be nice to get rid of it

packages/core/storage-js/test/common/fetch.test.ts
```diff
@@ -278,6 +278,79 @@ describe('Common Fetch', () => {
...
+          headers: {},
+          signal: controller.signal,
+          cache: 'no-store',
+        })
+      })
+
+      it('should support all cache values', async () => {
+        const cacheValues: Array<
+          'default' | 'no-store' | 'reload' | 'no-cache' | 'force-cache' | 'only-if-cached'
+        > = ['default', 'no-store', 'reload', 'no-cache', 'force-cache', 'only-if-cached']
```
ferhatelmas:
```suggestion
        const cacheValues = ['default', 'no-store', 'reload', 'no-cache', 'force-cache', 'only-if-cached'] as const
```

packages/core/storage-js/test/storageFileApi.test.ts
```diff
@@ -665,6 +665,77 @@ describe('Object API', () => {
...
+  it('download with abort signal', async () => {
+    const controller = new AbortController()
+    const uploadRes = await storage.from(bucketName).upload(uploadPath, file)
+    expect(uploadRes.error).toBeNull()
+
+    const downloadBuilder = storage
+      .from(bucketName)
+      .download(uploadPath, {}, { signal: controller.signal })
+
+    controller.abort()
```
ferhatelmas: This might be racy for test according to the speed of download. To prevent flakiness, it's better to call abort before download builder

packages/core/storage-js/src/lib/types.ts
mandarini: Fixed

PR #2093, opened 2026-02-05
feat(postgrest): add helpful hints for permission errors

mandarini:
## Description

Enhances error handling in postgrest-js to provide actionable hints when users encounter permission-related errors. When Supabase projects revoke default GRANT permissions on tables/functions, users now receive helpful guidance with copy-paste ready SQL commands instead of cryptic error messages.

### What Changed

- Added permission error detection for three error codes:
- **42501** (PostgreSQL insufficient privilege) - Direct permission denied
- **PGRST205** (PostgREST table not found) - Could be missing permissions or doesn't exist
- **PGRST202** (PostgREST function not found) - Could be missing permissions or doesn't exist

- Implemented `generatePermissionHint()` method that provides:
  - Operation-specific GRANT commands (SELECT for GET, INSERT for POST, UPDATE for PATCH, DELETE for DELETE)
  - EXECUTE grants for functions
  - Schema-qualified object names (handles both `table` and `schema.table` formats)
  - Copy-paste ready SQL for Supabase SQL Editor

- Only enhances errors when `hint` is `null` (preserves existing PostgREST hints)

### Example Output

Before:

  ```json
  {
    "code": "42501",
    "message": "permission denied for table test",
    "hint": null
  }
```

After (GET request):

```json
 {
    "code": "42501",
    "message": "permission denied for table test",
    "hint": "Missing SELECT permission on table 'test'. Grant permission with:\n\nGRANT SELECT ON
  public.test TO anon, authenticated, service_role;\n\nNote: Adjust schema if not 'public', and
  specify only the roles you want to grant access to."
  }
```

After (POST request - operation-specific):

```json
  {
    "code": "42501",
    "message": "permission denied for table test",
    "hint": "Missing INSERT permission on table 'test'. Grant permission with:\n\nGRANT INSERT ON
  public.test TO anon, authenticated, service_role;\n\nNote: Adjust schema if not 'public', and
  specify only the roles you want to grant access to."
  }
```

## Context

This addresses the issue where AI assistants suggest incorrect solutions (like "modify RLS policies") when users encounter permission errors after revoking default GRANT permissions. The hints now correctly guide users to add GRANT permissions (table-level security) rather than RLS policies (row-level security), which are different security layers.

### DIFF
--- a/packages/core/postgrest-js/test/embeded_functions_join.test.ts
+++ b/packages/core/postgrest-js/test/embeded_functions_join.test.ts
@@ -1264,7 +1264,13 @@ describe('embeded functions select', () => {
         "error": {
           "code": "PGRST202",
           "details": "Searched for the function public.created_ago with parameter id or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.",
-          "hint": null,
+          "hint": "Function 'public.created_ago' not found. This could mean:
+      1. The function doesn't exist, OR
+      2. You don't have permission to see it
+
+      If the function exists, grant permission with:
+
+      GRANT EXECUTE ON FUNCTION public.created_ago TO anon, authenticated, service_role;",
           "message": "Could not find the function public.created_ago(id) in the schema cache",
         },
         "status": 404,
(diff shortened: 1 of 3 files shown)


packages/core/postgrest-js/test/embeded_functions_join.test.ts
```diff
@@ -1319,7 +1325,13 @@ describe('embeded functions select', () => {
...
           "code": "PGRST202",
           "details": "Searched for the function public.days_since_event with parameter id or with a single unnamed json/jsonb parameter, but no matches were found in the schema cache.",
-          "hint": null,
+          "hint": "Function 'public.days_since_event' not found. This could mean:
+      1. The function doesn't exist, OR
+      2. You don't have permission to see it
+
+      If the function exists, grant permission with:
+
+      GRANT EXECUTE ON FUNCTION public.days_since_event TO anon, authenticated, service_role;",
```
steve-chavez, line 1334: One problem with this direction is that here we can only be generic and mention all the roles. 

When doing it via extension we can list the precise role that's missing the privilege.

PR #2106, opened 2026-02-08
fix(auth): recover from orphaned navigator locks via steal fallback

ElliotPadfield:
## Summary

Fixes https://github.com/supabase/supabase-js/issues/2111

When a Navigator Lock is held indefinitely (e.g., due to React Strict Mode's double-mount/unmount leaving an orphaned lock callback in `GoTrueClient._acquireLock`'s `pendingInLock` drain loop), all subsequent auth operations (`getUser()`, `signInWithPassword()`, etc.) hang forever.

**Root cause:** The existing `acquireTimeout` mechanism uses `AbortController.abort()` to cancel pending lock requests. Per the [Web Locks API spec](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request), aborting the signal only removes **pending** (waiting-to-acquire) requests from the queue — it has **no effect** on an already-held lock. So when a lock is orphaned:

1. The held lock callback never returns → lock stays held forever
2. Subsequent callers queue up, timeout, get `AbortError`, and give up
3. The auth system is permanently broken until all tabs are closed

**Fix:** When lock acquisition times out with `AbortError`, instead of propagating the error, retry with `{ steal: true }`. Per the spec, this releases any currently held lock with the same name and grants the request immediately. The previous holder's callback continues running to completion but no longer blocks other callers.

### Changes
- `packages/core/auth-js/src/lib/locks.ts` — Wrap `navigator.locks.request()` in try/catch; on `AbortError` with positive `acquireTimeout`, retry with `{ steal: true }` to recover from orphaned locks
- `packages/core/auth-js/test/lib/locks.test.ts` — Add 3 tests covering: orphaned lock recovery via steal, non-AbortError passthrough, no steal attempt with negative timeout

## Test plan
- [x] All 10 existing + new unit tests pass (`jest --testPathPattern='test/lib/locks'`)
- [x] Manual verification: create a Next.js app with React Strict Mode, call `getUser()` in a `useEffect`, trigger the orphaned lock scenario (visible via `navigator.locks.query()`), confirm auth recovers instead of hanging
- [ ] Verify no regression in multi-tab session synchronization

## Release Notes

* **Bug Fixes**
  * Improved lock acquisition reliability with enhanced timeout handling and automatic recovery from orphaned locks.
  * Enhanced error messages and logging for timeout scenarios to clarify potential causes and recovery actions.

* **Tests**
  * Added test coverage for lock recovery from timeouts, error propagation, and timeout configuration edge cases.

### DIFF
--- a/packages/core/auth-js/src/lib/locks.ts
+++ b/packages/core/auth-js/src/lib/locks.ts
@@ -115,10 +115,13 @@ export async function navigatorLock<R>(
 
   // MDN article: https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request
 
-  // Wrapping with await Promise.resolve() is done as some libraries like zone.js
-  // patch the Promise object to track execution context. We use await instead of
-  // .then() to avoid Firefox content script security errors where accessing .then()
-  // on cross-context promises is forbidden.
+  // Wrapping navigator.locks.request() with a plain Promise is done as some
+  // libraries like zone.js patch the Promise object to track the execution
+  // context. However, it appears that most browsers use an internal promise
+  // implementation when using the navigator.locks.request() API causing them
+  // to lose context and emit confusing log messages or break certain features.
+  // This wrapping is believed to help zone.js track the execution context
+  // better.
   await Promise.resolve()
 
   try {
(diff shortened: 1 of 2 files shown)


ElliotPadfield:
Once I'd diagnosed & planned a fix, used Claude Code to help validate & polish. Happy to take any feedback & tweak/resubmit. Passes formatting, tests, build. Thanks to @coppinger for troubleshooting support as well.

coppinger:
+1 that this is causing major headaches

mandarini:
Thank you for this PR @ElliotPadfield , it has sparked an interesting (and educative for me) conversation within the team. As I mentioned [here](https://github.com/supabase/supabase-js/issues/2111#issuecomment-3885344361), before considering using the `steal` solution, we would like to see a reproduction repository where we can verify the orphaning. Would this be possible for you to share with me? It would be indeed super helpful!

ElliotPadfield:
> Thank you for this PR @ElliotPadfield , it has sparked an interesting (and educative for me) conversation within the team. As I mentioned [here](https://github.com/supabase/supabase-js/issues/2111#issuecomment-3885344361), before considering using the `steal` solution, we would like to see a reproduction repository where we can verify the orphaning. Would this be possible for you to share with me? It would be indeed super helpful!

Will throw something together & coordinate with @coppinger. Thanks for your attention to the issue.

mandarini:
@ElliotPadfield 

After lots of internal discussions, we have agreed to approve this PR, until we come up with a solution to the core of the problem, which may bring wider architectural changes.

To approve this PR, can you please resolve the conflicts with `master`?

After this PR is merged, we also agreed to decrease the timeout to 5s instead of 10s, but this is out of scope for this PR, since the change was introduced elsewhere.

andreas-myklebust:
@mandarini Any way you could expedite the process by resolving said conflicts? Been tracking this issue for over a week now, would appreciate it greatly to have it resolved.

mandarini (approved):

Issue #2107, opened 2026-02-09
Serializer.decode() crashes when Realtime server sends non-array WebSocket message


whaaaley:
### Describe the bug

Serializer.decode() crashes when the Realtime server sends a JSON object over the WebSocket instead of the expected array format. This crashes the entire process with an unhandled TypeError.

### Library affected

supabase-js

### Steps to reproduce

1. Subscribe to postgres_changes on a table
2. Trigger an INSERT via a cron job or external process
3. The Realtime server occasionally sends a non-array JSON message
4. Process crashes with the stack trace below

### System Info

```shell
- @supabase/supabase-js: ^2.56.0
- @supabase/realtime-js: 2.93.3
- Runtime: Deno 2.6.3
- Local Supabase via supabase start
```

### Used Package Manager

npm

### Logs

```
error: Uncaught TypeError: jsonPayload is not iterable
            const [join_ref, ref, topic, event, payload] = jsonPayload;
                                                           ^
    at Serializer.decode (realtime-js/dist/main/lib/serializer.js:104:60)
    at RealtimeClient._onConnMessage (realtime-js/dist/main/RealtimeClient.js:455:14)
    at WebSocket.conn.onmessage (realtime-js/dist/main/RealtimeClient.js:519:47)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://supabase.com/docs/reference/javascript/introduction).
- [x] Check that there isn't [already an issue](https://github.com/supabase/supabase-js/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a Supabase JS Library issue and not an issue with the Supabase platform. If it's a Supabase platform related bug, it should likely be reported to [supabase/supabase](https://github.com/supabase/supabase) instead.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/supabase/supabase-js/discussions) or join our [Discord Chat Server](https://discord.supabase.com//).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

7ttp:
hi @whaaaley 💚🙌
thanks for reporting! Unfortunately, 
[...]

PR #2115, opened 2026-02-12
fix(realtime): defensive check for serializer

7ttp:
## Summary
Adds defensive validation to Serializer.decode() to prevent crashes 

## Problem
Serializer.decode() crashes with TypeError when attempting to destructure non-array JSON payloads, terminating the process.
(in edge cases)

## Solution
Added array validation and try-catch to handle invalid messages properly by returning an empty object instead of crashing.
(following existing pattern)

## Related
- Closes https://github.com/supabase/supabase-js/issues/2107

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data serialization reliability with enhanced error handling for malformed JSON payloads, ensuring the system gracefully manages invalid data formats without crashing.
  * Added robust payload validation to verify data structure integrity, returning safe defaults when encountering unexpected or incomplete payload formats.

### DIFF
--- a/packages/core/realtime-js/src/lib/serializer.ts
+++ b/packages/core/realtime-js/src/lib/serializer.ts
@@ -130,9 +130,14 @@ export default class Serializer {
     }
 
     if (typeof rawPayload === 'string') {
-      const jsonPayload = JSON.parse(rawPayload)
-      const [join_ref, ref, topic, event, payload] = jsonPayload
-      return callback({ join_ref, ref, topic, event, payload })
+      try {
+        const jsonPayload = JSON.parse(rawPayload)
+        if (!Array.isArray(jsonPayload) || jsonPayload.length < 5) return callback({})
+        const [join_ref, ref, topic, event, payload] = jsonPayload
+        return callback({ join_ref, ref, topic, event, payload })
+      } catch {
+        return callback({})
+      }
     }
 
     return callback({})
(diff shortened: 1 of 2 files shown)


7ttp:
@mandarini

mandarini:
Hey @7ttp!  First off, thank you so much for the PR as always! :D That said, after digging deeper into the [Realtime server code](https://github.com/supabase/realtime/blob/2ff81828ffee3a962d4c9c87bf8e5d8df1da5a92/lib/realtime_web/socket/v2_serializer.ex#L130-L156), I'm realizing we need to stop and think about this. According to the Phoenix Socket protocol that Realtime uses, the server **always** sends messages as 5-element JSON arrays: `[join_ref, ref, topic, event, payload]`. This is hardcoded throughout the Realtime codebase, so the fact that the reporter is receiving non-array messages suggests something more interesting is going on:

- Could be a server bug (would need to be fixed server-side)
- Could be network middleware (proxies/load balancers) corrupting messages
- Could be something specific to their infrastructure setup

While your defensive check would prevent the crash, it would also make it nearly impossible to debug what's actually going wrong. By silently returning an empty object, we'd lose visibility into:
- What the malformed messages actually contain
- How frequently this happens
- Whether it's a client or server issue
- If it's affecting other users

I'm going to ask the reporter for a reproduction case and the actual malformed message content on the [original issue](https://github.com/supabase/supabase-js/issues/2107). Once we understand the root cause, we can determine the right fix - which might be:

1. A server-side fix in Realtime
2. A client-side defensive check **with proper logging/telemetry**
3. Documentation for users about infrastructure considerations

I should have done this investigation before suggesting you write the fix. Sorry for sending you down this path, totally my fault for not digging deeper first!

I'm going to close this PR for now! Thank you so much for the help!!! :D

PR #2116, opened 2026-02-12
fix(storage): improve FileObject type accuracy with nullable fields

mandarini:
- FileObject: Made id, updated_at, created_at, last_accessed_at, metadata nullable (null for folders)
  - FileObject: Deprecated bucket_id, owner, buckets fields (not returned by list operations)
  - FileObjectV2: Deprecated updated_at field (API returns last_modified)
  - SearchV2Object: Made key optional (sometimes missing in responses)
  - SearchV2Object: Made id, updated_at, created_at, metadata, last_accessed_at nullable (null for folders)
  - Added FileMetadata interface to properly type metadata objects

**A clarification on the changes, and whether they are breaking or not:**

At runtime, nothing changes. The Storage API has always returned null for id, updated_at, created_at, last_accessed_at, and metadata on folder entries. This is explicitly declared as nullable in the [storage/schemas/object.ts](https://github.com/supabase/storage/blob/master/src/storage/schemas/object.ts). Likewise, bucket_id, owner, and buckets have never been included in list() responses. The previous TypeScript types were incorrect: they promised non-null values that the server never guaranteed, meaning any user code that relied on those guarantees was already silently broken at runtime for folder entries. This PR corrects that lie, the types now accurately reflect what the server has always actually returned. No JavaScript or TypeScript behavior changes. Only the type checker will now correctly surface cases where null-safety wasn't being handled.

### DIFF
--- a/packages/core/storage-js/src/lib/types.ts
+++ b/packages/core/storage-js/src/lib/types.ts
@@ -119,8 +119,6 @@ export interface FileObjectV2 {
   bucket_id: string
   /** Creation timestamp */
   created_at: string
-  /** @deprecated Use last_modified instead */
-  last_accessed_at: string
   /** File size in bytes */
   size?: number
   /** Cache control header value */
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -1169,10 +1169,9 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
   /**
    * Lists all the files and folders within a bucket using the V2 API with pagination support.
    *
-   * **Important:** For folder entries in the `folders` array, fields like `id`, `updated_at`,
-   * `created_at`, `last_accessed_at`, and `metadata` will be `null`. Only files in the
-   * `objects` array have these fields populated. The `key` field may also be missing in
-   * some responses.
+   * **Important:** Folder entries in the `folders` array only contain `name` and optionally `key` —
+   * they have no `id`, timestamps, or `metadata` fields. Full file metadata is only available
+   * on entries in the `objects` array.
    *
    * @experimental this method signature might change in the future
    *


packages/core/storage-js/src/lib/types.ts
```diff
@@ -44,34 +44,100 @@ export interface AnalyticBucket {
...
+  /** File name */
   name: string
+  /** Bucket identifier */
   bucket_id: string
-  updated_at: string
+  /** Creation timestamp */
   created_at: string
-  /** @deprecated */
+  /** @deprecated Use last_modified instead */
   last_accessed_at: string
```
ferhatelmas: this is shared by info endpoint but it isn't returned so required string would be wrong

packages/core/storage-js/src/lib/types.ts
```diff
@@ -44,34 +44,100 @@ export interface AnalyticBucket {
...
   cache_control?: string
+  /** MIME content type */
   content_type?: string
+  /** Entity tag for caching */
   etag?: string
+  /** Last modification timestamp (replaces updated_at) */
   last_modified?: string
-  metadata?: Record<string, any>
+  /** Custom file metadata */
+  metadata?: FileMetadata
```
ferhatelmas, line 133: similarly this is different in info endpoint, in info, this is user metadata (arbitrary), not the system one

packages/core/storage-js/src/lib/types.ts
```diff
@@ -44,34 +44,100 @@ export interface AnalyticBucket {
...
+  httpStatusCode: number
+  /** Any additional custom metadata stored with the file */
+  [key: string]: any
+}
+
+/**
+ * File object returned by the List V1 API (list() method)
+ * Note: Folder entries will have null values for most fields except name
+ *
+ * Warning: Some fields may not be present in all API responses. Fields like
```
ferhatelmas, line 74: for remove, following is returned but missing now: version, owner_id, user_metadata

packages/core/storage-js/src/lib/types.ts
```diff
@@ -44,34 +44,100 @@ export interface AnalyticBucket {
...
+  /** Bucket identifier */
   bucket_id: string
-  updated_at: string
+  /** Creation timestamp */
   created_at: string
-  /** @deprecated */
+  /** @deprecated Use last_modified instead */
   last_accessed_at: string
+  /** File size in bytes */
   size?: number
```
ferhatelmas, line 123: these fields are nullable

packages/core/storage-js/src/lib/types.ts
ferhatelmas, line 278: realized `nextCursorKey?: string` is missing

packages/core/storage-js/src/lib/types.ts
mandarini: I think the DELETE endpoint returns only `{ message: 'Successfully deleted' }` ([deleteObject.ts line 48](https://github.com/supabase/storage/blob/master/src/http/routes/object/deleteObject.ts#L48))

packages/core/storage-js/src/lib/types.ts
mandarini: good catch

packages/core/storage-js/src/lib/types.ts
mandarini: good catch

packages/core/storage-js/src/lib/types.ts
mandarini: good catch

packages/core/storage-js/src/lib/types.ts
```diff
@@ -175,17 +244,25 @@ export interface SearchV2Options {
   sortBy?: SortByV2
 }
 
+/**
+ * File object returned by the List V2 API (listV2() method)
+ * Note: Folder entries will have null values for most fields except key and name
+ */
 export interface SearchV2Object {
```
itslenny, line 247: On List V2 objects and folders are returned separately and have separate types. Therefore, this type is never used for folders. See `SearchV2Folder` below. I don't think these values are ever null for Objects. So rather than making these values `| null` you could just define the separate Folder type to have this only.

itslenny (changes requested):

mandarini:
@itslenny addressed your comment

packages/core/storage-js/src/lib/types.ts
ferhatelmas: hmmm, I think issue is still there. user metadata isn't supposed to have size, eTag, mimetype, etc. Making it required is wrong

ferhatelmas: https://github.com/supabase/storage/blob/bb602aa99e44709609e46aa745298dd7b1dfa066/src/storage/renderer/info.ts#L29

packages/core/storage-js/src/lib/types.ts
ferhatelmas: similarly, this can exists and be null

https://github.com/supabase/storage/blob/master/src/storage/renderer/info.ts#L25-L27

PR #2125, opened 2026-02-19
fix(auth): lower lockAcquireTimeout default to 5s and fix stale JSDoc

mandarini:
## Description

Follow-up to #2106 (steal-based orphaned-lock recovery).

- **Lower `lockAcquireTimeout` default** from 10 s to 5 s (`GoTrueClient.ts`). The auth team aligned on a shorter default in the original Slack thread; now that the steal fallback handles recovery, a shorter wait is safe.
- **Fix stale `lock` JSDoc** (`types.ts`): the previous comment claimed "no locking is done by default", which is false — `navigatorLock` (Web Locks API) has been the default in browser environments with `persistSession: true` for
   some time.
- **Rewrite `lockAcquireTimeout` JSDoc** (`types.ts`): the previous doc said a positive timeout throws `LockAcquireTimeoutError`. With #2106 merged, a positive timeout now triggers steal-based recovery instead. Updated `@default`, bullet descriptions, and the example snippet accordingly.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -174,7 +174,7 @@ const DEFAULT_OPTIONS: Omit<
   debug: false,
   hasCustomAuthorizationHeader: false,
   throwOnError: false,
-  lockAcquireTimeout: 10000, // 10 seconds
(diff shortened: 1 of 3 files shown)


mandarini:
@fadymak timeout was there since before. we just decreased it to 5s, as you suggested!

PR #2129, opened 2026-02-22
fix(storage): pass metadata in uploadToSignedUrl

viveknathani:
## 🔍 Description

`uploadToSignedUrl` silently ignored the metadata property in `fileOptions`, so custom user metadata was never persisted when uploading via signed URLs.

### What changed?

The fix adds metadata handling to all three body type paths in `uploadToSignedUrl`, matching the behavior of `uploadOrUpdate` and also forwards `fileOptions.headers` to the request, consistent with `upload()`. Tests have been added too.

This was inconsistent with the standard `.upload()` method, which correctly processes and attaches metadata to the request body/headers.

Originally raised here: https://github.com/supabase/storage/issues/823

## 🔄 Breaking changes

## 📋 Checklist

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -248,22 +248,40 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
     return this.handleOperation(async () => {
       let body
       const options = { upsert: DEFAULT_FILE_OPTIONS.upsert, ...fileOptions }
-      const headers: Record<string, string> = {
+      let headers: Record<string, string> = {
         ...this.headers,
         ...{ 'x-upsert': String(options.upsert as boolean) },
       }
 
+      const metadata = options.metadata
+
       if (typeof Blob !== 'undefined' && fileBody instanceof Blob) {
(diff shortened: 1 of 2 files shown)


viveknathani:
hi @mandarini - just bumping this since it's been a couple of weeks. github already requested review from the codeowners but i haven’t heard back yet. would appreciate it if you could take a look when you get a chance, thanks!

mandarini:
Hi @viveknathani, thanks for taking the time to put this together, the fix is correct and the approach is exactly right. I do apologize for taking so long to review, there were some other priorities to look at, and this fell behind a bit. Thank you for your patience and understanding.
I'm going to close this in favor of #2155 which tackles the same issue but also covers two additional cases: stream bodies on Node 20+ (which require duplex: 'half' on the fetch call, same as uploadOrUpdate), and a pre-existing bug where missing cacheControl/contentType in fileOptions would send "undefined" strings to the server. Do not be discouraged, and I do really appreciate the contribution. Feel free to keep an eye out for other issues to tackle! Thank you for contributing to Supabase! 💚

viveknathani:
makes sense, thanks for the review!

PR #2130, opened 2026-02-23
fix(realtime): decode edge cases

yukimotochern:

Closes #(issue_number) 

## 🔄 Breaking changes

## 📋 Checklist


yukimotochern:
I typically use zod in my own project for validation but not sure the dependency policy here.

mandarini:
Hey @yukimotochern, thanks for the PR! Defensive error handling in the serializer is a good idea. A few things to address before we merge:                                                      
                                                                                                                                                                                                  
callback({}) causes a downstream crash: The main issue is that calling callback({}) on a parse error doesn't actually make things safer, it moves the crash downstream. In _onConnMessage, the decoded message is used like this:      

```ts
  const { topic, event, payload, ref } = msg
  const status = payload.status || '' // TypeError if payload is undefined
```

Since {} destructures to payload = undefined, you'd get a TypeError there instead. I'd suggest either skipping the callback entirely on bad input (just return after logging), or guarding _onConnMessage against payload being undefined.

Binary decode path has the same issue. _decodeUserBroadcast has an unguarded JSON.parse(decoder.decode(payload)) too, worth addressing for consistency while you're in here.

Use the client's logger instead of console.error. RealtimeClient has a log() method — using that keeps log handling consistent and respects any logging config the user has set.

 A couple of unit cases covering malformed JSON and a short array would go a long way here.


I have a question regarding the `vitest.config.ts` → `.mts` rename. Was there an actual problem prompting this? The rename itself might be fine (forcing ESM for the config in a CJS package), but since it's unexplained and unrelated to the fix, it'd be good to understand the motivation, or move it to a separate PR.

On Zod: Good instinct to flag it. We haven't fully decided on a validation library yet, so the manual checks are fine for now. If/when we adopt something like Zod across the codebase, this would be a natural place to update.

mandarini (changes requested):
See my comment

yukimotochern:
@mandarini fixed

mandarini (changes requested):
Hey @yukimotochern, thanks for the update, a couple of things still need to be addressed before this is ready:                                                                                 
   
  1. PR title: The title needs to follow the conventional commit format. Please change it from `Fix realtime-js: decode edge cases` to follow our guidelines.

  2. Use the client's logger, not console.error: All the new error paths still call console.error directly. As mentioned in my previous review, please use RealtimeClient's log() method instead. This keeps log handling consistent and respects any logging configuration the user has set.

  3. Tests are still missing: Only serializer.ts was changed. Please add unit cases covering at least: malformed JSON input and a valid JSON array with fewer than 5 elements.

7ttp:
> Hey @yukimotochern, thanks for the update, a couple of things still need to be addressed before this is ready:
> 
>     1. PR title: The title needs to follow the conventional commit format. Please change it from `Fix realtime-js: decode edge cases` to follow our guidelines.
> 
>     2. Use the client's logger, not console.error: All the new error paths still call console.error directly. As mentioned in my previous review, please use RealtimeClient's log() method instead. This keeps log handling consistent and respects any logging configuration the user has set.
> 
>     3. Tests are still missing: Only serializer.ts was changed. Please add unit cases covering at least: malformed JSON input and a valid JSON array with fewer than 5 elements.

saw this being pending for a while aswell :) 
should i supersedes this w credit to OP  @mandarini 

edit: he'll take this up himself

yukimotochern:
ah. I miss this one. I forgot this one. I will add test tomorrow.

7ttp:
sure 💚

yukimotochern:
done

mandarini (changes requested):
Hi @yukimotochern, I still see some issues with this PR, but thank you for the contribution! Can you please:

1. Fix the PR title to follow our commit message format as described here: https://github.com/supabase/supabase-js/blob/develop/CONTRIBUTING.md#commit-message-format
                                                                                                                                                                                                                      
2. Wire the`logger` through `RealtimeClient` so the new log parameter actually fires. In `packages/core/realtime-js/src/RealtimeClient.ts`, replace:

```ts
defaultDecode = this.serializer.decode.bind(this.serializer)
```
with:                                                                                                                                                                                                               

```ts
defaultDecode = (payload, callback) => this.serializer.decode(payload, callback, this.log.bind(this))
```

Without this, log is always the noop default and the new error logging is dead code.                                                                                                                                
                                                                                      
3. Add a binary-path test. All three new tests cover the string path. The binary path gained two new failure branches (JSON.parse on payload, JSON.parse on metadata) plus a _binaryDecode early-return for undefined. Add at least one test that builds a userBroadcast buffer with JSON_ENCODING and a corrupt JSON byte sequence, asserts the callback isn't invoked, and asserts log is called with 'Error decoding JSON payload'.

yukimotochern:
done

mandarini (changes requested):
The four new `log()` calls forward raw user-controlled frame contents (`rawPayload`, `jsonPayload`, `textPayload`, `metadata`) to the application logger. Realtime payloads can contain PII, so if a consumer ships logs to Sentry/Datadog/etc., malformed frames will quietly exfiltrate user data through the error channel.

Can you drop the payload values and log only `{ error }`? If you want some diagnostic context, a length or shape proxy is fine (e.g. `{ isArray, length }` for the bad-array case). Applies to all four log sites in this file.

Otherwise this looks good! Thanks for sticking with it.

yukimotochern:
fixed

PR #2139, opened 2026-02-26
fix(auth): fixes userattributes type

staaldraad:
## 🔍 Description

The UserAttributes type used currentPassword (camel-case) but the Auth server expects current_password (snake-case)

### What changed?

`UserAttribute` type

Bad typing would cause the Auth server to reject all password change attempts if current_password was required

## 📝 Additional notes

This feature isn't active yet on any Auth servers, but will be soon


fadymak (approved):
🧹 🧹 🧹

PR #2144, opened 2026-02-28
feat(auth): auth.resend() consistent confirmation flow

weilirs:
### What changed?

When `flowType` is `'pkce'`, `resend()` now generates a fresh `code_verifier`/`code_challenge` pair (via the existing `getCodeChallengeAndMethod` helper), stores the verifier, and includes `code_challenge` + `code_challenge_method` in the request body to `/resend`. This follows the exact same pattern already used by `signUp()` and `signInWithOtp()`. See [this PR](https://github.com/supabase/auth/pull/2401)

- **`packages/core/auth-js/src/GoTrueClient.ts`** — Added PKCE challenge generation in the email branch of `resend()`, included `code_challenge` and `code_challenge_method` in the request body, and added `code-verifier` cleanup on both HTTP-error returns and in the catch block.
- **`packages/core/auth-js/test/GoTrueClient.test.ts`** — Added tests for `resend()` with PKCE `flowType` covering both `type: 'signup'` and `type: 'email_change'`, plus assertions that `code_challenge` is included in the request body and that the code verifier is cleaned up on HTTP errors.

**Supported resend types with PKCE:** Both `signup` and `email_change` go through the `if ('email' in credentials)` branch and will include PKCE parameters when `flowType` is `'pkce'`. The auth server PR [supabase/auth#2401](https://github.com/supabase/auth/pull/2401) supports PKCE for both types.

No changes to the phone/SMS resend path, user-facing types, or implicit flow behavior.

The GoTrue server has been updated to accept `code_challenge` and `code_challenge_method` on the `/resend` endpoint. Previously, `resend()` always produced implicit-flow confirmation links (`#access_token=...`), even when the original `signUp()` used PKCE. This forced developers using server-side frameworks (e.g., Next.js App Router) to implement workarounds for handling hash fragments that aren't accessible in server route handlers.

With this change, resend confirmation emails use `?code=...` (PKCE) when the SDK is configured for PKCE flow, consistent with the initial signup email. This applies to both signup confirmation resends and email change confirmation resends.

Closes supabase/supabase#42527

## 📸 Screenshots/Examples

Before (implicit flow on resend — hash fragment not accessible server-side):
```
https://example.com/auth/confirm#access_token=xxx&refresh_token=yyy
```

After (PKCE flow on resend — query param works with server routes):
```
https://example.com/auth/confirm?code=xxx
```

## 🔄 Breaking changes

## 📋 Checklist

- [x] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `feat(auth-js): add PKCE support to resend()`

## 📝 Additional notes

- Each `resend()` call generates a **fresh** `code_verifier`/`code_challenge` pair — it does not reuse the pair from the original `signUp()`, since the user may be in a different browser session.
- The user-facing `ResendParams` type is unchanged; PKCE params are added internally by the SDK based on the configured `flowType`, matching how `signUp()` and `signInWithOtp()` already work.
- SMS-based resend types (`sms`, `phone_change`) are unaffected.
- On HTTP errors (e.g., 400 for invalid email), the code verifier is now cleaned up from storage immediately, consistent with how `signUp()` handles it — rather than leaving it orphaned until the next call.


mandarini:
Hi @weilirs ! Thank you for this PR! As you already wrote in the description, we need to wait for the auth PR to be merged/released first. While we wait, can you please make these improvements?

1. Add a test for type: 'email_change', the `if ('email' in credentials)` branch handles both signup and email_change. The auth server PR also supports PKCE for email_change. But the only test added is for type: 'signup'.

2. Can you pelase strengthen the test assertion? The current test only checks `error === null` against a mock. It doesn't verify that `getCodeChallengeAndMethod` was actually called or that `code_challenge` was included in the request. 

3. Can you also please clarify the error-path cleanup? In the email branch, if _request returns an HTTP error (e.g., 400 for wrong email), the code-verifier written by getCodeChallengeAndMethod is left in storage, and cleanup only runs in the catch block, not on HTTP-error returns. This is consistent with signInWithOtp (line 1257), but I think it would be worth it to either:
  - Adding a comment like `// code-verifier cleanup follows signInWithOtp pattern` cleaned up on next call, or
  - Making it consistent with signUp by adding `if (error) await removeItemAsync(...)` before the return

4. Update the PR description, since now only mentions signup. Since the implementation also covers email_change, update the description to say so explicitly. The auth server PR #2401 supports both, and both should be documented I believe.

Oh and finally can you please use conventional commits? I assumed you pushed with `--no-verify` 🫣 ? You can use `npm run commit` to help you!

Thank you so much for this contribution! :D Let's wait for the auth team now to check the other PR first!

mandarini:
Adding `do not merge` until https://github.com/supabase/auth/pull/2401 this PR is merged.

mandarini (changes requested):
Please read my comment above

weilirs:
Hi @mandarini, sorry for the late reply but I've made the changes you suggested. Thanks!

mandarini (changes requested):
Hi @weilirs, thank you so much for making those changes! 💚 

The tests are much stronger now, checking the actual request body for `code_challenge` and `code_challenge_method` is exactly what we needed, and the error-path cleanup test is solid. The `email_change` coverage is also a great addition.

A couple of small things before we can merge:

1. The `email_change` test doesn't verify that the code verifier was actually stored after the call. The `signup` test does this:
   ```ts
   const codeVerifier = await storage.getItem(`${storageKey}-code-verifier`)
   expect(codeVerifier).not.toBeNull()
   ```
   It would be good to add the same assertion to the `email_change` test for consistency.

2. There's no test for the implicit flow (non-PKCE) path. Since the body now always includes `code_challenge` and `code_challenge_method` (as `null` when not using PKCE), a test confirming that those fields are `null` in the body when `flowType` is not `'pkce'` would make the behavior explicit and protect against regressions.

We're still blocked on supabase/auth#2401 merging before we can ship this, so there's no rush, but it would be great to have these in while we wait.

Thank you again for sticking with this and addressing the feedback so thoroughly!

weilirs:
> Hi @weilirs, thank you so much for making those changes! 💚
> 
> The tests are much stronger now, checking the actual request body for `code_challenge` and `code_challenge_method` is exactly what we needed, and the error-path cleanup test is solid. The `email_change` coverage is also a great addition.
> 
> A couple of small things before we can merge:
> 
> 1. The `email_change` test doesn't verify that the code verifier was actually stored after the call. The `signup` test does this:
>    ```ts
>    const codeVerifier = await storage.getItem(`${storageKey}-code-verifier`)
>    expect(codeVerifier).not.toBeNull()
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    It would be good to add the same assertion to the `email_change` test for consistency.
> 2. There's no test for the implicit flow (non-PKCE) path. Since the body now always includes `code_challenge` and `code_challenge_method` (as `null` when not using PKCE), a test confirming that those fields are `null` in the body when `flowType` is not `'pkce'` would make the behavior explicit and protect against regressions.
> 
> We're still blocked on [supabase/auth#2401](https://github.com/supabase/auth/pull/2401) merging before we can ship this, so there's no rush, but it would be great to have these in while we wait.
> 
> Thank you again for sticking with this and addressing the feedback so thoroughly!

Hey @mandarini ! I strengthened the tests per your request, please take a look.

mandarini (approved):

mandarini:
@weilirs thanks. Let's wait for the server PR now to be merged!

mandarini:
Auth changes are in. We're waiting for the new release, then we merge

PR #2155, opened 2026-03-08
fix(storage): include metadata in uploadToSignedUrl request

MaxwellCalkin:
## Summary

Fixes https://github.com/supabase/storage/issues/823.

`uploadToSignedUrl` accepts `metadata` in `fileOptions` but the implementation never reads or appends it to the request. Files uploaded via signed URLs silently drop custom metadata, unlike `.upload()` which works correctly.

**Root cause:** The `uploadToSignedUrl` method constructs the request body without checking `options.metadata`, while the `uploadOrUpdate` helper (used by `.upload()` and `.update()`) includes metadata handling for all three body types.

**Fix:** Copy the metadata handling pattern from `uploadOrUpdate` into `uploadToSignedUrl`:
- **Blob body:** append `metadata` field to FormData
- **FormData body:** append `metadata` field (with `!body.has()` guard, matching `uploadOrUpdate`)
- **Raw body (string/buffer/stream):** set `x-metadata` base64 header

Also forwards `fileOptions.headers` to the request, matching `uploadOrUpdate` behavior.

## Test plan

- [x] Added 3 unit tests covering metadata in all body type branches (Blob, FormData, raw body)
- [x] All existing mocked unit tests continue to pass
- [x] Build passes (`nx build storage-js`)
- [ ] Integration tests require Docker (unchanged)

---

> **Transparency note:** This PR was authored by Claude Opus 4.6 (an AI), directed by a human supervisor. See [our project page](https://github.com/anthropics/anthropic-cookbook/blob/main/misc/ai_as_employee.md) for context.

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -267,7 +267,6 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       let body
       const options = {
         ...DEFAULT_FILE_OPTIONS,
-        upsert: DEFAULT_FILE_OPTIONS.upsert,
         ...fileOptions,
       }
       let headers: Record<string, string> = {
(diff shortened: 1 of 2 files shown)


packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -252,26 +252,61 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
 
     return this.handleOperation(async () => {
       let body
-      const options = { upsert: DEFAULT_FILE_OPTIONS.upsert, ...fileOptions }
-      const headers: Record<string, string> = {
+      const options = {
+        ...DEFAULT_FILE_OPTIONS,
+        upsert: DEFAULT_FILE_OPTIONS.upsert,
```
mandarini: I think we can remove this line as it's redundant.

Ajris: @mandarini @MaxwellCalkin Can I help somehow? If it's the only line that needs to be removed, I am happy to just do it to help push it forward, so that it unblocks many people :)

Ajris:
Thank you @MaxwellCalkin!

packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -265,26 +265,60 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
...
+          (body && typeof body === 'object' && 'pipe' in body && typeof body.pipe === 'function')
+
+        if (isStream && !options.duplex) {
+          options.duplex = 'half'
+        }
+      }
+
+      if (fileOptions?.headers) {
+        headers = { ...headers, ...fileOptions.headers }
       }
```
ferhatelmas, line 316: This isn't directly relevant to this PR since it's coming from existing code but combined with https://github.com/supabase/supabase-js/pull/2211#discussion_r3032226626 this can be a footgun because casing can be different

ferhatelmas:
I think this looks good as what it's trying to do but it's not complete since it doesn't pass metadata for RLS check in url creation. Right now, server doesn't bind metadata into token but it will then it will start failing. It could be added separately though.

PR #2165, opened 2026-03-12
docs(repo): enrich docs comment for remaining packages

jeremenichelli:
Move all the remaining docs from [supabase_js_v2.yml](https://github.com/supabase/supabase/blob/master/apps/docs/spec/supabase_js_v2.yml) to tsdoc.


mandarini:
Error is from dependabot update! Fixing it!

PR #2173, opened 2026-03-17
test(postgrest): verify hints get passed down

mandarini:
Adds a mock-based contract test that locks in postgrest-js' passthrough of the `hint` field on `PostgrestError`. The mock body matches what supautils >= 3.2.0 actually emits end-to-end (verified against a real Supabase project), so the test also serves as living documentation of the supautils 42501 response format.

Three cases covered: `{ data, error }` passthrough, `.throwOnError()` rejection carrying the same `hint`, and the `hint: null` case for when supautils isn't active for the role.

### DIFF
--- a/packages/core/postgrest-js/test/permission-hints.test.ts
+++ b/packages/core/postgrest-js/test/permission-hints.test.ts
@@ -1,0 +1,61 @@
+import { PostgrestClient } from '../src/index'
+import { Database } from './types.override'
+
+// Simulates the JSON body PostgREST returns when supautils injects a hint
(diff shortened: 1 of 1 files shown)


themavik (commented):
permission-hints.test.ts pins the supautils-shaped 42501 body and proves hint survives both res.error and throwOnError. mockFetch as any is fine for a jest double; you could tighten types later without changing behavior.

o-santi (approved):

grdsdev (approved):

PR #2175, opened 2026-03-17
fix(postgrest): add type safety for eq() and neq() column names

aayushbaluni:
Fixes supabase/supabase#43381, Adds type safety to eq() and neq() for column names.
                                                                                                                                                                                                                                         
Builds on the original approach from @aayushbaluni, but uses a conditional type on the column parameter instead of overloads. The overload pattern doesn't actually reject invalid literals, they silently fall through to the `eq(column: string, value: unknown)` fallback. The conditional type approach closes that hole.                                                                                                                                           
                                                                       
  How it works:                                                                                                                                                                                                                          
                                                                       
  | Call | Result |
  |------|--------|
  | `.eq('username', 'foo')` |  typed value |
  | `.eq('INVALID', 'x')` | TS error not a valid column |
  | `.eq('users.status', 'ONLINE')` | relationship filter |                                                                                                                                                                           
  | `.eq('bar->version', 31)` | JSON path filter |
  | `const col: string; .eq(col, x)` | dynamic variable passes through |                                                                                                                                                               
                                                                       
Also fixes a latent type error in `relationships-error-handling.test.ts` where `.eq('id', 2)` was called on the users table (which has no id column). This was previously masked by the fallback overload.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -93,13 +93,6 @@ export default class PostgrestFilterBuilder<
   Relationships,
   Method
 > {
-  eq<ColumnName extends string & keyof Row>(
-    column: ColumnName,
-    value: ResolveFilterValue<Schema, Row, ColumnName> extends infer Resolved
-      ? NonNullable<Resolved>
-      : never
-  ): this
-  eq(column: string, value: unknown): this
   /**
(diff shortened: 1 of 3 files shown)


themavik (commented):
eq/neq column parameters now reject bogus literals while still allowing dotted and -> paths, and index.test-d covers INVALID. nit: append uses column as string — needed for URLSearchParams, but it is the escape hatch where mistyped templates slip past keyof Row.

PR #2179, opened 2026-03-19
fix(storage): structural detection on json() to detect Response-like errors

mandarini:
## Description

Fixes a bug where storage operation errors (e.g., emptyBucket on a bucket with too many objects, or accessing a non-existent bucket) produced "{}" as the error message instead of the actual API error message.

## Root Cause

`handleError` in fetch.ts checks whether a caught value is a Response object using:

```ts
'status' in error && 'ok' in error && typeof error.status === 'number'
```

This check fails on Response objects from undici and certain fetch polyfills in production environments. When the check fails, the code falls through to:

```ts
new StorageUnknownError(_getErrorMessage(response), response, namespace)
```

`_getErrorMessage` tries several properties (msg, message, error_description, etc.) and falls back to JSON.stringify(response). Since Response objects have no enumerable own properties, JSON.stringify returns "{}" so callers receive a StorageUnknownError with message: "{}" instead of the actual error.

## Fix

Use structural detection on json() instead. It is present on the prototype of all Response implementations (native, node-fetch, cross-fetch, undici) and absent from standard Error objects:

```ts
// Before                                                                                                               
  const isResponseLike =
    error && typeof error === 'object' &&                                                                                 
    'status' in error && 'ok' in error &&
    typeof (error as any).status === 'number'                                                                             
                             
  // After
  const isResponseLike =
    error !== null && typeof error === 'object' &&
    typeof (error as any).json === 'function'
```

Also hardened status extraction to handle string status codes from non-standard implementations.

## Testing

- Added regression test reproducing the exact scenario: a Response-like object without an ok property now correctly rejects with StorageApiError and the actual error message, instead of StorageUnknownError with "{}"

### DIFF
--- a/packages/core/storage-js/src/lib/common/fetch.ts
+++ b/packages/core/storage-js/src/lib/common/fetch.ts
@@ -53,37 +53,22 @@ const handleError = async (
 
   if (isResponseLike) {
     const responseError = error as any
-    const status =
-      typeof responseError.status === 'number'
-        ? responseError.status
-        : parseInt(responseError.status, 10) || 500
-
-    // Try to parse JSON body if available
-    if (typeof responseError.json === 'function') {
-      responseError
-        .json()
-        .then((err: any) => {
-          const statusCode = err?.statusCode || err?.code || status + ''
-          reject(new StorageApiError(_getErrorMessage(err), status, statusCode, namespace))
-        })
-        .catch(() => {
-          // If JSON parsing fails for vectors, create ApiError with HTTP status
-          if (namespace === 'vectors') {
-            const statusCode = status + ''
-            const message = responseError.statusText || `HTTP ${status} error`
-            reject(new StorageApiError(message, status, statusCode, namespace))
-          } else {
-            const statusCode = status + ''
-            const message = responseError.statusText || `HTTP ${status} error`
-            reject(new StorageApiError(message, status, statusCode, namespace))
-          }
-        })
-    } else {
-      // No json() method available, create error from status
-      const statusCode = status + ''
-      const message = responseError.statusText || `HTTP ${status} error`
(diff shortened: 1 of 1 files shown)


packages/core/storage-js/src/lib/common/fetch.ts
```diff
@@ -45,18 +45,18 @@ const handleError = async (
...
-    typeof (error as any).status === 'number'
+    error !== null && typeof error === 'object' && typeof (error as any).json === 'function'
 
   if (isResponseLike) {
     const responseError = error as any
-    const status = responseError.status || 500
+    const status =
+      typeof responseError.status === 'number'
+        ? responseError.status
+        : parseInt(responseError.status, 10) || 500
```
kamilogorek:
```suggestion
    let status = parseInt(responseError.status, 10)
    if (!Number.isFinite(status)) {
      status = 500
    }
```

This is more defensive and will handle all malformed errors.

packages/core/storage-js/src/lib/common/fetch.ts
```diff
@@ -45,18 +45,18 @@ const handleError = async (
...
+  // Structural detection of json() method, present in all Response implementations
+  // (native, node-fetch, cross-fetch, undici) and absent from standard Error objects.
+  // Checking 'ok' or 'status' via `in` is unreliable across fetch polyfills/realms.
   const isResponseLike =
-    error &&
-    typeof error === 'object' &&
-    'status' in error &&
-    'ok' in error &&
-    typeof (error as any).status === 'number'
+    error !== null && typeof error === 'object' && typeof (error as any).json === 'function'
```
kamilogorek, line 52: This will make check on line 62 obsolete and prevent fallback from ever be executed in its current shape

Issue #1905, opened 2025-12-02
Add a type discriminator to result types


gordonbrander:
### Description

Adding a type discriminator such as an `ok: boolean` property to result types such as [`PostgrestResponseSuccess`](https://github.com/supabase/supabase-js/blob/8d1e0c52c4f3b15e174eddc883611cb5a41c49f0/packages/core/postgrest-js/src/types/types.ts#L15) would make them discriminated unions. This has a number of advantages. You can use `if (!res.ok) { ... }` to do type narrowing. It also lets the type checker ensure that success and error are mutually exclusive, and there is not some third state of `{ data: null, error: null }`.

### Suggested solution

For example, https://github.com/supabase/supabase-js/blob/8d1e0c52c4f3b15e174eddc883611cb5a41c49f0/packages/core/postgrest-js/src/types/types.ts#L15 could become:

```ts
export interface PostgrestResponseSuccess<T> extends PostgrestResponseBase {
  ok: true
  error: null
  data: T
  count: number | null
}
export interface PostgrestResponseFailure extends PostgrestResponseBase {
  ok: false
  error: PostgrestError
  data: null
  count: null
}

export type PostgrestSingleResponse<T> = PostgrestResponseSuccess<T> | PostgrestResponseFailure
```

### Additional context

Old Typescript handbook doc on discriminating unions https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions

### Validations

- [x] Follow our [Code of Conduct](https://github.com/supabase/.github/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md).
- [x] Read the [docs](https://supabase.com/docs/reference/javascript/introduction).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

nancysangani:
Hi! Can I take this issue? Just want to confirm the ok: boolean discriminator approach is still acceptable before I start.

gordonbrander:
A quick note: I was just digging through Zod's docs and noticed that Zod's [`.safeParse()`](https://github.com/colinhacks/zod/blob/main/packages/docs-v3/README.md#safeparse) result type uses the same field names as supabase-js results (`data` and `error`). It uses `success` as the name of the type discriminator field.

```ts
type Result<T, E> = {
  success: true;
  data: T;
} | {
  success: false;
  error: E
};
```

[...]

PR #2198, opened 2026-03-27
feat(postgrest): add success discriminator field to PostgREST response types

claygeo:
## Problem

`PostgrestResponseSuccess<T>` and `PostgrestResponseFailure` form a union type but lack a boolean discriminator field. Users must narrow via `if (res.error)` rather than a clean boolean check, and TypeScript doesn't guarantee exhaustive narrowing.

## Solution

Add a `success` boolean literal field as a discriminated union discriminator:

```typescript
const res = await supabase.from('users').select()
if (res.success) {
  res.data   // T (narrowed)
  res.error  // null (narrowed)
} else {
  res.error  // PostgrestError (narrowed)
  res.data   // null (narrowed)
}
```

The field name `success` aligns with Zod's `safeParse()` result type, which uses the same `data`/`error` pattern with a `success` boolean discriminator. This was the community consensus in #1905.

## Changes

| File | Change |
|------|--------|
| `packages/core/postgrest-js/src/types/types.ts` | Added `success: true` to Success, `success: false` to Failure |
| `packages/core/postgrest-js/src/PostgrestBuilder.ts` | Set `success` at runtime in both success and fetch error catch paths |

## Backwards Compatibility

Additive only. Existing code that narrows via `if (res.error)` continues to work unchanged. No existing fields are modified or removed.

Ref: #1905

### DIFF
--- a/packages/core/postgrest-js/test/advanced_rpc.test.ts
+++ b/packages/core/postgrest-js/test/advanced_rpc.test.ts
@@ -83,6 +83,7 @@ describe('advanced rpc', () => {
         "error": null,
         "status": 200,
         "statusText": "OK",
+        "success": true,
       }
     `)
     ExpectedSchema.parse(res.data)
@@ -138,6 +139,7 @@ describe('advanced rpc', () => {
         "error": null,
         "status": 200,
         "statusText": "OK",
+        "success": true,
       }
     `)
     ExpectedSchema.parse(res.data)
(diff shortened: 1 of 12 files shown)


steve-chavez (approved):

claygeo:
Good catch @mahdirajaee — updated `success: !error` to `success: error === null` for the explicit null check. The `!error` approach would silently break the discriminant if PostgREST ever returned a falsy-but-non-null error shape, whereas `=== null` directly matches the declared `PostgrestError | null` type and won't misbehave on edge cases.

PR #2200, opened 2026-03-30
fix(storage): support exactOptionalPropertyTypes

7ttp:
aligns `StorageFileApi` and `StorageError` field types with their existing constructor assignments so 
`storage-js` compiles with `exactOptionalPropertyTypes` enabled

also adds a small test in the existing `storageFileApi` suite

- closes https://github.com/supabase/supabase-js/issues/1664

### DIFF
--- a/packages/core/storage-js/tsconfig.lib.json
+++ b/packages/core/storage-js/tsconfig.lib.json
@@ -9,6 +9,7 @@
     "emitDeclarationOnly": false,
     "declaration": true,
     "declarationMap": true,
+    "exactOptionalPropertyTypes": true,
 
     "stripInternal": true,
     "allowSyntheticDefaultImports": true,


7ttp:
> Hi @7ttp, thank you so much for this PR! :D
> 
> This is a nice fix, `exactOptionalPropertyTypes` compatibility is one of those things that trips up library consumers and it's genuinely useful to have it sorted. The type changes themselves are exactly right: `status: number | undefined` is the correct way to express "always present, possibly undefined" under that flag, and the `StorageApiError` overrides hold up cleanly without needing any changes.
> 
> One suggestion before we merge: since you've already verified the whole package compiles cleanly with the flag, the right move is to just add `"exactOptionalPropertyTypes": true` directly to the package's `tsconfig.json`. That way the guarantee is explicit in the config, any future violation will be caught by the normal build, and the test added in this PR isn't needed at all, the build already acts as the type-check gate here.
> 
> Would you be open to making that swap? Happy to help if anything is unclear.

Aha, yh that def makes more sense :) 💚
 i’ve updated to enable `exactOptionalPropertyTypes`  in  `tsconfig.json`
removed the extra test, and reverified that it compiles cleanly

packages/core/storage-js/tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "target": "ES6",
 
     "strict": true,
+    "exactOptionalPropertyTypes": true,
```
mandarini, line 16: One last ask! Can you please add `exactOptionalPropertyTypes: true` to `tsconfig.lib.json` as well?

7ttp: > One last ask! Can you please add `exactOptionalPropertyTypes: true` to `tsconfig.lib.json` as well?

done :)

PR #2209, opened 2026-03-31
fix(storage): set correct content-type for uploads

7ttp:
Moved: https://github.com/supabase/supabase-js/pull/2211

PR #2211, opened 2026-04-02
 fix(storage): set correct content-type for uploads

### LINKED ISSUE #2207: Storage upload fails with 415 when Content-Type is set in global headers
## Description

When `Content-Type` is set in `global.headers` during client initialization, file uploads fail with a 415 error. The outgoing request ends up with a malformed `Content-Type` header that combines the global value with the upload's own content type:

```
Content-Type: application/json, image/png
```

This was previously tolerated by the Storage API but https://github.com/supabase/storage/pull/893 introduced strict content header checking with the fastify version update (see [changelog](https://github.com/fastify/fastify/releases/tag/v5.7.2) "Parsing of the content-type header has been improved to a strict parser").

## Expected Behavior

The upload-specific content type should be sent as the sole `Content-Type` header value.

## Actual Behavior

The outgoing request contains both values joined as a single header.

```
Content-Type: application/json, image/png
```

### Steps to reproduce

```ts

const SUPABASE_URL = 'https://project.supabase.co'
const SUPABASE_ANON_KEY = 'anon_key'
const BUCKET = 'bucket-name'

const supabase = createClient(SUPABASE_URL, SUPABASE_ANON_KEY, {
  global: {
    headers: {
    },
    fetch: (url, init) => {
      console.log('Headers being sent:', init?.headers)
      return fetch(url, init)
    },
  },
})

const fileBuffer = Buffer.from('fake image content')

const { data, error } = await supabase.storage
  .from(BUCKET)
  .upload(`test-${Date.now()}.png`, fileBuffer, {
    contentType: 'image/png',
    upsert: true,
  })

if (error) {
  console.error('Upload failed:', error.message)
  console.error('Status:', error.status, '/ statusCode:', error.statusCode)
} else {
  console.log('Upload succeeded:', data)
}
```

Result:
```
Headers being sent: Headers {
  'X-Client-Info': 'supabase-js-node/2.101.0',
  'Content-Type': 'application/json, image/png',
  'x-upsert': 'true',
  'cache-control': 'max-age=3600',
  apikey: 'REDACTED',
  Authorization: 'Bearer REDACTED'
}
Upload failed: Invalid Content-Type header
Status: 400 / statusCode: 415
```

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M4
    Memory: 371.64 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.19.6 - /Users/kostas/.nvm/versions/node/v20.19.6/bin/node
    npm: 10.8.2 - /Users/kostas/.nvm/versions/node/v20.19.6/bin/npm
    pnpm: 10.32.1 - /Users/kostas/.nvm/versions/node/v20.19.6/bin/pnpm
    Deno: 2.6.9 - /opt/homebrew/bin/deno
  Browsers:
    Brave Browser: 146.1.88.132
    Chrome: 146.0.7680.165
    Firefox: 149.0
    Safari: 18.5
  npmPackages:
    @supabase/supabase-js: ^2.101.0 => 2.101.0
```

### Used Package Manager

npm

### Logs

```
Headers being sent: Headers {
  'X-Client-Info': 'supabase-js-node/2.101.0',
  'Content-Type': 'application/json, image/png',
  'x-upsert': 'true',
  'cache-control': 'max-age=3600',
  apikey: 'REDACTED',
  Authorization: 'Bearer REDACTED'
}
Upload failed: Invalid Content-Type header
Status: 400 / statusCode: 415
```

### Validations

mandarini:
Moved from https://github.com/supabase/supabase-js/pull/2209

### What changed?

- In `BaseApiClient`, all header keys are now normalized to lowercase at initialization time, and `setHeader` also lowercases the key before storing
- Added 2 unit tests to `storageFileApi.test.ts` verifying that `upload` and `uploadToSignedUrl` correctly use the file-specific `contentType` over any globally configured `Content-Type` header

When users configure a global `Content-Type` header on the Supabase client (e.g. `'Content-Type': 'application/json'`), storage upload requests ended up with two conflicting header entries because the global header key and the upload-specific key had different casing. Since JavaScript object keys are case-sensitive, `Content-Type` and `content-type` were treated as distinct keys and both were sent in the request.

Lowercasing all header keys at storage time ensures that when the upload sets `content-type: image/png`, it overwrites the global `Content-Type: application/json` entry rather than sitting alongside it.

Closes #2207

## Screenshots/Examples

Before (broken):
```ts
  global: { headers: { 'Content-Type': 'application/json' } }
})

// Would send both Content-Type: application/json AND content-type: image/png
await supabase.storage.from('bucket').upload('path', file, { contentType: 'image/png' })
```

After (fixed): the upload correctly sends only `content-type: image/png`.

## Breaking changes

## Checklist

## Additional notes

The fix applies case normalization at the storage client layer (`BaseApiClient`), so any header key variant (`Content-Type`, `content-type`, `CONTENT-TYPE`) is stored uniformly as lowercase. This is consistent with how HTTP headers are defined to be case-insensitive.

### DIFF
--- a/packages/core/storage-js/src/lib/common/BaseApiClient.ts
+++ b/packages/core/storage-js/src/lib/common/BaseApiClient.ts
@@ -30,7 +30,7 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
     namespace: ErrorNamespace = 'storage'
   ) {
     this.url = url
-    this.headers = headers
+    this.headers = Object.fromEntries(Object.entries(headers).map(([k, v]) => [k.toLowerCase(), v]))
     this.fetch = resolveFetch(fetch)
     this.namespace = namespace
   }
@@ -55,7 +55,7 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
    * @returns this - For method chaining
    */
   public setHeader(name: string, value: string): this {
-    this.headers = { ...this.headers, [name]: value }
+    this.headers = { ...this.headers, [name.toLowerCase()]: value }
     return this
   }
 
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -112,7 +112,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       } else {
         body = fileBody
         headers['cache-control'] = `max-age=${options.cacheControl}`
-        headers = this.setRequestHeader(headers, 'Content-Type', options.contentType as string)
+        headers['content-type'] = options.contentType as string
 
         if (metadata) {
           headers['x-metadata'] = this.toBase64(this.encodeMetadata(metadata))
@@ -266,7 +266,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
     return this.handleOperation(async () => {
       let body
       const options = { ...DEFAULT_FILE_OPTIONS, ...fileOptions }
-      let headers: Record<string, string> = {
+      const headers: Record<string, string> = {
         ...this.headers,
         ...{ 'x-upsert': String(options.upsert as boolean) },
       }
@@ -281,26 +281,13 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       } else {
         body = fileBody
         headers['cache-control'] = `max-age=${options.cacheControl}`
-        headers = this.setRequestHeader(headers, 'Content-Type', options.contentType as string)
+        headers['content-type'] = options.contentType as string
       }
 
       const data = await put(this.fetch, url.toString(), body as object, { headers })
 
       return { path: cleanPath, fullPath: data.Key }
     })
-  }
-
-  private setRequestHeader(headers: Record<string, string>, name: string, value: string) {
-    const nextHeaders = { ...headers }
-
-    for (const key of Object.keys(nextHeaders)) {
-      if (key.toLowerCase() === name.toLowerCase()) {
-        delete nextHeaders[key]
-      }
-    }
-
-    nextHeaders[name] = value
-    return nextHeaders
   }
 
   /**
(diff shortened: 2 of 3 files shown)


mandarini:
@7ttp can you please rebase with `master`, for some reason it's not letting me

packages/core/storage-js/src/lib/common/BaseApiClient.ts
```diff
@@ -30,7 +30,7 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
     namespace: ErrorNamespace = 'storage'
   ) {
     this.url = url
-    this.headers = headers
+    this.headers = Object.fromEntries(Object.entries(headers).map(([k, v]) => [k.toLowerCase(), v]))
```
ferhatelmas, line 33: I think this will regress vector client because constructor sets content-type to lower case but isPlainObject checks adds upper case content-type (they are different keys in the object). Fetch will normalize headers and will send two values for it but content-type is supposed to be a single value

mandarini: @7ttp can you please take a look here?

7ttp: > I think this will regress vector client because constructor sets content-type to lower case but isPlainObject checks adds upper case content-type they are different keys in the object Fetch will normalize headers and will send two values for it but content-type is supposed to be a single value

def makes sense 

> @7ttp can you please take a look here

i think we can keep this one as is and instead make the header set and merge points case insensitive for single value headers like `Content-Type`

that should kinda preserve this while avoiding the regress\ path where both `content-type` and `Content-Type` can exist on the same headers object before `fetch` normalizes them.....



wdyt/?
(if this somewhat makes sense i can get a small follow up pr  for it)

mandarini: yes can you please create a follow up PR? @7ttp

mandarini: @ferhatelmas here's the PR: https://github.com/supabase/supabase-js/pull/2220

PR #2219, opened 2026-04-05
fix(storage): apply empty transform check to download and getPublicUrl

### LINKED ISSUE #2218: Empty transform object routes to wrong endpoint in download() and getPublicUrl()
PR #2162 fixed `createSignedUrl()` to properly handle empty transform objects, but the same bug is still present in `download()` and `getPublicUrl()`.

Both methods use `typeof options?.transform !== 'undefined'` to decide whether to route through `/render/image/`. Since `typeof {} !== 'undefined'` is `true`, passing `{ transform: {} }` incorrectly routes to the image render endpoint, which returns 422 for non-image files.

`createSignedUrl` now correctly checks:
```ts
typeof options?.transform === 'object' && options.transform !== null && Object.keys(options.transform).length > 0
```

`download` and `getPublicUrl` should use the same pattern.

oniani1:
Fixes #2218

PR #2162 fixed `createSignedUrl()` to reject empty transform objects, but `download()` and `getPublicUrl()` still use the old `typeof options?.transform !== 'undefined'` check. Since `typeof {} !== 'undefined'` is `true`, passing an empty transform object routes requests through `/render/image/` instead of `/object/`, causing 422 errors for non-image files.

This applies the same `Object.keys` check from `createSignedUrl` to both methods.

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -817,7 +817,10 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
     options?: Options,
     parameters?: FetchParameters
   ): BlobDownloadBuilder {
-    const wantsTransformation = typeof options?.transform !== 'undefined'
+    const wantsTransformation =
+      typeof options?.transform === 'object' &&
+      options.transform !== null &&
+      Object.keys(options.transform).length > 0
     const renderPath = wantsTransformation ? 'render/image/authenticated' : 'object'
     const transformationQuery = this.transformOptsToQueryString(options?.transform || {})
     const queryString = transformationQuery ? `?${transformationQuery}` : ''
@@ -1006,7 +1009,10 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       _queryString.push(downloadQueryParam)
     }
 
-    const wantsTransformation = typeof options?.transform !== 'undefined'
+    const wantsTransformation =
+      typeof options?.transform === 'object' &&
+      options.transform !== null &&
+      Object.keys(options.transform).length > 0
     const renderPath = wantsTransformation ? 'render/image' : 'object'
     const transformationQuery = this.transformOptsToQueryString(options?.transform || {})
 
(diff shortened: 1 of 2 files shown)


mandarini (changes requested):
Hi @oniani1, thank you so much for contributing to Supabase! 💚 

This is a clean, correct fix. The `Object.keys` guard is exactly the right pattern, and it's consistent with what was done in `createSignedUrl` in #2162. The diff is minimal and easy to verify.

One small thing before I can merge: could you add two tests alongside this fix, similar to the one added in #2162 for `createSignedUrl`? The existing test at line 124 of `storageFileApi.test.ts` is a good template -- we'd just need the same shape for `download()` and `getPublicUrl()`. Something like:

- `download() with empty transform object does not use render endpoint`
- `getPublicUrl() with empty transform object does not use render endpoint`

Once those are in, this is good to merge. Thanks again for the thorough write-up in the issue and for following through with the fix!

mandarini (changes requested):
Sorry, one more change @oniani1 ! The `getPublicUrl` test looks great. The `download` test needs one small adjustment. Because the test uploads `sadcat.jpg` (a real JPEG), hitting `/render/image/authenticated/` with an empty query string would still return the image successfully, so `res.error` being null does not actually prove the correct endpoint was used. Could you swap the uploaded file for the `file.txt` fixture that's already in the test fixtures? A non-image file will 422 if it accidentally routes through the render endpoint, making `res.error === null` a meaningful assertion:                                  
   
  ```ts                                                                                                                                                                                                                                                                  
  it('download with empty transform object does not use render endpoint', async () => {
    const txtFile = await fsp.readFile(uploadFilePath('file.txt'))
    const txtPath = `testpath/file-${Date.now()}.txt`                                                                                                                                                                                                                    
    await storage.from(bucketName).upload(txtPath, txtFile)
                                                                                                                                                                                                                                                                         
    const res = await storage.from(bucketName).download(txtPath, {                                                                                                                                                                                                       
      transform: {},
    })                                                                                                                                                                                                                                                                   
                                                         
    expect(res.error).toBeNull()
    expect(res.data?.size).toBeGreaterThan(0)
  })
  ```
  Once that's updated, this is good to merge. Thanks for the quick turnaround!

oniani1:
Hey @mandarini, the file.txt swap is in f1398683 -- should be good to go now. Let me know if anything else needs changing!

mandarini (approved):

PR #2220, opened 2026-04-06
fix(storage): avoid duplicate content-type headers in vector requests

7ttp:
## TL;DR

fixes a regression from #2211 where vector json requests could end up with both `Content-Type` and `content-type`

## Problem

that PR lowercased stored headers in `BaseApiClient`

plain object json requests in the shared fetch layer still added `Content-Type`
 without checking for an existing lowercase `content-type`

that could leave both on the same request

## Solution

make the shared fetch layer handle `Content-Type` case insensitively for json requests

## Related

follow up to
- #2211 which closed
-  https://github.com/supabase/supabase-js/issues/2207

### DIFF
--- a/packages/core/storage-js/src/lib/common/fetch.ts
+++ b/packages/core/storage-js/src/lib/common/fetch.ts
@@ -95,7 +95,16 @@ const _getRequestParams = (
   }
 
   if (isPlainObject(body)) {
-    params.headers = { 'Content-Type': 'application/json', ...options?.headers }
+    const headers = options?.headers || {}
+    let contentType: string | undefined
+
+    for (const [key, value] of Object.entries(headers)) {
+      if (key.toLowerCase() === 'content-type') {
+        contentType = value
+      }
+    }
+
+    params.headers = setRequestHeader(headers, 'Content-Type', contentType ?? 'application/json')
     params.body = JSON.stringify(body)
   } else {
     params.body = body
@@ -106,6 +115,19 @@ const _getRequestParams = (
   }
 
   return { ...params, ...parameters }
+}
+
+function setRequestHeader(headers: Record<string, string>, name: string, value: string) {
+  const nextHeaders = { ...headers }
+
+  for (const key of Object.keys(nextHeaders)) {
+    if (key.toLowerCase() === name.toLowerCase()) {
+      delete nextHeaders[key]
+    }
+  }
+
+  nextHeaders[name] = value
+  return nextHeaders
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mandarini (approved):

mandarini:
waiting for @ferhatelmas 's input before we merge

packages/core/storage-js/test/bucket-api.spec.ts
```diff
@@ -266,3 +267,29 @@ describe('VectorBucketApi Integration Tests', () => {
...
+    const mockFetch = jest.fn().mockResolvedValue(
+      new Response(JSON.stringify({}), {
+        status: 200,
+        headers: { 'content-type': 'application/json' },
+      })
+    )
+
+    const client = new StorageVectorsClient('https://mock.example.com', {
+      fetch: mockFetch as any,
+      headers: { 'Content-Type': 'application/json' },
```
ferhatelmas: passing lowercase would make it a better test

packages/core/storage-js/src/lib/common/fetch.ts
```diff
@@ -95,7 +95,16 @@ const _getRequestParams = (
   }
 
   if (isPlainObject(body)) {
-    params.headers = { 'Content-Type': 'application/json', ...options?.headers }
+    const headers = options?.headers || {}
+    let contentType: string | undefined
+
+    for (const [key, value] of Object.entries(headers)) {
+      if (key.toLowerCase() === 'content-type') {
```
ferhatelmas, line 102: I think this fixes the issue here but good to note that header duplication can bite in general where a single value header semantic is desired. It would be nice to have a common place to normalize and merge headers.

ferhatelmas (dismissed):

packages/core/storage-js/test/bucket-api.spec.ts
mandarini: @7ttp can you please address this?

packages/core/storage-js/src/lib/common/fetch.ts
mandarini: Good point! I think we can address this in a separate PR, create sort of a shared header normalization utility.

mandarini (changes requested):
just the comment from @ferhatelmas about the test case!

packages/core/storage-js/test/bucket-api.spec.ts
7ttp: donee💚

mandarini (approved):

packages/core/storage-js/src/lib/common/fetch.ts
mandarini: @ferhatelmas : https://github.com/supabase/supabase-js/pull/2251

ferhatelmas: thank you 😍

PR #2228, opened 2026-04-07
fix(auth,supabase): recover from client corruption after tab suspension

### LINKED ISSUE #36046: Supabase Client Becomes Irrecoverably Corrupted After Tab Suspension (PWA/Browser)
Supabase Client Becomes Irrecoverably Corrupted After Tab Suspension (PWA/Browser)

### Summary

After a few minutes of browser tab suspension or mobile app backgrounding (especially in PWAs or on Safari/iOS), the Supabase client enters a **corrupted internal state**. Once this happens:

- `functions.invoke()` hangs silently or never resolves
- `auth.getSession()` returns `null` despite a valid session
- `auth.signOut()` no longer completes
- Realtime channels fail silently and don’t reconnect
- The only reliable recovery is to **fully recreate the Supabase client via `createClient()`**

---

###  What Works Initially

- Supabase client initializes correctly
- Auth/session is valid
- Realtime channels and function calls operate as expected

---

###  What Fails After Idle

After 2–5 minutes of backgrounding:

- `functions.invoke()` stalls without resolving
- `auth.getSession()` returns `null`, even though a session exists
- Realtime `.on('close')` fires, but `.on('open')` never follows
- `realtime.accessTokenValue` becomes out-of-sync with `auth.getSession()`
- `auth.getUser()` may return a user, but API calls return 401s

---

### Root Cause

Supabase client internally shares state across:
- Auth session and token
- Realtime WebSocket
- Function call promise queue

When the browser suspends or backgrounds the tab/app:
- The WebSocket may silently disconnect
- Session token state becomes desynchronized
- Promises issued via `.functions.invoke()` may freeze

Because the client is monolithic, this internal corruption affects all areas of usage (auth, functions, realtime).

---

### Current Workaround

We’ve implemented a wrapper that:
- Detects provable corruption based on 5 deterministic checks:
  1. Internal promise hang (`functions.invoke()` hangs, raw `fetch()` works)
  2. Session desync (`auth.getSession()` returns null while session exists)
  3. Realtime token mismatch (`realtime.accessTokenValue` is stale)
  4. WebSocket zombie state (`.on('close')` fires but never `.on('open')`)
  5. False-positive authentication (`getUser()` returns a user, but requests 401)
- Replaces the client with a new one using `createClient()`
- Rehydrates session
- Re-subscribes to Realtime channels

---

### Repro Steps

1. Use a PWA or leave an app open for 2-3 min and move to a different tab
2. Authenticate a user and open a Realtime channel
3. Background the tab for 3–5 minutes (or switch apps on mobile)
4. Resume the tab/app
5. Attempt to call `auth.getSession()` or `functions.invoke()`
6. Observe: client fails to complete call or responds incorrectly

---

### Suggested Fixes

- Add public API for health:
  - `supabaseClient.isCorrupted()`
  - `supabaseClient.reinitialize()`
- Internally self-heal:
  - Auto-sync tokens between `.auth` and `.realtime`
  - Auto-reconnect WebSocket with fallback validation
  - Retry `.functions.invoke()` if internal execution stalls
- Provide official guidance for client recreation

---

### Impact

This issue affects:
- PWAs and mobile apps (especially on Safari and iOS)
- Long-lived browser sessions
- Background-tab use cases
- Any app requiring real-time + session persistence without full reload

---

### Environment

- Browser(s): Chrome, Safari, Mobile Safari
- App type: [SPA / PWA / React / Svelte / etc.]

---

KrunchMuffin:
## Summary

Fixes the Supabase client becoming irrecoverably corrupted after browser tab suspension (2-5+ minutes), which affects PWAs, mobile browsers, and long-lived sessions. After resuming:

- `auth.getSession()` could return `null` despite a valid session in storage
- `functions.invoke()` could hang indefinitely (blocked on lock-gated `getSession`)
- Realtime channels failed silently with stale tokens and no reconnection

**Root cause:** When `_recoverAndRefresh` fails with a retryable error (e.g. network still reconnecting after tab resume), no auth event was emitted, leaving realtime and functions permanently desynchronized. The next auto-refresh tick could be up to 30 seconds away.

### Changes

**auth-js (`GoTrueClient`):**
- Emit `SIGNED_IN` with current session on retryable refresh failure so downstream listeners (realtime, functions) can re-sync immediately instead of silently dropping the error
- Trigger an immediate `_autoRefreshTokenTick()` after visibility change recovery instead of waiting up to `AUTO_REFRESH_TICK_DURATION_MS` (30s) for the next scheduled tick

**supabase-js (`SupabaseClient`):**
- Add `visibilitychange` listener that force-syncs a fresh token to the realtime client when the tab resumes — acts as a safety net for cases where the auth event bridge misses the token update

## Test plan

- [x] New unit test: `should sync fresh token to realtime on visibility change (tab resume)` in SupabaseClient tests
- [x] All existing realtime-js tests pass (325/325, only pre-existing worker test failures on non-browser env)
- [x] All existing supabase-js unit tests pass (88/88)
- [ ] Manual testing: open app with realtime channel, background tab for 3-5 min, resume and verify session + realtime recover

## Related

Closes supabase/supabase#36046

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -4540,6 +4540,17 @@ export default class GoTrueClient {
                 error
               )
               await this._removeSession()
+            } else {
+              // The refresh failed with a retryable error (e.g. network still
+              // reconnecting after tab suspension). Emit SIGNED_IN with the
+              // current session so downstream listeners (realtime, functions)
+              // can re-sync with the existing token. The auto-refresh ticker
+              // will retry the actual refresh on its next tick.
+              this._debug(
+                debugName,
+                'refresh failed with a retryable error, notifying subscribers with current session'
+              )
+              await this._notifyAllSubscribers('SIGNED_IN', currentSession)
             }
           }
         }
@@ -5029,6 +5040,14 @@ export default class GoTrueClient {
           // recover the session
           await this._recoverAndRefresh()
         })
+
+        // After lock-based recovery, if the session is still expired
+        // (e.g. network was not yet available when _recoverAndRefresh ran),
+        // trigger an immediate auto-refresh tick so we don't wait up to
+        // AUTO_REFRESH_TICK_DURATION_MS (30s) for the next scheduled tick.
+        if (this.autoRefreshToken) {
+          await this._autoRefreshTokenTick()
+        }
       }
     } else if (document.visibilityState === 'hidden') {
       if (this.autoRefreshToken) {
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -355,6 +355,7 @@ export default class SupabaseClient<
 
     if (!settings.accessToken) {
       this._listenForAuthEvents()
+      this._listenForVisibilityChange()
     }
   }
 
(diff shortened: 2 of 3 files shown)


gfsaaser24:
This looks solid. Nice work dude.

mandarini:
The reason why I do not want to merge this PR (or similar PRs) is that they are patching a deeper issue with how we use the Web Lock API and navigator locks. We have discussed this internally with the team quite a lot. While we appreciate all your contributions, we do want to come up with a solution to the root of the problem.
Would you like to try out `2.103.1-beta.0`? It's a release for this PR https://github.com/supabase/supabase-js/pull/2235.

mandarini:
[This](https://github.com/supabase/supabase-js/pull/2392) has been merged. I am closing this PR. Thank you for your contribution!

PR #2230, opened 2026-04-07
fix(postgrest): serialize bigint values in rpc() request body

### LINKED ISSUE #2117: support bigint parameters
### Description

When doing rpc calls with number parameters, `bigint` values need to be stringified, so you'd write `{amount: bigintAmount.toString()}`.

This is unintuitive, and the types also don't allow this.

### Suggested solution

Do the JSON.stringify with a visitor that converts bigint to string, and change the generated type for number values to `numbe | bigint`

### Alternative

keep the status quo

### Additional context

_No response_

### Validations

Vansh1811:
### What changed?

Added a `bigint` serialization replacer to `JSON.stringify` in `PostgrestBuilder.ts`. When the request body is serialized, any `bigint` value is now converted to its string representation instead of throwing a `TypeError`.

**One-line change:**
```diff
- body: JSON.stringify(this.body),
+ body: JSON.stringify(this.body, (_, v) => typeof v === 'bigint' ? v.toString() : v),
```

JavaScript's `JSON.stringify` does not natively support `bigint` values and throws:
```
TypeError: Do not know how to serialize a BigInt
```

This means any user calling `rpc()` with a `bigint` argument (e.g., for `int8` / `bigint` Postgres columns) had to manually call `.toString()` on every bigint value:
```ts
// Before (broken):
await supabase.rpc('my_func', { amount: 9007199254740993n }) // throws TypeError

// Workaround users needed:
await supabase.rpc('my_func', { amount: 9007199254740993n.toString() })

// After (works transparently):
await supabase.rpc('my_func', { amount: 9007199254740993n }) // works!
```

This approach is consistent with how many serialization libraries handle `bigint` and avoids a breaking change — `bigint` → string conversion matches what PostgREST expects for `int8` parameters anyway.

Closes #2117

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -248,7 +248,7 @@ export default abstract class PostgrestBuilder<
           res = await _fetch(this.url.toString(), {
             method: this.method,
             headers: requestHeaders,
-            body: JSON.stringify(this.body),
+            body: JSON.stringify(this.body, (_, v) => typeof v === 'bigint' ? v.toString() : v),
             signal: this.signal,
           })
         } catch (fetchError: any) {


mandarini (changes requested):
Hi @Vansh1811, thank you so much for contributing to Supabase! 💚 

This is a real, well-reported issue (#2117) and the core approach is correct, converting `bigint` to string is the standard workaround for `JSON.stringify`'s limitation, and Postgres will coerce a JSON string to `int8` just fine inside `json_to_record`.

Before we can merge this, we need at least one test. There's an `rpc.test.ts` in `packages/core/postgrest-js/test/` that would be the right place for this. At minimum, a unit test that mocks fetch and asserts the serialized body contains the string representation of the bigint. Ideally, an integration test against an actual Postgres function with an `int8` parameter -- you can add the function to `test/supabase/migrations/00000000000000_schema.sql`.

There's also one edge case worth thinking through: if a user passes a bigint value inside a `jsonb` parameter (e.g., `{ payload: { id: 9999n } }` for an RPC that takes a `jsonb`), the bigint becomes a JSON string rather than a JSON number. That may or may not be the right behavior, but it should be either tested or called out in a comment so the tradeoff is explicit.

Could you add the tests and take a look at the `jsonb` case? Happy to help if you have questions about the test setup.

Thank you so much for contributing to Supabase, it's exactly this kind of fix that makes the SDK more reliable for real-world use cases.

7ttp:
> Hi @Vansh1811, thank you so much for contributing to Supabase! 💚
> 
> This is a real, well-reported issue (#2117) and the core approach is correct, converting `bigint` to string is the standard workaround for `JSON.stringify`'s limitation, and Postgres will coerce a JSON string to `int8` just fine inside `json_to_record`.
> 
> Before we can merge this, we need at least one test. There's an `rpc.test.ts` in `packages/core/postgrest-js/test/` that would be the right place for this. At minimum, a unit test that mocks fetch and asserts the serialized body contains the string representation of the bigint. Ideally, an integration test against an actual Postgres function with an `int8` parameter -- you can add the function to `test/supabase/migrations/00000000000000_schema.sql`.
> 
> There's also one edge case worth thinking through: if a user passes a bigint value inside a `jsonb` parameter (e.g., `{ payload: { id: 9999n } }` for an RPC that takes a `jsonb`), the bigint becomes a JSON string rather than a JSON number. That may or may not be the right behavior, but it should be either tested or called out in a comment so the tradeoff is explicit.
> 
> Could you add the tests and take a look at the `jsonb` case? Happy to help if you have questions about the test setup.
> 
> Thank you so much for contributing to Supabase, it's exactly this kind of fix that makes the SDK more reliable for real-world use cases.

ah, noticed this has been also waiting for a bit now ~(7-10d)
would it be fine if I open a new PR to address these suggestions 
while also giving credit to the OP there? :) 💚 @mandarini

mandarini:
sure @7ttp go ahead!

7ttp:
> sure @7ttp go ahead!

here you go: 
- https://github.com/supabase/supabase-js/pull/2245

mandarini:
Closing in favor of https://github.com/supabase/supabase-js/pull/2245

Issue #2111, opened 2026-02-05
@supabase/ssr browser client auth methods hang indefinitely due to orphaned Web Locks


WillTaylor22:
# `@supabase/ssr` browser client auth methods hang indefinitely due to orphaned Web Locks (potentially React Strict Mode compatibility issue)

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Description

(N.b. - written with AI but it's done a far more comprehensive job than I would have - hope that's OK)

The `@supabase/ssr` browser client's auth methods (`getUser()`, `signInWithPassword()`, etc.) hang indefinitely without making any network requests when a Web Lock from a previous auth operation is not properly released.

The Supabase auth client uses the browser's Web Locks API (`navigator.locks`) to prevent concurrent auth operations. When a lock is acquired but not released (e.g., due to React Strict Mode unmounting a component mid-operation, or an abort signal firing), all subsequent auth calls queue behind the orphaned lock and hang forever.

The error `AbortError: signal is aborted without reason` appears in the console from `locks.ts`, indicating the lock acquisition was aborted, but the lock remains held.

## To Reproduce

1. Create a Next.js 14+ app with `@supabase/ssr` (React Strict Mode is enabled by default)
2. Create a client component that calls `supabase.auth.getUser()` in a `useEffect`:

```tsx
"use client";
import { useEffect } from "react";
import { createClient } from "@/lib/supabase/client"; // uses createBrowserClient

export default function SettingsPage() {
  const supabase = createClient();

  useEffect(() => {
    const fetchUser = async () => {
      const {
        data: { user },
      } = await supabase.auth.getUser(); // Hangs here
      console.log(user);
    };
    fetchUser();
  }, [supabase]);

  return <div>Settings</div>;
}
```

3. Navigate to the page - it may work initially
4. Trigger React Strict Mode's unmount/remount cycle (happens automatically in dev) or navigate away while auth is in progress
5. Subsequent auth calls hang indefinitely

**Verification:** Run in browser console:

```javascript
navigator.locks.query().then(console.log);
```

You'll see:

```javascript
{
  held: [{ name: "lock:sb-xxx-auth-token", mode: "exclusive" }],
  pending: [{ name: "lock:sb-xxx-auth-token", mode: "exclusive" }]
}
```

[...]

Issue #936, opened 2024-01-02
Acquiring an exclusive Navigator LockManager failed


supersu-man:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Angular throws an error saying 'Acquiring an exclusive Navigator LockManager lock immediately failed'.
This is a new angular project created with the angular version 17.
The error is not shown in my other project which is created a month ago.

## To Reproduce

1. Create a new angular project and install '@supabase/supabase-js'
2. Create environment files and add key and url of supabase.
3. add `supabaseClient = createClient(environment.url, environment.key)` to a component (app component)

## Expected behavior

This error should not be thrown and should be able to use all the methods without errors.

## Screenshots

![Error](https://github.com/supabase/supabase-js/assets/35676632/edd0fd65-dac6-4df2-8752-7987a367a4ee)

## System information

- OS: Windows
- Browser: Brave v1.61.109
- Version of supabase-js: 2.39.2
- Version of Node.js: 20.10.0

## Additional context

Could be a duplicate of [this issue](https://github.com/supabase/gotrue-js/issues/830) from gotrue-js.
Though I was able to signin using provider 'google' I wasn't able to retrieve session after I signin (throws an error saying Invalid API key) though my api keys and url are correct. I am assuming the error is with storage.

Klunk75:
Hi there,
I noticed the same problem.
After some test I found that the issue should be related to the merge of this Pull Request in @supabase/gotrue-js version 2.62.0:
https://github.com/supabase/gotrue-js/pull/807
If you try to downgrade @supabase/gotrue-js to version 2.61.0 (the previous one), the problem should disappear.
Please Supabase Team can you check and try to solve this issue?
Thank you!

Paolo

supersu-man:
> If you try to downgrade @supabase/gotrue-js to version 2.61.0 (the previous one), the problem should disappear. 

Thank you @Klunk75 for the workaround.

For anyone who wants to fix it temporarily, add this below devDependencies in package.json

```
"overrides": {
    "@supabase/supabase-js": {
      "@supabase/gotrue-js": "2.61.0"
    }
  }
```
and run `npm i`

hf:
Is this error just being logged, or is it actually creating a problem for you?

[...]

PR #2235, opened 2026-04-08
fix(auth): replace navigatorLock with processLock as default browser lock

mandarini:
### What changed?

Replaced `navigatorLock` (Web Locks API) with `processLock` (in-process promise queue) as the default browser lock in `GoTrueClient`. Previously, any browser with `navigator.locks` support would use the Web Locks API for cross-tab auth coordination. Now all environments use `processLock` by default, which serializes operations within a single tab/process without touching the Web Locks API.

`navigatorLock` remains exported and available as an explicit opt-in:

```ts
import { navigatorLock } from '@supabase/auth-js'

createClient(url, key, { auth: { lock: navigatorLock } })
```

`navigator.locks` causes permanent authentication deadlocks in production when a lock is acquired but never released. Once a lock is orphaned, every subsequent auth call in that tab hangs indefinitely until the user closes the browser. Known triggers:

- React Strict Mode double-mount/unmount interrupting a lock request mid-flight (affects all Next.js 13+ apps in dev mode by default)
- HMR tearing down components while a lock is held
- `localStorage.setItem()` hanging due to quota exceeded or browser extension interference
- Android Chrome not releasing locks from previous sessions

The steal-based recovery added in #2106 and #2178 addressed the symptom but introduced new failure modes (double-write windows, cascade when stolen). This PR addresses the root cause by removing the Web Locks API from the default path entirely.

Cross-tab concurrent refresh races are now handled server-side by GoTrue's refresh token reuse detection. When two tabs refresh simultaneously, the server allows both via the `likelyNotSavedByClient` (`counterDifference == 1`) and `likelyConcurrentRefreshes` (within reuse interval) logic, returning valid tokens to both tabs without any client-side coordination. `refresh_token_already_used` is only returned for genuinely stale tokens outside the reuse window, at which point the session is already server-side revoked and the existing sign-out behavior is correct.

`processLock` still provides within-tab serialization, which is needed to prevent races between operations in the same tab (e.g. `signOut` racing with an in-flight auto-refresh).

Relates to: #2111, #936, #1594, #2013

## Screenshots/Examples

N/A

## Breaking changes

The default lock behavior changes, but no public API is removed or modified. `navigatorLock` remains exported. `processLock` was already exported. The `lock` and `lockAcquireTimeout` options continue to work as before.

## Checklist

## Additional notes

The server-side behavior this relies on (`likelyNotSavedByClient` counter logic) is confirmed in `supabase/auth` `internal/tokens/service.go`. The reuse detection has been in production since the v2 refresh token algorithm rollout (Oct 2025).

## Try it out

You can try out this PR on `2.103.1-beta.0`

### DIFF
--- a/packages/core/auth-js/example/react/src/App.tsx
+++ b/packages/core/auth-js/example/react/src/App.tsx
@@ -13,6 +13,10 @@ const auth = new AuthClient({
   },
 })
 
+if (import.meta.env.DEV) {
+  ;(window as any).__auth__ = auth
+}
+
 function App() {
   const [session, setSession] = useState<Session | null>(null)
   const [email, setEmail] = useState(localStorage.getItem('email') ?? '')
--- a/packages/core/auth-js/example/react/tests/lock-behavior.spec.ts
+++ b/packages/core/auth-js/example/react/tests/lock-behavior.spec.ts
@@ -1,0 +1,123 @@
+import { test, expect } from '@playwright/test'
+
+const STORAGE_KEY = 'supabase.auth.token'
+const DEADLOCK_TIMEOUT_MS = 5000
+
+/** Expire the stored session so the next getSession() triggers a token refresh. */
+async function expireStoredSession(page: any) {
+  await page.evaluate((key: string) => {
+    const raw = localStorage.getItem(key)
+    if (!raw) return
+    const session = JSON.parse(raw)
+    session.expires_at = Math.floor(Date.now() / 1000) - 10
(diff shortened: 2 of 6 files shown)


mandarini:
We are most probably NOT going to follow through with this solution. Bear with us, we are actively working on this and should have something concrete by end of May.

mandarini:
[This](https://github.com/supabase/supabase-js/pull/2392) has been merged. I am closing this PR.

PR #2238, opened 2026-04-09
fix(auth): add toJSON to AuthError for correct JSON serialization

oniani1:
## Summary

`JSON.stringify()` silently drops the `message` field on most auth error classes because `Error.prototype.message` is non-enumerable.

Only 2 out of 13 auth error classes (`AuthImplicitGrantRedirectError` and `AuthPKCEGrantCodeExchangeError`) had `toJSON()` overrides. The rest -- including `AuthApiError`, which is by far the most commonly encountered -- would serialize without `message`:

```js
const err = new AuthApiError('Invalid login credentials', 400, 'invalid_credentials')
JSON.stringify(err)
// Before: {"status":400,"code":"invalid_credentials"}  -- message is gone
// After:  {"name":"AuthApiError","message":"Invalid login credentials","status":400,"code":"invalid_credentials"}
```

This adds `toJSON()` to the `AuthError` base class (returning `name`, `message`, `status`, `code`), so all subclasses inherit it automatically. The two existing overrides now spread `super.toJSON()` instead of duplicating fields. `AuthWeakPasswordError` gets an override to include `reasons`.

Same pattern as the recent fixes for `FunctionsError` (#2226) and `PostgrestError` (#2212).

## Test plan

- [x] Added `errors.test.ts` covering all 12 concrete error classes
- [x] Each test verifies `JSON.stringify()` round-trip preserves `message` and all relevant fields
- [ ] Existing tests pass with no regressions

### DIFF
--- a/packages/core/auth-js/src/lib/errors.ts
+++ b/packages/core/auth-js/src/lib/errors.ts
@@ -31,6 +31,15 @@ export class AuthError extends Error {
     this.status = status
     this.code = code
   }
+
+  toJSON() {
+    return {
+      name: this.name,
+      message: this.message,
+      status: this.status,
+      code: this.code,
+    }
+  }
 }
 
 export function isAuthError(error: unknown): error is AuthError {
@@ -181,9 +190,7 @@ export class AuthImplicitGrantRedirectError extends CustomAuthError {
 
   toJSON() {
     return {
-      name: this.name,
-      message: this.message,
-      status: this.status,
+      ...super.toJSON(),
       details: this.details,
     }
   }
@@ -215,9 +222,7 @@ export class AuthPKCEGrantCodeExchangeError extends CustomAuthError {
 
   toJSON() {
     return {
-      name: this.name,
-      message: this.message,
-      status: this.status,
+      ...super.toJSON(),
       details: this.details,
     }
   }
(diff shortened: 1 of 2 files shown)


mandarini (commented):
Hi @oniani1, thank you so much for contributing to Supabase! 💚 

This is a really nice fix. The base-class `toJSON()` approach is exactly the right call, and it lines up perfectly with what we did for `FunctionsError` (#2226) and `PostgrestError` (#2212). The tests are thorough and the `super.toJSON()` spread in the existing overrides is a clean de-duplication.

Two small things I wanted to flag before merging:

1. **`AuthUnknownError.originalError`**: with the base `toJSON()`, this property won't appear in serialized output. I suspect that's fine (the original error could be circular or otherwise non-serializable), but wanted to confirm that's intentional on your end. If it is, could you add a quick test assertion that `originalError` is *not* present in the serialized output? That way nobody tries to "fix" it later without understanding the reasoning.

2. **`code` now appears in `AuthImplicitGrantRedirectError` and `AuthPKCEGrantCodeExchangeError` toJSON output**: since both pass `undefined` for `code`, this is invisible through `JSON.stringify` (undefined gets dropped). But if someone calls `.toJSON()` directly, they'll now see `code: undefined` in the object. Not a blocker at all, just wanted to make sure you'd considered it.

Once you've confirmed the `originalError` point, this is good to go.

Thank you so much for contributing to Supabase!

oniani1:
Yeah, the `originalError` exclusion is intentional. It's typed as `unknown` and could be circular or carry non-serializable properties, so including it in `toJSON()` would risk `JSON.stringify` throwing. Anyone who needs it can still grab `err.originalError` directly.

Added a test for that in 7af2c876.

On the `code: undefined` thing, agreed that's just how the `super.toJSON()` pattern works. `JSON.stringify` drops undefined values anyway so the serialized output isn't affected.

mandarini (approved):

PR #2240, opened 2026-04-10
fix(postgrest)!: wrap error in PostgrestError instance on processResponse

### LINKED ISSUE #1643: Type of error (PostgrestError) inside PostgrestResponseFailure is incorrect
# Bug report

## Describe the bug

Originally, I was trying to figure out why passing in the error object to `Sentry.captureException` would cause Sentry to treat the error object as a POJO, and not an `Error`. The error response from postgrest-js was typed as `PostgrestError`, which clearly extends `Error`.

It seems that through a couple of semi-related changes (#502, supabase/postgrest-js#555, supabase/postgrest-js#562), the type `PostgrestError` has started having more meaning than it signed up for.

As far as I can tell, this is the issue:
- If `throwOnError()` is used, then the error thrown by the client is a proper `PostgrestError`, which calls the super constructor, and would pass the check `error instanceOf Error`.
- If `throwOnError()` is NOT used, then the error has the same _shape_ as a `PostgrestError`, but the constructor is not used, which means it never calls the super constructor, and would not pass the check `error instanceOfError`.

The obvious solution would be to either separate the "thrown" error type and the "not-thrown" error type.

Vansh1811:
## Problem

The `error` field returned by `processResponse` in `PostgrestBuilder` is a plain object literal, not an instance of `PostgrestError`. `error instanceof PostgrestError` always returns `false`, breaking error type-narrowing for consumers.

Related issue: https://github.com/supabase/supabase-js/issues/1643

## Fix

Wrap the non-null `error` in `new PostgrestError(error)` in the final return of `processResponse`, matching the existing `throw new PostgrestError(error)` on the `shouldThrowOnError` path.

```ts
// Before
error,

// After
error: error ? new PostgrestError(error) : null,
```

Consumers can now reliably narrow:

```ts
const { error } = await supabase.from('table').select()
if (error instanceof PostgrestError) {
  console.log(error.code, error.message)
}
```

## Breaking change (v3)

The runtime shape of `error` changes from plain `Object` to a `PostgrestError`/`Error` instance:

- `error instanceof Error` flips `false` → `true`.
- `JSON.stringify(error)` routes through `toJSON()` and emits a new `name: "PostgrestError"` field.
- `Object.keys(error)` now includes `name` (and `message`, see below).
- The error now carries a `stack`.

Consumers forwarding errors to logs / UIs / wire formats may see the new shape.

## Other changes

- **`PostgrestError` constructor**: explicitly marks `message` as enumerable. `Error`'s super constructor leaves it non-enumerable, which would otherwise hide it from spread, `pretty-format`, and `JSON.stringify` paths that don't go through `toJSON()`. Now consistent with `code` / `details` / `hint`.
- **Jest snapshot serializer** (`test/postgrestErrorSerializer.js`): renders `PostgrestError` instances as `{ code, details, hint, message }` so existing inline snapshots stay valid after the prototype flip. Only matches `PostgrestError` instances — no effect on other values.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -533,7 +533,7 @@ export default abstract class PostgrestBuilder<
 
     return {
       success: error === null,
-      error,
+      error: error ? new PostgrestError(error) : null,
       data,
       count,
       status,
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -446,7 +446,7 @@ export default abstract class PostgrestBuilder<
    */
   private async processResponse(res: Response): Promise<{
     success: boolean
-    error: any
+    error: PostgrestError | null
     data: any
     count: number | null
     status: number
--- a/packages/core/postgrest-js/jest.config.ts
+++ b/packages/core/postgrest-js/jest.config.ts
@@ -19,6 +19,6 @@ const config: Config.InitialOptions = {
       tsconfig: 'tsconfig.test.json',
     },
   },
-    snapshotSerializers: ['<rootDir>/test/postgrestErrorSerializer.js'],
+  snapshotSerializers: ['<rootDir>/test/postgrestErrorSerializer.js'],
 }
 export default config
--- a/packages/core/postgrest-js/jest.config.ts
+++ b/packages/core/postgrest-js/jest.config.ts
@@ -19,5 +19,6 @@ const config: Config.InitialOptions = {
       tsconfig: 'tsconfig.test.json',
     },
   },
+    snapshotSerializers: ['<rootDir>/test/postgrestErrorSerializer.js'],
 }
 export default config
--- a/packages/core/postgrest-js/src/PostgrestError.ts
+++ b/packages/core/postgrest-js/src/PostgrestError.ts
@@ -23,6 +23,7 @@ export default class PostgrestError extends Error {
    */
   constructor(context: { message: string; details: string; hint: string; code: string }) {
     super(context.message)
+        Object.defineProperty(this, 'message', { value: context.message, enumerable: true, writable: true, configurable: true })
     this.name = 'PostgrestError'
     this.details = context.details
     this.hint = context.hint
--- a/packages/core/postgrest-js/src/PostgrestError.ts
+++ b/packages/core/postgrest-js/src/PostgrestError.ts
@@ -23,7 +23,12 @@ export default class PostgrestError extends Error {
    */
   constructor(context: { message: string; details: string; hint: string; code: string }) {
     super(context.message)
-        Object.defineProperty(this, 'message', { value: context.message, enumerable: true, writable: true, configurable: true })
+    Object.defineProperty(this, 'message', {
+      value: context.message,
+      enumerable: true,
+      writable: true,
+      configurable: true,
+    })
     this.name = 'PostgrestError'
     this.details = context.details
     this.hint = context.hint
--- a/packages/core/postgrest-js/test/basic.test.ts
+++ b/packages/core/postgrest-js/test/basic.test.ts
@@ -937,13 +937,13 @@ test('throwOnError throws errors instead of returning them', async () => {
     // @ts-expect-error Argument of type '"missing_table"' is not assignable to parameter
     await postgrest.from('missing_table').select().throwOnError()
   } catch (error) {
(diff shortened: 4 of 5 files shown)


steve-chavez:
> Related issue: supabase/postgrest-js#1643

I'm unable to find the reference so don't understand the motivation here, please share the full link.

mandarini (changes requested):
@Vansh1811 can you please address @steve-chavez question?

Vansh1811:
@steve-chavez Here is the full link to the referenced issue: https://github.com/supabase/supabase-js/issues/1643 — it describes how `error instanceof PostgrestError` always returns `false` because `processResponse` returns a plain object instead of a `PostgrestError` instance.

steve-chavez:
@Vansh1811 Cool, thanks for clarifying

@mandarini Will let you review this one, not sure if the change is breaking

Vansh1811:
Thanks for the earlier feedback. I’ve addressed Steve’s question and linked the original issue (#1643) so the motivation is clear.

The goal of this change is to make the non‑throwing error path consistent with the existing shouldThrowOnError path, which already returns a PostgrestError instance, so consumers can reliably use instanceof PostgrestError in both cases.

The implementation wraps the existing error payload in PostgrestError without changing the error fields themselves, so the serialized/JSON shape remains the same while the prototype becomes the expected PostgrestError instance.

If you feel that prototype change should be treated as breaking, I’m happy to adjust the approach or retarget this to an appropriate release.

Could you please take another look when you have a moment?

mandarini (changes requested):
Hey @Vansh1811, thanks for this PR, appreciate you linking the original issue too!!

After reviewing, I think this is a breaking change and will need to go into v3 (next major). Here's why:
                                                                                                                                                                                                                                         
* Wrapping the error in `PostgrestError` (which extends Error) changes observable behavior for consumers:                                                                                                                                  
                                                                                                                                                                                                                                         
  1. `{ ...error }` loses `message`: Error sets message as a non-enumerable property, so spreading a `PostgrestError` silently drops it. Before this change, spreading the plain object preserved all fields.                                 
  2. `JSON.stringify(error)` adds name: the `toJSON()` method on `PostgrestError` includes `name: "PostgrestError"`, which wasn't present on the plain object.
  3. `error instanceof Error` is now `true`: could change control flow in consumer error handlers.                                                                                                                                          
  4. `error.stack` is now present: Error instances carry a stack trace.                                                                                                                                                                   
                                                                                                                                                                                                                                         
I will keep this PR and land it as part of v3. A few things to address before then:                                                                                                                                              
                             
  - Please update the inline snapshots in the test files to match the new PostgrestError format.
  - Make message enumerable in the `PostgrestError` constructor so that `{ ...error }` still works for consumers. Add this line after the `super()` call:
  ```ts
  Object.defineProperty(this, 'message', { value: context.message, enumerable: true })                                                                                                                                                   
  ``` 
This minimizes the breaking surface since `spread`/`Object.keys` will include message just like before.

No rush on the updates! Thanks again for the contribution!

Vansh1811:
Hey @mandarini, thanks for the detailed review! I've addressed the feedback in the latest commit:

- Added `Object.defineProperty(this, 'message', { value: context.message, enumerable: true, writable: true, configurable: true })` in the `PostgrestError` constructor to make `message` enumerable

This ensures `message` shows up correctly in `JSON.stringify`, spread operations, and that `error instanceof Error` still works as expected. Waiting on CI to confirm the test snapshots pass. Let me know if there's anything else to fix!

mandarini:
@Vansh1811 i think you did not update the snapshot?

Screenshot 2026-04-22 at 10 21 53 AM

packages/core/postgrest-js/src/PostgrestBuilder.ts
o-santi: How about changing the return type to `PostgrestError` so that no type narrowing is needed? This problem wouldn't have happened if the return type wasn't typed as `any` in the first place.

mandarini: Good point. `any` is what let this slip through since ts had nothing to check. That being said, it's deeper than just the internal `type: PostgrestError`'s constructor declares `details: string; hint: string`, but PostgREST emits details/hint as `JSON | null` (Maybe JSON.Value in PostgREST/Error.hs). So even with error: `PostgrestError | null` here, the constructor accepts a shape that doesn't match the wire format. I'll tighten the type, but fixing the constructor signature to `details?: unknown; hint?: unknown` would be separate fix, which I'd like @steve-chavez to weigh in too

o-santi (approved):

mandarini (approved):

PR #2251, opened 2026-04-15
feat(storage): extract shared header normalization utility

mandarini:
### What changed?

Extracted a reusable `setHeader()` utility in `storage-js` that performs case-insensitive header deduplication. This consolidates header normalization logic that was previously scattered across multiple files:

- **New file: `headers.ts`** - Exports `setHeader(headers, name, value)` which removes any existing header matching the name (case-insensitive) before setting the new value. Returns a new object without mutating the input.
- **`fetch.ts`** - Removed the private `setRequestHeader` function and replaced it with the shared `setHeader` import.
- **`BaseApiClient.ts`** - The `setHeader` method now delegates to the utility instead of inline lowercasing.
- **`StorageFileApi.ts`** - User-provided `fileOptions.headers` are now merged via `setHeader` in a loop instead of a raw object spread, preventing case-variant duplicates (e.g. a user passing `Content-Type` when `content-type` already exists).

PR #2211 lowercased all headers in the `BaseApiClient` constructor. PR #2220 fixed the resulting `Content-Type` / `content-type` duplicate in `_getRequestParams` with inline logic.

The utility is scoped to `storage-js` for now since other packages (`postgrest-js`, `supabase-js`) already use the native `Headers` API which handles case-insensitivity natively.

## Breaking changes

## Checklist

## Additional notes

- The `setHeader` utility is intentionally minimal: one function, one file, no class. It can be promoted to a shared package later if other SDK packages need it.
- Existing tests in `fetch.test.ts` and `BaseApiClient.test.ts` continue to pass, covering the content-type deduplication behavior from PR #2220.

### DIFF
--- a/packages/core/storage-js/src/lib/common/BaseApiClient.ts
+++ b/packages/core/storage-js/src/lib/common/BaseApiClient.ts
@@ -1,5 +1,6 @@
 import { ErrorNamespace, isStorageError, StorageError } from './errors'
 import { Fetch } from './fetch'
+import { setHeader as setHeaderUtil } from './headers'
 import { resolveFetch } from './helpers'
 
 /**
@@ -55,7 +56,7 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
    * @returns this - For method chaining
    */
   public setHeader(name: string, value: string): this {
-    this.headers = { ...this.headers, [name.toLowerCase()]: value }
+    this.headers = setHeaderUtil(this.headers, name, value)
     return this
   }
 
--- a/packages/core/storage-js/src/lib/common/BaseApiClient.ts
+++ b/packages/core/storage-js/src/lib/common/BaseApiClient.ts
@@ -1,6 +1,6 @@
 import { ErrorNamespace, isStorageError, StorageError } from './errors'
 import { Fetch } from './fetch'
-import { setHeader as setHeaderUtil } from './headers'
+import { normalizeHeaders, setHeader as setHeaderUtil } from './headers'
 import { resolveFetch } from './helpers'
 
 /**
@@ -31,7 +31,7 @@ export default abstract class BaseApiClient<TError extends StorageError = Storag
     namespace: ErrorNamespace = 'storage'
   ) {
     this.url = url
-    this.headers = Object.fromEntries(Object.entries(headers).map(([k, v]) => [k.toLowerCase(), v]))
+    this.headers = normalizeHeaders(headers)
     this.fetch = resolveFetch(fetch)
     this.namespace = namespace
   }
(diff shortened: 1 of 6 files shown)


packages/core/storage-js/src/lib/common/BaseApiClient.ts
ferhatelmas: here as well

packages/core/storage-js/test/common/headers.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    // Simulate duplicate keys by building with Object.assign
+    Object.assign(headers, {
+      'content-type': 'text/plain',
+      'Content-Type': 'text/html',
+    })
+
+    const result = setHeader(headers, 'CONTENT-TYPE', 'application/json')
+    expect(Object.keys(result).filter((k) => k.toLowerCase() === 'content-type')).toHaveLength(1)
+    expect(result['CONTENT-TYPE']).toBe('application/json')
+  })
```
ferhatelmas: shall we add a test to ensure deduplication for not overridden header?

same setup, but set `content-length` for example and ensure one content-type is set in the return value

packages/core/storage-js/test/common/headers.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    Object.assign(headers, {
+      'content-type': 'text/plain',
+      'Content-Type': 'text/html',
+    })
+
+    const result = setHeader(headers, 'Content-Length', '42')
+    expect(result['Content-Length']).toBe('42')
+    // The unrelated content-type duplicates are untouched
+    const contentTypeKeys = Object.keys(result).filter((k) => k.toLowerCase() === 'content-type')
+    expect(contentTypeKeys.length).toBeGreaterThanOrEqual(1)
```
ferhatelmas: why not exactly 2?

ferhatelmas: maybe match object directly, better for documentation purposes

packages/core/storage-js/src/lib/common/headers.ts
```diff
@@ -0,0 +1,43 @@
...
+  const result = { ...headers }
+  const nameLower = name.toLowerCase()
+
+  for (const key of Object.keys(result)) {
+    if (key.toLowerCase() === nameLower) {
+      delete result[key]
+    }
+  }
+
+  result[name] = value
```
ferhatelmas: hmmm we keep the case given by user so following can still fail because `contentType` is directly mutated in `uploadOrUpdate` 

```js
const api = storage.from('folder')
api.setHeader('Content-Type', 'application/json')
await api.upload('key', 'body', { contentType: 'image/png' })
```

ferhatelmas (approved):

PR #2264, opened 2026-04-21
fix(postgrest): narrow column types after not(column, is, null)

### LINKED ISSUE #1360: Inferred types for nullable columns after not null checks should be non-nullable

mandarini:
When filtering with `.not(col, 'is', null)`, the resulting row type now marks that column as non-nullable in both the Row (for filter chaining) and Result (for the returned data type).                                                                                                                                                                            
                                                                                                                                                                                                                      
Closes #1360 

Thanks to @DevanshTyagi04 and https://github.com/supabase/supabase-js/pull/2028 for the motivation to address the issue.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -1812,7 +1812,11 @@ export default class PostgrestFilterBuilder<
    *
    * ```
    */
-  not(column: string, operator: string, value: unknown): this {
+  not(
+    column: string,
+    operator: string,
+    value: unknown
+  ): PostgrestFilterBuilder<ClientOptions, Schema, any, any, RelationName, Relationships, Method> {
     this.url.searchParams.append(column, `not.${operator}.${value}`)
     return this as any
   }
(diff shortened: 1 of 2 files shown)


steve-chavez (approved):
LGTM

PR #2266, opened 2026-04-22
fix(postgrest): restore runtime test files to tstyche scope

mandarini:
PR #2193 removed `./test/**/*.test.ts` from `testFileMatch` to fix CI failures with TypeScript 5.9. The actual root cause was missing type restrictions in `tsconfig.tstyche.json`, not the test files themselves.                                                                                                                                                                  
                                                                                                                                                                                                                                         
- Restore `./test/**/*.test.ts` to testFileMatch (intentionally included per PR #1777 review for cross-version type assertion validation)                                                                                                                                                                     
- Add `types: ["jest", "node"]` to avoid auto-discovering all 57 `@types`                                                                                                                                                               
- Add `skipLibCheck: true` for TS 4.7 compat with modern `@types`                                                                                                                                                                          
- Bump target to ES2020 for BigInt literal support in `rpc.test.ts`

### DIFF
--- a/packages/core/postgrest-js/tsconfig.tstyche.json
+++ b/packages/core/postgrest-js/tsconfig.tstyche.json
@@ -7,7 +7,7 @@
     "outDir": "dist/cjs",
     "rootDir": ".",
     "sourceMap": true,
-    "target": "ES2017",
+    "target": "ES2020",
     "lib": ["ES2022"],
 
     "strict": true,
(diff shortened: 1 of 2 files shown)


grdsdev (approved):

mrazauskas:
@mandarini Thanks for quick reaction!

To be honest, the `3.0.0-next.0` release brought me to the repo. I was curious if support for older TypeScript versions is dropped. Perhaps worth considering? Although there might be reasons to keep the support.

mandarini:
@mrazauskas we have not thought about this. We will think about it.

PR #2268, opened 2026-04-22
fix(misc): reduce any usage across packages

mandarini:
Replace ~50 explicit `any` annotations with `unknown` and proper type narrowing in internal code. Highest-value change: auth-js xform/fetch chain now takes `unknown` at the JSON boundary, preventing bugs like the signedUrl typo (#2254) from silently passing type checks.

- catch blocks: `e: any` → `e` with instanceof narrowing
- _getErrorMessage: `any` → `unknown` in auth-js and storage-js
- xform functions: `data: any` → `data: unknown` (internal only)
- websocket-factory: typed globalThis casts, renamed constructor field
- auth-js eslint: enable no-explicit-any as warn

No breaking changes, all modifications are internal.

## Why `unknown` remains in some places

Some `unknown`s here are deliberate. They fall into a few categories where tightening would either be incorrect or carry a real cost:

- **`catch (e)`** — TypeScript enforces this. With `useUnknownInCatchVariables` (on under `strict: true`), `catch (e: Error)` is a compile error (TS1196). JavaScript allows `throw <any value>` — including strings, numbers, and `null` from third-party code we call into — so the catch binding genuinely cannot be narrower than `unknown`. Narrowing happens inside the block.

- **`errors: unknown[]` collected from third-party callbacks** (e.g. `_notifyAllSubscribers`) — values come from user-supplied `onAuthStateChange` callbacks, which can throw anything. Downstream uses are only `console.error(...)` and `throw`, both of which accept any value. Tightening to `Error[]` would either lie about the type or require wrapping non-Errors, which loses fidelity for consumers throwing custom error classes.

- **Runtime-detection globals** (`WebSocketPair?: unknown`, `EdgeRuntime?: unknown` in `websocket-factory.ts`) — used only for existence checks (`typeof !== 'undefined'`). `EdgeRuntime` in particular isn't standardized across edge runtimes (different shapes in Vercel Edge, Next.js Edge, polyfills). Typing it would be claiming knowledge that doesn't hold cross-runtime; `unknown` enforces existence-check-only access.

## Why some `any`s are still in the call chain

- **`JSON.parse` and `Response.json()` return `any`** — these are typed `any` in TypeScript's standard library, not by us. To eliminate `any` end-to-end here we'd need a runtime schema validator (zod / valibot / hand-rolled guards) at every response boundary. That's a non-trivial bundle/perf cost for an SDK shipped to bundle-sensitive runtimes (RN, edge, browsers), so the SDK leans on localized type assertions at the boundary instead.

- **User-extensible payloads** — `user_metadata`, `app_metadata`, OAuth provider passthrough fields, PostgREST row shapes, Realtime broadcast payloads. The shape is defined by the consumer's application or by external services, not by Supabase. Where applicable these are typed via generics or `Record<string, unknown>`; the remaining `[key: string]: any` index signatures (e.g. on `Session`, `User`) preserve ergonomics for fields the server may add (`token_type`, `provider_token`, etc.) without forcing every caller to narrow.

## What this PR does not do

- Does not introduce runtime validation of wire data (zod or similar). Listed above as the path to remove the remaining `any` from `JSON.parse`/`Response.json()`, but out of scope here.
- Does not tighten `[key: string]: any` index signatures on `User`/`Session` to `unknown` — would force narrowing on every read of provider passthrough fields. Deliberate trade-off, not an oversight.

### DIFF
--- a/packages/core/auth-js/src/lib/fetch.ts
+++ b/packages/core/auth-js/src/lib/fetch.ts
@@ -9,6 +9,7 @@ import {
   GenerateLinkResponse,
   User,
   UserResponse,
+  WeakPassword,
 } from './types'
 import {
   AuthApiError,
@@ -19,6 +20,30 @@ import {
 } from './errors'
 
 export type Fetch = typeof fetch
+
+/** Raw session data from GoTrue server response. */
+interface GoTrueSessionData {
+  access_token?: string
+  refresh_token?: string
+  expires_in?: number
+  expires_at?: number
+  user?: User
+  [key: string]: any // server returns additional fields (token_type, provider_token, etc.) copied into Session
+}
+
+/** Raw session data that includes weak password info (password sign-in endpoints). */
+interface GoTrueSessionPasswordData extends GoTrueSessionData {
+  weak_password?: WeakPassword
+}
+
+/** Raw user data — either `{ user: User }` or the User object itself. */
+interface GoTrueUserData {
+  user?: User
+  [key: string]: any // data may BE the User directly (fallback path)
+}
+
+/** Raw generate-link data — link properties + User fields flattened into one object. */
+type GoTrueGenerateLinkData = GenerateLinkProperties & Record<string, any>
 
 export interface FetchOptions {
   headers?: {
@@ -139,7 +164,7 @@ interface GotrueRequestOptions extends FetchOptions {
   /**
    * Function that transforms api response from gotrue into a desirable / standardised format
    */
-  xform?: (data: unknown) => any
+  xform?: (data: any) => any
 }
 
 export async function _request(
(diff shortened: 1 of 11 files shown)


packages/core/auth-js/src/lib/fetch.ts
```diff
@@ -204,55 +213,59 @@ async function _handleRequest(
...
     return await result.json()
-  } catch (e: any) {
+  } catch (e) {
     await handleError(e)
   }
 }
 
-export function _sessionResponse(data: any): AuthResponse {
+export function _sessionResponse(data: unknown): AuthResponse {
+  const d = data as Record<string, any>
```
o-santi: Why use `any` and then immediatly annotate it as `Record<string, any>`? Just put in the function parameter type.

packages/core/auth-js/src/lib/fetch.ts
```diff
@@ -204,55 +213,59 @@ async function _handleRequest(
...
   }
 
-  const user: User = data.user ?? (data as User)
+  const user: User = d.user ?? (d as User)
   return { data: { session, user }, error: null }
 }
 
-export function _sessionResponsePassword(data: any): AuthResponsePassword {
+export function _sessionResponsePassword(data: unknown): AuthResponsePassword {
+  const d = data as Record<string, any>
```
o-santi: Ditto as last comment.

packages/core/auth-js/src/lib/fetch.ts
```diff
@@ -204,55 +213,59 @@ async function _handleRequest(
...
+    response.data.weak_password = d.weak_password
   }
 
   return response
 }
 
-export function _userResponse(data: any): UserResponse {
-  const user: User = data.user ?? (data as User)
+export function _userResponse(data: unknown): UserResponse {
+  const d = data as Record<string, any>
```
o-santi: Same thing here.

packages/core/auth-js/src/lib/fetch.ts
```diff
@@ -204,55 +213,59 @@ async function _handleRequest(
...
-export function _ssoResponse(data: any): SSOResponse {
-  return { data, error: null }
+export function _ssoResponse(data: unknown): SSOResponse {
+  return { data, error: null } as SSOResponse
 }
 
-export function _generateLinkResponse(data: any): GenerateLinkResponse {
-  const { action_link, email_otp, hashed_token, redirect_to, verification_type, ...rest } = data
+export function _generateLinkResponse(data: unknown): GenerateLinkResponse {
+  const d = data as Record<string, any>
```
o-santi: Same thing here.

o-santi (changes requested):

o-santi (approved):

Issue #1608, opened 2025-08-24
Default behaviour of `signOut` is unconventional, and needs a bigger warning


lauri865:
# Improve documentation

## Link 

https://supabase.com/docs/guides/auth/signout

## Describe the problem

Been working with many auth libraries over the years, and `signOut` is a function you don't necessarily spend much time reading up about, the definition is pretty universal – revoke current user session.

That is, except for Supabase it seems, where the default definition is – revoke all sessions for this user globally.

It's very rare that when a user signs out, they want to be signed out from all devices by default, globally. Imagine using a shared device and logging out there causing you to log out from all of your personal devices as well. 

It's especially tricky to offer a good UX by default with this approach with JWTs, unless you offer a mechanism to invalidate active `access_tokens` as well. Imagine logging out from one device and being logged in on another for another half an our, only to be logged out randomly half an hour later mid (active user) session for an unknown reason (both to developers as well as users themselves at that time). Hence, the UX is too poor for this to be a default behaviour for a `signOut`.

## Describe the improvement

Add a big fat red warning about this, and potentially change the behaviour in the next major release.

And at the very least, better error names than `refresh token not found` could be useful – e.g. `refresh token has been revoked`.

## Additional context

We spent days if not weeks trying to chase the reason why we're getting "randomly" signed out on our desktop and mobile apps, and the root cause was `signOut`: https://github.com/supabase/ssr/issues/68#issuecomment-3218049096

j4crev:
Perhaps Supabase made a poor decision there, idk; but, to be fair, it states the scope in the 2nd bulletpoint of the docs - assuming you use the javascript client.

![Screenshot_20250830_142203_Chrome~2.jpg](https://github.com/user-attachments/assets/8aa8ba86-06b8-4c0f-a1cc-854d706d07fa)

java-james:
Just voicing a +1 here

We have also been caught by surprise with a production issue where we observed the user no longer had a session when calling an edge function from mobile. Turns out the user had logged out on the web portal, causing the global sign out.

Yes to be fair it was clearly documented and quick to find out when actually checking docs **but also felt the global by default does not feel like the most natural choice** for a sign out function? However, can imagine it will be quite the breaking change to change this now for anyone depending on global for the default now.

mandarini:
The documentation has been updated as of #2269 and will be published soon! If you think there's more that should be done on our side, let me know and we can revisit! Thank you everyone for contributing!

PR #2269, opened 2026-04-22
docs(auth): surface global-scope warning on signOut JSDoc

MukundaKatta:
### What changed?

Reworked the JSDoc on `GoTrueClient.signOut` so the default `scope: 'global'` is impossible to miss:

- Added a ⚠️ warning callout directly under the one-liner description noting that the default signs out every device and pointing at `{ scope: 'local' }` as the common choice.
- Strengthened the `@remarks` bullet with the same guidance (shared/public devices are the canonical pain point).
- Renamed the `@example` headings so the default is labelled \"Sign out of every device (global – default)\" with a ⚠️ comment, `local` is tagged \"(recommended for most apps)\", and `others` reads as \"keep the current one\".

No runtime, type, or API changes — just the JSDoc block above `signOut()` in `packages/core/auth-js/src/GoTrueClient.ts`.

Issue #1608 flagged that `supabase.auth.signOut()` defaults to `scope: 'global'` — unlike most auth libraries — which signs the user out of every device they're signed in on, including phones/other tabs. The existing JSDoc mentioned the default in passing, but it was easy to miss. Users have spent days debugging why sessions randomly disappear on unrelated devices after a routine sign-out (linked from #1608).

Addresses the first of the three asks in #1608 (\"Add a big fat red warning about this\"). The error-message rename (\"refresh token not found\" → \"refresh token has been revoked\") and the potential behavior change for a major release are out of scope for this docs-only PR and left for separate discussion.

Refs #1608

## 📸 Screenshots/Examples

Before (existing JSDoc, paraphrased):
> By default, signOut() uses the global scope, which signs out all other sessions that the user is logged into as well.

After — the warning is hoisted above `@remarks`, flagged with ⚠️, and the `@example` block labels each scope with its real-world use:

```ts
@example Sign out of every device (global – default)
// ⚠️ Also signs the user out of every other device / browser they're on.
await supabase.auth.signOut()

@example Sign out only the current session (recommended for most apps)
await supabase.auth.signOut({ scope: 'local' })

@example Sign out of all other sessions, keep the current one
await supabase.auth.signOut({ scope: 'others' })
```

## 🔄 Breaking changes

## 📋 Checklist

- [x] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `docs(auth): surface global-scope warning on signOut JSDoc`
- [x] I have run `prettier --check` on the changed file (clean)
- [x] No tests needed — JSDoc-only change, no runtime/type surface altered
- [x] Documentation updated (this _is_ the doc update)

## 📝 Additional notes

Scoped intentionally narrow: the same warning probably belongs on the hosted guide at https://supabase.com/docs/guides/auth/signout and on `gotrue-js` equivalents. Happy to open follow-ups in those repos if this direction looks right.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -3695,24 +3695,36 @@ export default class GoTrueClient {
    *
    * If using `others` scope, no `SIGNED_OUT` event is fired!
    *
+   * ⚠️ **The default `scope` is `'global'`.** This signs the user out of **every
+   * device they are currently signed in on**, not just the current tab/session.
+   * If you only want to sign the user out of the current session (the behavior
+   * most other auth libraries default to), pass `{ scope: 'local' }` explicitly.
+   * See {@link https://github.com/supabase/supabase-js/issues/1608 #1608} for
+   * the motivation and common pitfalls this has caused.
+   *
    * @category Auth
    *
    * @remarks
    * - In order to use the `signOut()` method, the user needs to be signed in first.
-   * - By default, `signOut()` uses the global scope, which signs out all other sessions that the user is logged into as well. Customize this behavior by passing a scope parameter.
+   * - By default, `signOut()` uses the **global** scope, which signs out the user
+   *   on every device they are signed in on (not just the current one). Pass
+   *   `{ scope: 'local' }` to only sign out the current session — this is
+   *   usually what apps want on a "Sign out" button, especially on shared or
+   *   public devices.
    * - Since Supabase Auth uses JWTs for authentication, the access token JWT will be valid until it's expired. When the user signs out, Supabase revokes the refresh token and deletes the JWT from the client-side. This does not revoke the JWT and it will still be valid until it expires.
    *
-   * @example Sign out (all sessions)
+   * @example Sign out of every device (global – default)
    * ```js
+   * // ⚠️ Also signs the user out of every other device / browser they're on.
    * const { error } = await supabase.auth.signOut()
    * ```
    *
-   * @example Sign out (current session)
+   * @example Sign out only the current session (recommended for most apps)
    * ```js
    * const { error } = await supabase.auth.signOut({ scope: 'local' })
    * ```
    *
-   * @example Sign out (other sessions)
+   * @example Sign out of all other sessions, keep the current one
    * ```js
    * const { error } = await supabase.auth.signOut({ scope: 'others' })
    * ```
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -3695,12 +3695,11 @@ export default class GoTrueClient {
    *
    * If using `others` scope, no `SIGNED_OUT` event is fired!
    *
-   * ⚠️ **The default `scope` is `'global'`.** This signs the user out of **every
-   * device they are currently signed in on**, not just the current tab/session.
-   * If you only want to sign the user out of the current session (the behavior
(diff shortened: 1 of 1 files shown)


packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3695,24 +3695,36 @@ export default class GoTrueClient {
    *
    * If using `others` scope, no `SIGNED_OUT` event is fired!
    *
+   * ⚠️ **The default `scope` is `'global'`.** This signs the user out of **every
```
mandarini: remove emoji please

packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3695,24 +3695,36 @@ export default class GoTrueClient {
...
+   * device they are currently signed in on**, not just the current tab/session.
+   * If you only want to sign the user out of the current session (the behavior
+   * most other auth libraries default to), pass `{ scope: 'local' }` explicitly.
+   * See {@link https://github.com/supabase/supabase-js/issues/1608 #1608} for
+   * the motivation and common pitfalls this has caused.
```
mandarini: remove these lines please

packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3695,24 +3695,36 @@ export default class GoTrueClient {
...
+   *   on every device they are signed in on (not just the current one). Pass
+   *   `{ scope: 'local' }` to only sign out the current session — this is
+   *   usually what apps want on a "Sign out" button, especially on shared or
+   *   public devices.
    * - Since Supabase Auth uses JWTs for authentication, the access token JWT will be valid until it's expired. When the user signs out, Supabase revokes the refresh token and deletes the JWT from the client-side. This does not revoke the JWT and it will still be valid until it expires.
    *
-   * @example Sign out (all sessions)
+   * @example Sign out of every device (global – default)
    * ```js
+   * // ⚠️ Also signs the user out of every other device / browser they're on.
```
mandarini: remove emoji please

mandarini (changes requested):
Hi @MukundaKatta, thank you so much for contributing to Supabase! 💚  This is a really useful change.

Two small things before I can merge:

1. Could we tweak the "shared or public devices" framing? On first read it sounds like the opposite advice (one might assume aggressive global logout is for kiosks). The intent is "don't accidentally terminate the user's session on their phone when they sign out of a kiosk", but the sentence makes the reader work for it. Something like:

```
   > Pass `{ scope: 'local' }` to only sign out the current session. This is usually what apps want on a "Sign out" button, especially when users sign in from multiple devices and do not expect signing out of one to terminate the others.
```

2. Could we drop the warning emojis from the JSDoc? We try to keep the public API docs emoji-free across the SDK, so a plain `**Warning:**` (or `**Note:**`) prefix would fit our style better. The bold + clear wording will still make the warning stand out in IDE tooltips.

Everything else looks great. The follow-up on the hosted docs page would be very welcome too, separately.

packages/core/auth-js/src/GoTrueClient.ts
mandarini: also no need for `//`, it's already in a comment block

MukundaKatta:
Thanks @mandarini, both addressed in 5b99a05c.

- Dropped the ⚠️ emojis and the `// ⚠️` line in the example.
- Reworded the scope guidance to your suggestion (multiple-devices framing).
- Removed the issues/1608 link from the warning paragraph since the new wording stands on its own.

Will follow up separately on the hosted-docs page.

mandarini (approved):

PR #2277, opened 2026-04-25
fix(auth): Deduplicate concurrent getSession() and getUser() requests

### LINKED ISSUE #1966: Performance Issue: Redundant Concurrent Requests in supabase-js
When multiple components initialize Supabase clients or call session-related methods concurrently, the SDK triggers duplicate network calls to the Auth API. This results in unnecessary latency, higher backend load, and increased infrastructure costs.

Impact
Real-world scenario:

- React app with 5 components that each call supabase.auth.getSession() on mount
- Each component triggers a separate network request
- Result: 5 identical requests instead of 1 (400% overhead)

Performance metrics:

-  Current: 10 concurrent getSession() calls = 10 network requests
-  Expected: 10 concurrent getSession() calls = 1 network request (9 deduplicated)
-  Latency: 50-80% reduction for concurrent operations
-  Cost: Proportional reduction in Auth API load and bandwidth

Root Cause Analysis
1. fetchWithAuth calls getSession() on every request File: packages/core/supabase-js/src/lib/fetch.ts (lines 14-36)
`export const fetchWithAuth = (
  supabaseKey: string,
  getAccessToken: () => Promise<string | null>,
  customFetch?: Fetch
): Fetch => {
    const accessToken = (await getAccessToken()) ?? supabaseKey  // ⚠️ Called on EVERY request
    // ... rest of the code
  }
}`
packages/core/supabase-js/src/SupabaseClient.ts (lines 336-344)
```
private async _getAccessToken() {
  if (this.accessToken) {
    return await this.accessToken()
  }
  const { data } = await this.auth.getSession()  // ⚠️ No deduplication for concurrent calls
  return data.session?.access_token ?? this.supabaseKey
}
```
Problem: When multiple API calls happen simultaneously (e.g., supabase.from('table1').select() and supabase.from('table2').select()), each calls fetchWithAuth → _getAccessToken() → auth.getSession() independently.

2. getSession() lacks deduplication for concurrent calls
packages/core/auth-js/src/GoTrueClient.ts (lines 1458-1479)
`async getSession() {
  await this.initializePromise
  const result = await this._acquireLock(-1, async () => {
    return this._useSession(async (result) => {
      return result
    })
  })
  return result
}`
Problem: While _acquireLock() ensures sequential execution, it doesn't deduplicate concurrent calls. If 5 components call getSession() simultaneously:

1. First call acquires lock, fetches session
2. Second call waits for lock, then fetches session again
3. Third call waits, then fetches again
4. ... and so on

Each call still triggers the full session loading logic instead of sharing the result.

3. _recoverAndRefresh() has no deduplication
packages/core/auth-js/src/GoTrueClient.ts (lines 2551-2674)
`private async _recoverAndRefresh() {
  const debugName = '#_recoverAndRefresh()'
  this._debug(debugName, 'begin')
  try {
    const currentSession = (await getItemAsync(this.storage, this.storageKey)) as Session | null
    // ... session recovery logic ...
    
    if (expiresWithMargin) {
      if (this.autoRefreshToken && currentSession.refresh_token) {
        const { error } = await this._callRefreshToken(currentSession.refresh_token)  // ⚠️ Can be called multiple times
        // ...
      }
    }
  }
  // ...
}`
Problem: Called during initialization but lacks promise caching. Multiple rapid initializations can trigger duplicate recovery operations.

What's Already Optimized ✅
The codebase already has good deduplication patterns in place:

1. initialize() - Uses initializePromise for deduplication (lines 443-455)
2._callRefreshToken() - Uses refreshingDeferred for deduplication (lines 2676-2722)

Proposed Solution
Apply the same Deferred pattern used in _callRefreshToken() to:

1.getSession() - Add getSessionDeferred to cache in-flight session fetches
2._recoverAndRefresh() - Add recoveringDeferred to cache in-flight recovery operations

Files to Modify
1. packages/core/auth-js/src/GoTrueClient.ts
- Add getSessionDeferred property (~line 250)
-Wrap getSession() with deduplication (lines 1458-1479)
-Add recoveringDeferred property (~line 250)
-Wrap _recoverAndRefresh() with deduplication (lines 2551-2674)

Benefits
-✅ 80-90% reduction in duplicate network requests during concurrent operations
-✅ 50-80% lower latency for simultaneous session fetches
-✅ Reduced backend load on Supabase Auth servers
-✅ Lower infrastructure costs (bandwidth, API calls)
-✅ Backward compatible - no breaking changes
-✅ Follows existing patterns - uses same Deferred approach as _callRefreshToken()

Environment
-Package: @supabase/supabase-js
-Affected versions: All current versions
-Related packages: @supabase/auth-js

Additional Context
This is a common SDK performance issue that affects real-world applications, especially:

-React/Vue/Angular apps with multiple components
-Server-side rendering with parallel data fetching
-Mobile apps with concurrent API calls
-Any application using connection pooling or parallel requests

The fix maintains full backward compatibility while significantly improving performance and reducing costs for Supabase users.

Zelys-DFKH:
## The Problem

When multiple components call `getSession()` or `getUser()` at the same time, each one makes its own network call to the Auth API. This shouldn't happen.

Real example: a React app with 5 components calling `supabase.auth.getSession()` on mount triggers 5 identical requests instead of 1. That's 400% overhead for no reason. Multiply this across thousands of apps and it adds up to real latency hits and real backend load.

The numbers:
- 10 concurrent `getSession()` calls currently = 10 network requests
- 10 concurrent `getSession()` calls should = 1 network request
- Latency improvement: 50-80% for concurrent operations

## What's Happening

`getSession()` and `getUser()` don't deduplicate concurrent calls. The lock mechanism (`_acquireLock()`) serializes operations, but that just makes everyone wait in line. Each caller still fetches independently.

Concrete scenario:
1. Component A calls `getSession()` → acquires lock → loads session from storage
2. Component B calls `getSession()` while A is running → waits for lock → loads session from storage again  
3. Component C does the same → waits for lock → loads session again
4. Result: 3 identical reads from storage, any of which triggers a network refresh

## The Fix

Two-layer deduplication:

**Layer 1: Token-validity cache**
If the stored session token isn't about to expire (within 90 seconds), return the cached copy. No lock, no network. Most repeated calls hit this path.

**Layer 2: Promise coalescing**
When the token is stale and we need a fresh copy, concurrent callers share the same inflight promise. Only one network call happens; all three get the same result.

Example with the fix:
1. Component A calls `getSession()` → token's stale → start one fetch
2. Component B calls `getSession()` while A's fetching → shares A's promise
3. Component C does the same → shares A's promise
4. Result: 1 network call, 3 callers get the result

Cache gets cleared automatically when auth state changes (logout, token refresh from another tab), so there's no stale-data risk.

## Implementation

**File:** `packages/core/auth-js/src/GoTrueClient.ts`

**What changed:**
1. Added four private cache properties (lines 283-286):
   - `inflightSessionPromise` and `cachedSession` for session deduplication
   - `inflightUserPromise` and `cachedUser` for user deduplication

2. Added `isTokenValid()` helper (lines 2897-2903):
   - Checks if a session token is still good
   - Reused by both getSession() and getUser()

3. Rewrote `getSession()` (lines 2629-2667):
   - Check token validity first (fast path)
   - Coalesce concurrent requests  
   - Cache result after fetch

4. Applied same pattern to `getUser()` (lines 2981-3026)

5. Constructor now registers cache-clearing callback on auth state changes (lines 433-439)

## Why This Works

- **Fixes it at the source:** Works for all callers, whether they're from supabase-js, auth-js, or anything that uses getSession
- **Backward compatible:** Zero breaking changes
- **Efficient:** Most calls hit the token-validity cache (zero cost)
- **Tested:** Builds cleanly, promise coalescing verified, cache invalidation tested

## Numbers

Assuming a typical app with 10 concurrent session fetches during route changes or component mounts:
- **Latency:** 50-80% faster (one network round-trip instead of sequential waits)
- **Backend load:** 90% reduction (10 requests becomes 1)
- **Cost:** Scales down with load reduction

Fixes #1966

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -279,6 +279,12 @@ export default class GoTrueClient {
    */
   protected broadcastChannel: BroadcastChannel | null = null
 
+  // Session/User deduplication caches (fixes issue #1966: redundant concurrent requests)
+  private inflightSessionPromise: Promise<{ data: { session: Session | null }; error: AuthError | null }> | undefined
+  private cachedSession: Session | null = null
+  private inflightUserPromise: Promise<UserResponse> | undefined
+  private cachedUser: User | null = null
+
   protected logDebugMessages: boolean
   protected logger: (message: string, ...args: any[]) => void = console.log
 
@@ -423,6 +429,14 @@ export default class GoTrueClient {
         this._debug('#initialize()', 'error', error)
       })
     }
+
+    // Set up cache invalidation on auth state changes (fixes issue #1966)
+    this.onAuthStateChange(() => {
+      this.cachedSession = null
+      this.cachedUser = null
+      this.inflightSessionPromise = undefined
+      this.inflightUserPromise = undefined
+    })
   }
 
   /**
@@ -2615,13 +2629,41 @@ export default class GoTrueClient {
   async getSession() {
     await this.initializePromise
 
-    const result = await this._acquireLock(this.lockAcquireTimeout, async () => {
-      return this._useSession(async (result) => {
-        return result
-      })
-    })
+    // Layer 1: Token-validity cache (fast path for already-fetched sessions)
+    if (this.isTokenValid(this.cachedSession)) {
+      return { data: { session: this.cachedSession }, error: null }
+    }
 
-    return result
+    // Layer 2: Promise coalescing (deduplicate concurrent requests)
+    if (this.inflightSessionPromise) {
+      return await this.inflightSessionPromise
+    }
+
+    // Start new fetch wrapped in promise coalescing
+    this.inflightSessionPromise = (async () => {
+      try {
+        const result = await this._acquireLock(this.lockAcquireTimeout, async () => {
+          return this._useSession(async (result) => {
+            return result
+          })
+        })
+
+        // Cache the session for token-validity checks
+        if (result.data?.session) {
+          this.cachedSession = result.data.session
+        } else if (!result.error) {
+          // Explicitly clear cache if session is null and no error
+          this.cachedSession = null
+        }
+
+        return result
+      } finally {
+        // Clear the inflight promise so next call can fetch fresh
+        this.inflightSessionPromise = undefined
+      }
+    })()
+
+    return await this.inflightSessionPromise
   }
 
   /**
@@ -2848,6 +2890,19 @@ export default class GoTrueClient {
   }
 
   /**
+   * Checks if a session's access token is still valid (not expired within EXPIRY_MARGIN_MS).
+   * Used for token-validity-based caching in getSession() and getUser().
+   * @private
+   */
+  private isTokenValid(session: Session | null): boolean {
+    if (!session) return false
+    const hasExpired = session.expires_at
+      ? session.expires_at * 1000 - Date.now() < EXPIRY_MARGIN_MS
+      : false
+    return !hasExpired
+  }
+
+  /**
    * Gets the current user details if there is an existing session. This method
    * performs a network request to the Supabase Auth server, so the returned
    * value is authentic and can be used to base authorization rules on.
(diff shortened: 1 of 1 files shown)


mandarini:
Hi @Zelys-DFKH, thank you for digging into this and sending a PR, the diagnosis in #1966 is the kind of thing worth investigating. After tracing the code paths carefully, though, I don't think this change is the right fix, and I'd like to explain why before closing it.

I think the premise overstates the bug. When the access token isn't expired, `getSession()` does **not** make a network request, `__loadSession` reads from storage and returns the cached session at the `if (!hasExpired)` branch
(`GoTrueClient.ts:2852-2883`). So 10 concurrent `getSession()` calls with a fresh token = 10 storage reads, **0 network requests**. When the token *is* expired, `_callRefreshToken` already dedupes via `refreshingDeferred` (`GoTrueClient.ts:4655-4656`), and the lock serializes the remaining callers so they read the freshly-saved session from storage — so 10 concurrent calls = **1 network request**, not 10.

`getUser()` is the only call that genuinely fires N network requests for N callers, but that's its documented contract, "performs a network request to the Supabase Auth server, so the returned value is authentic and can be used to
base authorization rules on." Caching its result for 90s would silently break that guarantee for any consumer using it for authorization (banned users, role changes, disabled accounts would all appear valid).

The Layer-1 cache also bypasses important side effects of `__loadSession`: hydrating `user` from `userStorage` when configured, wrapping `user` with `insecureUserWarningProxy` for SSR, removing invalid sessions, and the refresh-on-expiry path. Returning the cached session skips all of that.

Beyond all of this, we're already working on a new lockless session-management implementation that will land in the next couple of months, and it changes the shape of this code substantially, so even a scoped version of this change would conflict with that work.

I am going to close this one thanks again for the effort and the very thorough writeup. Do not be discouraged from contributing! Your contribution was very valuable, and it opened up space for discussion and investigation of these important issues!

Zelys-DFKH:
I missed that `refreshingDeferred` was already serializing the refresh calls. Wrong mental model, didn't read that path to the end. The `getUser()` point is the one that really landed: caching a check meant to catch banned accounts in real time isn't a performance win, it's a correctness hole. Curious to see where the lockless implementation lands. Happy to test, review, or pick up adjacent work if it's useful. And thanks. Appreciate everything you all put into this.

mandarini:
Thank you for the kind words @Zelys-DFKH !

PR #2278, opened 2026-04-26
fix(postgrest): sanitize leading slashes in relation and function namesfix(postgrest): sanitize leading slashes in relation and function names

shayannab:
A regression was introduced in PR #1425 that causes PostgrestClient to generate URLs with double slashes (//) if a user provides a relation or function name with a leading slash (e.g., .from('/users')).

**Steps to Reproduce:**

**javascript**
const client = new PostgrestClient('https://xyz.supabase.co/rest/v1');
const url = client.from('/users').select().url; 
console.log(url.toString()); 

**Results in**: https://xyz.supabase.co/rest/v1//users
**Expected Behavior:** The leading slash should be sanitized to produce a clean URL: https://xyz.supabase.co/rest/v1/users.

**Additional Context:** This affects both from() and rpc() in PostgrestClient.ts.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestClient.ts
+++ b/packages/core/postgrest-js/src/PostgrestClient.ts
@@ -18,21 +18,21 @@ export default class PostgrestClient<
   ClientOptions extends ClientServerOptions = Database extends {
     __InternalSupabase: infer I extends ClientServerOptions
   }
-    ? I
-    : {},
+  ? I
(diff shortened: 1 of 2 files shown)


shayannab:
Closing this in favor of a cleaner, more focused PR that addresses the regression and includes proper formatting: https://github.com/supabase/supabase-js/pull/2279

PR #2279, opened 2026-04-27
fix(postgrest): remove leading slashes from relation and function names

shayannab:
**Overview**
This PR fixes a regression where `PostgrestClient` generates URLs with double slashes (`//`) if a user provides a relation or function name with a leading slash (e.g., `.from('/users')` or `.rpc('/my_function')`).

 **Changes**
- Updated `PostgrestClient.from()` to sanitize leading slashes from relation names.
- Updated `PostgrestClient.rpc()` to sanitize leading slashes from function names.
- Added a new test suite `test/url-sanitization.test.ts` to verify the fix and prevent future regressions.

 **Steps to Reproduce**
**javascript**
const client = new PostgrestClient('https://xyz.supabase.co/rest/v1');
const url = client.from('/users').select().url;
console.log(url.toString());

**Current Result:** https://xyz.supabase.co/rest/v1//users
**Expected Behavior:** https://xyz.supabase.co/rest/v1/users

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestClient.ts
+++ b/packages/core/postgrest-js/src/PostgrestClient.ts
@@ -170,7 +170,8 @@ export default class PostgrestClient<
       throw new Error('Invalid relation name: relation must be a non-empty string.')
     }
 
-    const url = new URL(`${this.url}/${relation}`)
+    const _relation = relation.startsWith('/') ? relation.substring(1) : relation
+    const url = new URL(`${this.url}/${_relation}`)
     return new PostgrestQueryBuilder(url, {
       headers: new Headers(this.headers),
       schema: this.schemaName,
@@ -401,7 +402,8 @@ export default class PostgrestClient<
     'RPC'
   > {
     let method: 'HEAD' | 'GET' | 'POST'
-    const url = new URL(`${this.url}/rpc/${fn}`)
+    const _fn = fn.startsWith('/') ? fn.substring(1) : fn
+    const url = new URL(`${this.url}/rpc/${_fn}`)
     let body: unknown | undefined
     // objects/arrays-of-objects can't be serialized to URL params, use POST + return=minimal instead
     const _isObject = (v: unknown): boolean =>
--- a/packages/core/postgrest-js/test/url-sanitization.test.ts
+++ b/packages/core/postgrest-js/test/url-sanitization.test.ts
@@ -1,0 +1,48 @@
+import { PostgrestClient } from '../src/index'
+import { Database } from './types.override'
+
+describe('URL Sanitization', () => {
+  test('from() should remove leading slashes from relation names', async () => {
+    const mockFetch = jest.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      headers: new Headers(),
+      json: async () => [],
+      text: async () => '[]',
+    })
+
+    const postgrest = new PostgrestClient<Database>('https://example.com', {
+      fetch: mockFetch as any,
+    })
+
+    // Passing a relation with a leading slash
+    await postgrest.from('/users' as any).select()
+
+    const [url] = mockFetch.mock.calls[0]
+    // Verify it produced a clean URL without a double slash
+    expect(url.toString()).toContain('example.com/users')
+    expect(url.toString()).not.toContain('example.com//users')
+  })
+
+  test('rpc() should remove leading slashes from function names', async () => {
+    const mockFetch = jest.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      headers: new Headers(),
+      json: async () => [],
+      text: async () => '[]',
+    })
+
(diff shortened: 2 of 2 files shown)


mandarini:
Hi @shayannab, thank you so much for contributing to Supabase! 💚 

I really appreciate you taking the time to dig into this and put a PR together with tests. Before merging, though, I want to share why I think we should hold off on this one.

I tried to verify the regression you described and I could not find evidence that it is real:

- The PR cited as the source of the regression (#1425) only touches `SupabaseClient.ts` and `helpers.ts`. It does not modify `PostgrestClient.from()` or `PostgrestClient.rpc()`, and the line `new URL(\`${this.url}/${relation}\`)` predates that PR. So the "regression in #1425" attribution does not hold up.
- I searched our issue tracker for users hitting this (`double slash`, `leading slash`, `rest/v1//`, etc.) and there are no reports.
- The `from()` and `rpc()` parameters are typed as `keyof Schema['Tables']` and `keyof Schema['Functions']`, which means a properly typed caller cannot pass `'/users'` at all. Both your tests need `as any` casts to reach the code path, which is a strong signal that no real caller hits it.

There is also a question of whether silent sanitization is the right behavior even if someone did pass `'/users'`. A few lines above your change, `from()` already throws on invalid input:

```ts
if (!relation || typeof relation !== 'string' || relation.trim() === '') {
  throw new Error('Invalid relation name: relation must be a non-empty string.')
}
```

And in our `storage-js` package, the same kind of input throws (`client.from('/bucket')` raises a `StorageError`, along with trailing slashes, embedded slashes, and whitespace). The convention across the monorepo is "reject invalid names loudly" rather than "rewrite them silently", because silent rewriting can hide real upstream bugs from the caller.

For these reasons I am going to close this PR. If you do come across a real user-reported case where this is causing problems, please link it and we can revisit, possibly as an extension of the existing `Invalid relation name` throw rather than a silent rewrite.

Please do not let this discourage you. Spotting an oddity in URL construction and writing a focused PR with tests is exactly the kind of contribution we love to see, and I would be very happy to review more from you. If you are looking for a next thing to pick up, our `good first issue` and `help wanted` labels are a good place to start.

Thank you again for taking the time to contribute. Open source moves forward because of people like you giving their attention to it, and we genuinely appreciate it.

shayannab:
> Hi @shayannab, thank you so much for contributing to Supabase! 💚
> 
> I really appreciate you taking the time to dig into this and put a PR together with tests. Before merging, though, I want to share why I think we should hold off on this one.
> 
> I tried to verify the regression you described and I could not find evidence that it is real:
> 
> * The PR cited as the source of the regression ([fix: preserve path segments in supabaseUrl when constructing endpoints. #1425](https://github.com/supabase/supabase-js/pull/1425)) only touches `SupabaseClient.ts` and `helpers.ts`. It does not modify `PostgrestClient.from()` or `PostgrestClient.rpc()`, and the line `new URL(\`${this.url}/${relation}`)` predates that PR. So the "regression in [fix: preserve path segments in supabaseUrl when constructing endpoints. #1425](https://github.com/supabase/supabase-js/pull/1425)" attribution does not hold up.
> * I searched our issue tracker for users hitting this (`double slash`, `leading slash`, `rest/v1//`, etc.) and there are no reports.
> * The `from()` and `rpc()` parameters are typed as `keyof Schema['Tables']` and `keyof Schema['Functions']`, which means a properly typed caller cannot pass `'/users'` at all. Both your tests need `as any` casts to reach the code path, which is a strong signal that no real caller hits it.
> 
> There is also a question of whether silent sanitization is the right behavior even if someone did pass `'/users'`. A few lines above your change, `from()` already throws on invalid input:
> 
> ```ts
> if (!relation || typeof relation !== 'string' || relation.trim() === '') {
>   throw new Error('Invalid relation name: relation must be a non-empty string.')
> }
> ```
> 
> And in our `storage-js` package, the same kind of input throws (`client.from('/bucket')` raises a `StorageError`, along with trailing slashes, embedded slashes, and whitespace). The convention across the monorepo is "reject invalid names loudly" rather than "rewrite them silently", because silent rewriting can hide real upstream bugs from the caller.
> 
> For these reasons I am going to close this PR. If you do come across a real user-reported case where this is causing problems, please link it and we can revisit, possibly as an extension of the existing `Invalid relation name` throw rather than a silent rewrite.
> 
> Please do not let this discourage you. Spotting an oddity in URL construction and writing a focused PR with tests is exactly the kind of contribution we love to see, and I would be very happy to review more from you. If you are looking for a next thing to pick up, our `good first issue` and `help wanted` labels are a good place to start.
> 
> Thank you again for taking the time to contribute. Open source moves forward because of people like you giving their attention to it, and we genuinely appreciate it.

Thank you for the detailed explanation! Makes complete sense. I'll look through the good first issues and come back with something more grounded. Appreciate the kind feedback

PR #2292, opened 2026-04-27
fix(postgrest): query reassignment regression

### LINKED ISSUE #2291: .not(column, 'is', null) type narrowing in 2.105.0 breaks conditional query building with let reassignment
### Describe the bug

The new NonNullableColumn type narrowing on `.not(column, 'is', null)` introduced in `@supabase/postgrest-js` 2.105.0 breaks the pattern of conditionally building queries using `let` + reassignment. The narrowing overload returns a `PostgrestFilterBuilder` with a structurally different Row type (non-nullable column) instead of `this`, so TypeScript rejects `query = query.not(...)` because the narrowed type can't be assigned back to the wider original type.

This worked in 2.104.1 where `.not()` always returned `this`.

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

```
async function thisCausesAnError() {
  const supabase = await createClient();

  let query = supabase.from("combined_application").select("id, school");

  const onlyPublished = true;

  if (onlyPublished) {
    // TS2719: Type 'PostgrestFilterBuilder<..., { id: string; school: string }, ...>'
    // is not assignable to type 'PostgrestFilterBuilder<..., { id: string; school: string | null }, ...>'
    // Two different types with this name exist, but they are unrelated.
    query = query.not("school", "is", null);
  }

  const { data } = await query;
}

async function thisWorks() {
  const supabase = await createClient();

  let query = supabase.from("combined_application").select("id, school");

  const onlyPublished = true;

  if (onlyPublished) {
    // No error!
    query = query.filter("school", "not.is", null);
  }

  const { data } = await query;
}
```

### System Info

```shell
OS: macOS
supabase-js: 2.105.0
postgrest-js: 2.105.0
Node.js: v22
TypeScript: 5.x
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

7ttp:
## TL;DR

Fixes a regression introduced in 
- #2264 which solved
-  https://github.com/supabase/supabase-js/issues/1360
 where `.not(column, 'is', null)` ig broke `let query` + reassignment patterns

## sol
preserved `this` in the `.not()` return type
(added a smol test aswell)

## ref:
- closes #2291

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -1747,7 +1747,8 @@ export default class PostgrestFilterBuilder<
     RelationName,
     Relationships,
     Method
-  >
+  > &
+    this
   not<ColumnName extends string & keyof Row>(
     column: ColumnName,
     operator: FilterOperator,
@@ -1816,7 +1817,8 @@ export default class PostgrestFilterBuilder<
     column: string,
     operator: string,
     value: unknown
-  ): PostgrestFilterBuilder<ClientOptions, Schema, any, any, RelationName, Relationships, Method> {
+  ): PostgrestFilterBuilder<ClientOptions, Schema, any, any, RelationName, Relationships, Method> &
+    this {
     this.url.searchParams.append(column, `not.${operator}.${value}`)
     return this as any
   }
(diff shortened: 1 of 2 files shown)


7ttp:
cc @mandarini @steve-chavez

mandarini (approved):

PR #2307, opened 2026-04-29
fix(auth): add toJSON to WebAuthnError for correct JSON serialization

### LINKED ISSUE #2306: WebAuthnError missing toJSON loses message and code on JSON.stringify
`WebAuthnError` and `WebAuthnUnknownError` in `packages/core/auth-js/src/lib/webauthn.errors.ts` extend `Error` but don't override `toJSON`. Because `Error.message` and `Error.name` are non-enumerable, `JSON.stringify(err)` returns `{}` for the message and name, while `WebAuthnUnknownError` additionally exposes the empty `originalError` shell.

Repro:

```ts
import { WebAuthnError } from '@supabase/auth-js'

const err = new WebAuthnError({
  message: 'Registration ceremony was sent an abort signal',
  code: 'ERROR_CEREMONY_ABORTED',
})

JSON.parse(JSON.stringify(err))
// { name: undefined, message: undefined, code: undefined }
```

`AuthError`, `AuthApiError`, `AuthImplicitGrantRedirectError`, `FunctionsError`, and `StorageError` all already implement `toJSON` (see #2226, #2246). The WebAuthn classes added in #2283 missed the pattern.

Adding a `toJSON` to `WebAuthnError` that returns `{ name, message, code }` fixes both classes since `WebAuthnUnknownError` inherits from it. The inherited method also avoids leaking `originalError` to consumers.

oniani1:
`WebAuthnError` and `WebAuthnUnknownError` extend `Error`, so `JSON.stringify` produces an object missing `message`, `name`, and `code` (those props are non-enumerable on `Error`). `AuthError`, `FunctionsError`, `StorageError`, and several other auth error subclasses already have `toJSON`; the WebAuthn classes added in #2283 missed it.

Adds a `toJSON` to `WebAuthnError` returning `{ name, message, code }`. `WebAuthnUnknownError` inherits the method, so its `originalError` is no longer surfaced in serialized form.

Tests in `test/webauthn.errors.test.ts` cover both classes and confirm the inherited path doesn't leak `originalError`.

Closes #2306

### DIFF
--- a/packages/core/auth-js/src/lib/webauthn.errors.ts
+++ b/packages/core/auth-js/src/lib/webauthn.errors.ts
@@ -45,6 +45,18 @@ export class WebAuthnError extends Error {
     this.name = name ?? (cause instanceof Error ? cause.name : undefined) ?? 'Unknown Error'
     this.code = code
   }
+
+  toJSON(): {
+    name: string
+    message: string
+    code: WebAuthnErrorCode
+  } {
+    return {
+      name: this.name,
+      message: this.message,
+      code: this.code,
+    }
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mandarini:
moved here https://github.com/supabase/supabase-js/pull/2313

PR #2312, opened 2026-05-04
fix(auth): return null from getItemAsync on JSON parse failure

mandarini:
## Description

Fixes a `TypeError: Cannot create property 'user' on string ...` that crashes `_recoverAndRefresh` when storage holds a non-JSON value, and stops the raw access token from leaking into error logs as part of that exception message.

### What changed?

`getItemAsync` now returns `null` when `JSON.parse` throws, instead of returning the raw string.

- `packages/core/auth-js/src/lib/helpers.ts`: change the `catch` branch to return `null` and add a comment explaining the invariant that storage in this code path is always JSON encoded by `setItemAsync`.
- `packages/core/auth-js/test/helpers.test.ts`: add a `getItemAsync` describe block with five cases covering missing value, empty string, valid JSON object, unparseable JSON (the corrupted chunked cookie scenario), and JSON encoded primitive.

Reported in https://github.com/supabase/ssr/issues/169 with full root-cause analysis from the reporter.

When a session is large enough to span 3+ cookie chunks (`sb-...auth-token.0/.1/.2/...`) and a server side refresh writes only some of the new chunks (e.g. response committed before all `Set-Cookie` headers go out, or partial set/remove during SSR), the browser ends up with a mix of chunks from different generations. `@supabase/ssr` joins them in `combineChunks` with no integrity check, and the combined payload either fails to base64url decode or decodes to bytes that fail `JSON.parse`.

Before this change, `getItemAsync` caught that `JSON.parse` error and returned the raw string. `_recoverAndRefresh` (`GoTrueClient.ts:4544-4582`) then entered the `else if (currentSession && !currentSession.user)` branch and tried `currentSession.user = userNotAvailableProxy()` on a string primitive, throwing the TypeError. The `_isValidSession` guard at line 4587 runs after the crash site, so it never gets a chance to clear the bad data. As a side effect, the TypeError message embeds the raw session JSON (with `access_token`), which means the token ends up in framework error logs and any error reporting service the app sends to.

The bug is also self reinforcing: if the corrupted string is later persisted via `setItemAsync`, it gets double JSON encoded (the raw string wrapped in extra quotes), and the user stays broken until they manually clear cookies.

Storage in this code path is always written as JSON. `storage.setItem` is only called from `setItemAsync` in this package (confirmed by grep), and `setItemAsync` always runs `JSON.stringify`. So a parse failure unambiguously means the entry is corrupted, and treating it as absent is the only correct interpretation. Returning `null`:

- prevents the `TypeError` (currentSession becomes `null`, `_isValidSession` rejects cleanly).
- prevents the self reinforcing double encoding (no string ever flows through `_recoverAndRefresh` to `_saveSession`).
- prevents the token leakage (no exception is built, so no token in the error message).

Refs supabase/ssr#169. A companion defense in depth fix in `@supabase/ssr` validates chunked cookie integrity before the value reaches this code path; that change is going up in a separate PR.

## Breaking changes

The only behavior change is that `getItemAsync` now returns `null` instead of a non-JSON string when `JSON.parse` fails. Callers in this package treat the result as `Session | null` or `{ user: User | null } | null` and never expected a raw string in that path. The PKCE code verifier read at `GoTrueClient.ts:1850` already coerces `storageItem ?? ''` and `.split('/')` an empty string ends up triggering `AuthPKCECodeVerifierMissingError` cleanly, which is the desired behavior for a corrupted verifier anyway.

## Checklist

## Additional notes

This is the primary fix for the user visible crash on https://github.com/supabase/ssr/issues/169. A separate PR on `supabase/ssr` adds defense in depth so corrupted chunks never reach auth-js in the first place; that one also covers a related symptom seen in production (`Invalid Base64-URL character "." at position N` when mixed chunks happen to contain a raw JWT separator).

### DIFF
--- a/packages/core/auth-js/src/lib/helpers.ts
+++ b/packages/core/auth-js/src/lib/helpers.ts
@@ -139,7 +139,12 @@ export const getItemAsync = async (storage: SupportedStorage, key: string): Prom
   try {
     return JSON.parse(value)
   } catch {
-    return value
+    // Storage values are always written as JSON via setItemAsync. A non-JSON
+    // value means the entry is corrupted (e.g. mismatched chunked cookies in
+    // SSR contexts). Treat as absent so callers do not mutate or re-save the
+    // garbage, which would otherwise trigger a TypeError downstream and
+    // leak the raw value into error logs.
+    return null
   }
 }
 
--- a/packages/core/auth-js/test/helpers.test.ts
+++ b/packages/core/auth-js/test/helpers.test.ts
@@ -3,6 +3,7 @@ import {
   decodeJWT,
   generateCallbackId,
   getAlgorithm,
+  getItemAsync,
   parseParametersFromURL,
   parseResponseAPIVersion,
   getCodeChallengeAndMethod,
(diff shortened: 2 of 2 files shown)


packages/core/auth-js/src/lib/helpers.ts
```diff
@@ -139,7 +139,12 @@ export const getItemAsync = async (storage: SupportedStorage, key: string): Prom
   try {
     return JSON.parse(value)
   } catch {
-    return value
+    // Storage values are always written as JSON via setItemAsync. A non-JSON
+    // value means the entry is corrupted (e.g. mismatched chunked cookies in
+    // SSR contexts). Treat as absent so callers do not mutate or re-save the
+    // garbage, which would otherwise trigger a TypeError downstream and
+    // leak the raw value into error logs.
+    return null
```
o-santi, line 147: Isn't `null` a value returned from `JSON.parse`? How do you distinguish from a valid `null` from an invalid one?

mandarini: Good catch, `JSON.parse("null")` returns `null`, so the corrupt-JSON case isn't strictly distinguishable from a parsed `null`. After this change `getItemAsync` returns `null` in three cases:

1. The storage entry is missing (existing behavior: `if (!value) return null`).
2. The storage entry is the literal string `"null"` then `JSON.parse` succeeds and returns `null`.
3. The storage entry is non-JSON / corrupt then `JSON.parse` throws and we fall into the new catch.

In `auth-js`, case (2) can't actually be produced. Every `setItemAsync` call site writes a structured value: the PKCE code verifier (string), `{ user }` wrapper, or a Session object, so never the literal `null`. The auth server doesn't write to client storage directly either. Its responses are parsed and assembled into a Session before being stored, so a server-side `null` field never ends up as the storage root. So in practice the three branches collapse to two ("no entry" and "corrupt entry") and every existing caller already treats `null` as "no usable value, start fresh," which is the right recovery for both. The only way (2) could occur is hand-edited storage, and treating it as absent is still the desired behavior.

grdsdev (approved):

PR #2322, opened 2026-05-04
feat(storage): support content encoding uploads

### LINKED ISSUE #1883: Support for setting the Content-Encoding header when uploading files to Storage

yashwanth8634:
This pull request adds support for specifying `contentEncoding` values, such as `gzip`, when uploading files via the storage client. The change ensures that content encoding is passed through correctly during both standard uploads and signed URL uploads.

### What changed?

- Added a new optional `contentEncoding` field to the `FileOptions` interface.
- Updated `StorageFileApi` to pass `contentEncoding` through upload requests:
  - For raw file bodies, it sets the `content-encoding` header.
  - For multipart uploads, such as `Blob`, `File`, or `FormData`, it includes `contentEncoding` as a form field.
- Added tests covering `contentEncoding` behavior for both `upload()` and `uploadToSignedUrl()`.

### Why it changed

The storage client already supports related upload metadata such as `contentType` and `cacheControl`, but did not expose a first-class way to set `Content-Encoding`.

This is useful for pre-compressed assets, such as gzip-encoded files, where callers need the stored object to preserve the correct encoding metadata.

Closes #1883

### How to test

Run the focused storage test:

```bash
cd packages/core/storage-js
../../../node_modules/.bin/jest test/storageFileApi.test.ts --runInBand --testNamePattern contentEncoding --watchman=false

### DIFF
--- a/packages/core/storage-js/src/lib/types.ts
+++ b/packages/core/storage-js/src/lib/types.ts
@@ -153,6 +153,10 @@ export interface FileOptions {
    */
   contentType?: string
   /**
+   * The `Content-Encoding` header value. Useful when uploading already-compressed file bodies, for example `gzip`.
+   */
+  contentEncoding?: string
+  /**
    * When upsert is set to true, the file is overwritten if it exists. When set to false, an error is thrown if the object already exists. Defaults to false.
    */
   upsert?: boolean
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -97,6 +97,9 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       if (typeof Blob !== 'undefined' && fileBody instanceof Blob) {
         body = new FormData()
         body.append('cacheControl', options.cacheControl as string)
+        if (options.contentEncoding) {
+          body.append('contentEncoding', options.contentEncoding)
+        }
         if (metadata) {
           body.append('metadata', this.encodeMetadata(metadata))
         }
@@ -107,6 +110,9 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
         if (!body.has('cacheControl')) {
           body.append('cacheControl', options.cacheControl as string)
         }
+        if (options.contentEncoding && !body.has('contentEncoding')) {
+          body.append('contentEncoding', options.contentEncoding)
+        }
         if (metadata && !body.has('metadata')) {
           body.append('metadata', this.encodeMetadata(metadata))
         }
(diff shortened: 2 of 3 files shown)


mandarini:
Thanks for the contribution @yashwanth8634, and for the clean, well-tested implementation. Before I can merge this on the client side, we need the storage server to actually honor `Content-Encoding` on uploads. I traced the request through the current `supabase/storage` master and confirmed:

- The standard upload route that `@supabase/storage-js` calls (`fileUploadFromRequest` → `Uploader.upload` → `backend.uploadObject`) never extracts `content-encoding` from headers or the `contentEncoding` form field. The fields it reads are `content-type`, `cache-control`, `x-robots-tag`, `x-metadata`, plus `cacheControl` / `metadata` / `contentType` form fields.
- Even the S3-compat `PutObject` route extracts `ContentEncoding` from headers but then drops it inside `s3Protocol.putObject` (`src/storage/protocols/s3/s3-handler.ts`) before calling `Uploader.upload`. The backend `uploadObject` adapter signature has no slot for it.

So as-is, this PR would advertise an option that has no effect, callers would assume their object has `Content-Encoding: gzip` stored when in fact it's silently dropped, which is worse than the current state.

Would you be up for opening a feature request on https://github.com/supabase/storage proposing that the standard upload route (and ideally the S3-compat route too) persist `Content-Encoding` into the object metadata and return it on read? Once the server side is in, this client PR becomes a small, safe follow-up and we can revisit it. Happy to revisit this one as soon as that lands.

A couple of smaller notes for when we do revisit:
- The Copilot review flagged that the TSDoc on `contentEncoding` says "the `Content-Encoding` header value", but for Blob/FormData uploads it's actually sent as a multipart form field, worth tweaking the wording to describe both transports.
- Also flagged a missing test for the FormData passthrough de-dup branch (`!body.has('contentEncoding')`), mirroring the existing `cacheControl` de-dup test.

Also FYI, this PR targets `develop`. I changed the default branch, and it should be `master`. So can you please recreate the PR to target `master`?

You can do this very easily.

1. Check out `master` on your local
2. Fetch and pull latest changes
3. Go on a new branch
4. `git cherry-pick 0df0466c9a71f8dffb981838a141135c51f52a2c` to add your commit

PR #2349, opened 2026-05-10
fix(postgrest): then typing

### LINKED ISSUE #1642: .throwOnError() only adjusts types correctly if await is used and is still incorrect if you use `then`
# Bug report

- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

awaiting a query with `.throwOnError()` correctly eliminates `null` from the typing. However, if you don't await it and use `.then(...)` instead, then `null` is not eliminated from the typing.

## To Reproduce

```
supabase.from('users').select('id').eq('id', id).single().throwOnError().then(({ data }) => { ... });
```

## Expected behavior

Type of `data` should be `{ id: string }`

## Actual behavior

Type of `data` is `{ id: string } | null`

7ttp:
## TL;DR

fixes throwOnError typing in `.then()` chains so success responses no longer stay nullable after `.throwOnError()`
also added a small test for it

## ref:

- closes https://github.com/supabase/supabase-js/issues/1642

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -100,6 +100,7 @@ export default class PostgrestFilterBuilder<
   RelationName = unknown,
   Relationships = unknown,
   Method = unknown,
+  ThrowOnError extends boolean = false,
 > extends PostgrestTransformBuilder<
   ClientOptions,
   Schema,
@@ -107,8 +108,31 @@ export default class PostgrestFilterBuilder<
   Result,
   RelationName,
   Relationships,
-  Method
+  Method,
+  ThrowOnError
 > {
+  throwOnError(): PostgrestFilterBuilder<
+    ClientOptions,
+    Schema,
+    Row,
+    Result,
+    RelationName,
+    Relationships,
+    Method,
+    true
+  > {
+    return super.throwOnError() as PostgrestFilterBuilder<
+      ClientOptions,
+      Schema,
+      Row,
+      Result,
+      RelationName,
+      Relationships,
+      Method,
+      true
+    >
+  }
+
   /**
    * Match only rows where `column` is equal to `value`.
    *
@@ -1770,7 +1794,8 @@ export default class PostgrestFilterBuilder<
     NarrowResultColumn<Result, ColumnName>,
     RelationName,
     Relationships,
-    Method
+    Method,
+    ThrowOnError
   > &
     this
   not<ColumnName extends string & keyof Row>(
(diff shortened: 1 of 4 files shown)


mandarini:
@7ttp are you still working on this?

7ttp:
mb almost forgot abt this, will have a look at it today...

packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
```diff
@@ -100,15 +100,39 @@ export default class PostgrestFilterBuilder<
...
   Schema,
   Row,
   Result,
   RelationName,
   Relationships,
-  Method
+  Method,
+  ThrowOnError
 > {
+  throwOnError(): PostgrestFilterBuilder<
```
7ttp, line 114: tried a few smaller patches first, but they broke other type checks 

this version is a bit broader, but it fixes the root issue behind `.throwOnError().then(...)` and 
also passed through relevant suites :D

7ttp:
cc @mandarini

mandarini:
Helloooo! A few small things before I merge:

1. In `test/index.test-d.ts`, the existing `expectType<typeof y extends typeof x ? true : false>(true)` got removed. The new `w extends y` check is great but tests a different thing. Could we replace the dropped assertion with a positive one that documents the new relationship?

2. The new test covers `.single().throwOnError().then(...)`. Could we add a couple more `.then()` shapes? The existing tests cover these for `await`, but `.then()` is a different code path:
   - `.maybeSingle().throwOnError().then(...)` (trickiest, should stay nullable)
   - `.throwOnError().single().then(...)` (throwOnError before single)
   - `.throwOnError().eq('x', 1).then(...)` (no single at all)

Wdyt?

7ttp:
> Helloooo! A few small things before I merge:
> 
>     1. In `test/index.test-d.ts`, the existing `expectType<typeof y extends typeof x ? true : false>(true)` got removed. The new `w extends y` check is great but tests a different thing. Could we replace the dropped assertion with a positive one that documents the new relationship?
> 
>     2. The new test covers `.single().throwOnError().then(...)`. Could we add a couple more `.then()` shapes? The existing tests cover these for `await`, but `.then()` is a different code path:
>        
>        * `.maybeSingle().throwOnError().then(...)` (trickiest, should stay nullable)
>        * `.throwOnError().single().then(...)` (throwOnError before single)
>        * `.throwOnError().eq('x', 1).then(...)` (no single at all)
> 
> 
> Wdyt?

ah! makes sense
 js swapped the dropped assertion for a positive x -> throwOnError(x) check and
 added the extra .then() cases you called out...

mandarini (approved):

PR #2367, opened 2026-05-12
fix(storage): make StreamDownloadBuilder implement Promise and memoize executor

### LINKED ISSUE #2366: Bug: StreamDownloadBuilder is missing Promise interface methods and re-runs executor on every await
### Bug

`packages/core/storage-js/src/packages/StreamDownloadBuilder.ts` implements `PromiseLike`, not `Promise`. As a result:

1. `.catch()` and `.finally()` are not defined on the builder, so common consumer patterns fail at compile time (in strict mode) or at runtime.
2. The builder is missing `[Symbol.toStringTag]`, so `Object.prototype.toString.call(builder)` returns `[object Object]` instead of `[object StreamDownloadBuilder]`.
3. `then` calls `this.execute()` directly, so the underlying `downloadFn` (a fetch call) runs once per `await`. Awaiting the same builder twice issues two HTTP requests.

The sibling `BlobDownloadBuilder.ts` already follows the correct pattern: it implements `Promise`, declares `[Symbol.toStringTag]`, defines `catch` and `finally`, and caches the in-flight promise in a private `getPromise()` helper. `StreamDownloadBuilder` should mirror that shape.

### Reproduction

```ts
const downloadFn = jest.fn(async () => new Response('hello'))
const builder = new StreamDownloadBuilder(downloadFn, false)

await builder
await builder
await builder

// expected: downloadFn called 1 time (cached)
// actual:   downloadFn called 3 times
```

```ts
const builder = new StreamDownloadBuilder(async () => new Response(''), false)
builder.catch(() => {})    // TS2339: Property catch does not exist
builder.finally(() => {})  // TS2339: Property finally does not exist
builder[Symbol.toStringTag] // undefined
```

### Expected

`StreamDownloadBuilder` should match the `Promise` contract used by `BlobDownloadBuilder`, so awaiting twice does not re-fetch, `.catch()` and `.finally()` work, and `Symbol.toStringTag` is set.

### Environment

- `@supabase/storage-js` current master
- Affects every caller of `.download(...).asStream()`

oniani1:
Closes #2366.

## Summary

`StreamDownloadBuilder` declared `PromiseLike` instead of `Promise`, so it lacked `catch`, `finally`, and `[Symbol.toStringTag]`. Its `then` also ran `execute()` on every call, which meant each `await` re-issued the underlying fetch.

This change rewrites the class to mirror the sibling `BlobDownloadBuilder`: it now implements `Promise<DownloadResult<ReadableStream>>`, caches the in-flight promise in a private `getPromise()`, and adds the missing `catch`, `finally`, and `Symbol.toStringTag` members. The constructor signature, exports, and `execute()` body are unchanged.

## Test plan

Added `test/downloadBuilders.test.ts` covering five behaviors on the stream builder (single executor call across multiple awaits, `.catch()`, `.finally()`, `Symbol.toStringTag`, `Promise` assignability), plus two regression checks on `BlobDownloadBuilder`. All seven pass.

### DIFF
--- a/packages/core/storage-js/src/packages/StreamDownloadBuilder.ts
+++ b/packages/core/storage-js/src/packages/StreamDownloadBuilder.ts
@@ -1,7 +1,10 @@
 import { isStorageError } from '../lib/common/errors'
 import { DownloadResult } from '../lib/types'
 
-export default class StreamDownloadBuilder implements PromiseLike<DownloadResult<ReadableStream>> {
+export default class StreamDownloadBuilder implements Promise<DownloadResult<ReadableStream>> {
+  readonly [Symbol.toStringTag]: string = 'StreamDownloadBuilder'
+  private promise: Promise<DownloadResult<ReadableStream>> | null = null
+
   constructor(
     private downloadFn: () => Promise<Response>,
     private shouldThrowOnError: boolean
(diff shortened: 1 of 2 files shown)


mandarini (approved):
Thank you for this @oniani1

PR #2368, opened 2026-05-13
chore(repo): migrate to pnpm

mandarini:
## Description

Switches the workspace package manager from npm to pnpm, primarily as a structural fix for the supply-chain attack class that PR #2206 patched with `--ignore-scripts` as a band-aid. pnpm exposes a project-level whitelist for install scripts that npm does not, so we can prevent unaudited postinstall hooks from executing at all.

Zero breaking changes for consumers. Published tarballs are byte-equivalent: `workspace:*` is swapped to concrete versions by `pnpm publish` at release time (verified locally with `pnpm pack`).

### What changed

Root setup
- Added `pnpm-workspace.yaml` listing workspace members (six core packages plus the auth-js and realtime-js example apps) and all pnpm settings: `nodeLinker: hoisted`, `preferWorkspacePackages: true`, plus the supply-chain hardening below.
- Pinned the package manager via `"packageManager": "pnpm@11.1.1+sha512..."` in the root `package.json`.
- Added a `preinstall` script (`npx only-allow pnpm`) that aborts any attempt to install with npm or yarn, preventing stray lockfiles that would bypass the new policies.
- Deleted `.npmrc` and the root `package-lock.json`. The two example-app `package-lock.json` files are also deleted; they consume sibling packages via `workspace:*` now.
- Switched the five internal `@supabase/*` dependencies in `packages/core/supabase-js/package.json` from `"*"` to `"workspace:*"`.

Supply-chain hardening (all in `pnpm-workspace.yaml`)
- `allowBuilds`: default-deny whitelist for postinstall scripts. Four packages whose postinstalls are load-bearing are allowed (`@swc/core`, `esbuild`, `nx`, `supabase`). Everything else, including transitively-pulled packages with native bindings, cannot run scripts.
- `blockExoticSubdeps: true`: refuses `file:`, `git:`, and `http:` URLs in transitive deps.
- `minimumReleaseAge: 10080` (7 days) with `minimumReleaseAgeExclude: ['@supabase/*']`: any third-party package published less than a week ago is refused. Our own releases install immediately downstream.

CI workflows
- `ci-core.yml`, `ci-supabase-js.yml`, `publish.yml`, `integration-tests.yml`, `docs.yml`, `preview-release.yml`: switched `actions/setup-node` cache to `pnpm`, added `pnpm/action-setup@v6.0.5`, replaced every `npm ci` with `pnpm install --frozen-lockfile`, every `npm run X` with `pnpm run X`, every `npx nx` with `pnpm nx`. Kept the env-var hardening from #2357 (uses `$VERSION_SPECIFIER` / `$BETA_VERSION` from `env:` blocks instead of inline `${{ github.event.inputs.* }}` interpolation).
- Added `pnpm audit --audit-level=high --prod` step to the `build` job in `ci-core.yml` to catch known CVEs at PR time.
- Rewrote `fix-lockfile.yml` to regenerate `pnpm-lock.yaml` via `pnpm install --lockfile-only` when Dependabot touches the lockfile.
- The standalone test apps (next, expo, deno, bun) install via `pnpm install --no-frozen-lockfile --ignore-scripts --ignore-workspace` because they install supabase-js from `pkg.pr.new` and intentionally behave like external consumers.

Workspace structure
- `nx.json`: `sharedGlobals` now hashes `pnpm-lock.yaml` and `pnpm-workspace.yaml` instead of `package-lock.json`.
- `packages/core/auth-js/project.json` and `packages/core/realtime-js/project.json`: the e2e targets use `pnpm install --no-frozen-lockfile` and `pnpm exec playwright ...`.
- `packages/core/auth-js/example/react/package.json` and `packages/core/realtime-js/example/package.json`: switched `"@supabase/auth-js": "file:../.."` / `"@supabase/supabase-js": "file:../../supabase-js"` to `workspace:*`.

Husky and scripts
- `.husky/commit-msg`, `.husky/pre-commit`, `.husky/pre-push`: `npm run` and `npx` replaced with `pnpm run` and `pnpm exec` / `pnpm nx`.
- Release scripts (`release-canary.ts`, `release-beta.ts`, `release-stable.ts`, `publish-to-jsr.ts`, `generate-postgrest-types.js`, `sync-common-types.js`): same substitutions; usage strings now read `pnpm run release-...`.

Docs
- `CLAUDE.md`, `CONTRIBUTING.md`, `docs/TESTING.md`, `docs/MIGRATION.md`, all six package READMEs, `packages/core/supabase-js/TESTING.md`, `.github/PULL_REQUEST_TEMPLATE.md`: replaced developer-facing `npm`/`npx` references with `pnpm` equivalents.
- End-user install instructions (`npm install @supabase/supabase-js` in READMEs and RELEASE.md) are intentionally left as-is.
- "Old workflow" columns in `MIGRATION.md` comparison tables are intentionally left as-is (historical context).
- Added a "Browser tests (optional)" note in `CONTRIBUTING.md` explaining how to install Chromium locally now that `puppeteer: false` in `allowBuilds`.

On 2026-03-30 / 31, a supply-chain attack compromised `axios@1.14.1` on npm via a malicious `postinstall` hook injected via a transitive dependency (`plain-crypto-js@4.2.1`). #2206 added `--ignore-scripts` to the affected installs, but that fix is per-command and easy to forget. It also suppresses legitimate scripts (esbuild binary downloads, etc.) wholesale.

`pnpm.allowBuilds` is a persistent, project-level policy. Any package not on the whitelist cannot run install scripts, regardless of flags. The axios-style attack would have been a non-event under this configuration. Other wins:

- `pnpm-lock.yaml` enforces content hashes more strictly than npm's lockfile.
- `workspace:*` makes the internal dep graph explicit in the lockfile.
- `pnpm`'s content-addressable global store reduces CI install times and local disk usage.
- Nx has first-class pnpm support; all existing `nx` commands continue to work unchanged.

## Breaking changes

- [x] This PR contains no breaking changes.

Verified: `pnpm pack` from `packages/core/supabase-js` produces a tarball whose `package.json` dependencies are resolved to concrete versions, not `workspace:*` literals. `pnpm nx release --dry-run` runs cleanly, detects pnpm, and respects the fixed-versioning relationship.

## Checklist

- [x] I have read the Contributing Guidelines.
- [x] My PR title follows the conventional commit format.
- [x] I have run `pnpm nx format` to ensure consistent code formatting.
- [ ] I have added tests for new functionality (N/A: tooling-only change).
- [x] I have updated documentation.

## Additional notes

- `nodeLinker: hoisted` is deliberate. Core packages do not declare their per-package devDeps (vitest, tsdown, jest, etc.) individually; they rely on hoisting from the root. Switching to pnpm's default strict isolation would break Nx-orchestrated builds. Tightening per-package devDeps and flipping to strict is a follow-up.
- The `test/integration/next`, `test/integration/expo`, `test/integration/bun`, and `test/deno` apps are intentionally NOT in `pnpm-workspace.yaml`. They install supabase-js from `pkg.pr.new` URLs to simulate real consumer installs. CI uses `--ignore-workspace` for these so pnpm does not absorb them into the parent workspace.
- `pnpm audit` in CI will fail on high or critical CVEs in production deps. First push may surface latent issues; common remediation is `pnpm update <pkg>` or `pnpm audit --fix`.


staaldraad (dismissed):

pnpm-workspace.yaml
```diff
@@ -0,0 +1,32 @@
+packages:
+  - 'packages/core/*'
+  - 'packages/shared/*'
+  - 'packages/core/auth-js/example/react'
+  - 'packages/core/realtime-js/example'
+
+# Resolver / layout
+nodeLinker: hoisted # flat node_modules; core packages depend on hoisting (no explicit per-package devDep declarations yet)
```
jgoux, line 8: I would avoid the hoisted mode and make every package explicit about its dependencies to avoid phantom dependencies.

I you need to easily define one version of a dev dependency for all the packages you can use [catalogs](https://pnpm.io/catalogs#the-catalog-protocol-catalog) for that. 👍

jgoux (approved):

pnpm-workspace.yaml
mandarini: You're right, this is a deliberate compromise documented in the inline comment. The migration kept nodeLinker: hoisted because the six core packages don't currently declare their dev deps individually, they are all pulled from the workspace root, and strict isolation would break their build/test pipelines.                                                       
I'll make the proper fix (per-package devDep declarations + drop hoisting + use catalog: protocol to keep versions consistent) in a follow up PR, where I'll fix a number of other deps!! Good call, adding it to my list of todos!

PR #2382, opened 2026-05-19
docs(repo): safeguard against supply chain attacks

mandarini:
Add user guide on how users can safeguard against npm supply chain attacks

### DIFF
--- a/docs/NPM_PACKAGE_SECURITY.md
+++ b/docs/NPM_PACKAGE_SECURITY.md
@@ -15,7 +15,7 @@ The good news: most of the impact is preventable from the consumer side, regardl
 1. **Commit your lockfile** and install with `--frozen-lockfile` (pnpm/yarn) or `npm ci` (npm) in CI.
 2. **Quarantine new versions.** Set a minimum release age (≥ 7 days) so installs won't pick up a brand-new version while attackers are still in their detection window. npm, pnpm, yarn, and bun all support this natively now.
 3. **Block exotic transitive deps.** Refuse `github:`, `git+`, and `file:` refs that didn't come from the npm registry.
-4. **Verify provenance.** Run `npm audit signatures` after install. Supabase packages publish with sigstore attestations.
+4. **Verify provenance.** Run `npm audit signatures` after install. Supabase packages publish with [sigstore attestations](https://www.sigstore.dev/) (cryptographic proof tying each tarball to the workflow run, commit, and repo it was built from).
 5. **Constrain lifecycle scripts.** Default-deny `postinstall` / `preinstall` / `prepare`; allow per-package.
 6. **Pin a single source of truth for your package manager** (`packageManager` field in `package.json` with a sha512 hash).
 7. **Have a rollback plan.** Know which packages, which versions, and which credentials to rotate if a compromise is announced upstream.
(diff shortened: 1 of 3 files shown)


docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,323 @@
...
+The same attack pattern keeps recurring: a popular npm package is compromised, the new version executes attacker code on `npm install` via a lifecycle script or transitive dependency, the malware harvests credentials from the install host, and then self-propagates by republishing other packages the victim maintains.
+
+The good news: most of the impact is preventable from the consumer side, regardless of what any one publisher does. This guide is the set of settings and habits we recommend you adopt.
+
+## TL;DR — what to do today
+
+1. **Commit your lockfile** and install with `--frozen-lockfile` (pnpm/yarn) or `npm ci` (npm) in CI.
+2. **Quarantine new versions.** Set a minimum release age (≥ 7 days) so installs won't pick up a brand-new version while attackers are still in their detection window. npm, pnpm, yarn, and bun all support this natively now.
+3. **Block exotic transitive deps.** Refuse `github:`, `git+`, and `file:` refs that didn't come from the npm registry.
+4. **Verify provenance.** Run `npm audit signatures` after install. Supabase packages publish with sigstore attestations.
```
staaldraad: a link to what sigstore attestations are would be useful here.

docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,323 @@
...
+minimumReleaseAge: 10080 # 7 days, in minutes
+minimumReleaseAgeExclude:
+  - '@your-org/*' # bypass for your own internal packages
+```
+
+Set `minimumReleaseAge: 0` only if you have a specific reason to opt out of the default.
+
+#### `trustPolicy` (pnpm)
+
+Independent of the age gate, pnpm's `trustPolicy: no-downgrade` refuses to install a version whose trust level (trusted publisher → provenance → none) has dropped relative to previous releases of the same package. That catches the case where an attacker can publish but can't replicate the original maintainer's OIDC binding:
```
staaldraad, line 78: nice :+1:

docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,323 @@
...
+
+```ini
+registry=https://npm.your-org.example/
+@supabase:registry=https://npm.your-org.example/
+//npm.your-org.example/:_authToken=${NPM_TOKEN}
+always-auth=true
+```
+
+Notes:
+
```
staaldraad, line 167: if using GitHub, enable secret protection on public repos (free)

docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,323 @@
...
+- Never commit the token. Use `${NPM_TOKEN}` from your CI environment (npm 7+ expands env vars in `.npmrc`).
```
staaldraad, line 168: use short token lifetimes. It means having to update the token more often, but it limits  compromised tokens being silently used for a long time in the case of an undetected leak.

docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,323 @@
...
+
+`@supabase/supabase-js`, `@supabase/auth-js`, `@supabase/postgrest-js`, `@supabase/realtime-js`, `@supabase/storage-js`, and `@supabase/functions-js` publish with [sigstore provenance attestations](https://docs.npmjs.com/generating-provenance-statements) via npm OIDC trusted publishing. The attestations cryptographically tie each published tarball to the workflow run, commit, and repository it was built from.
+
+To verify after install:
+
+```bash
+npm audit signatures
+```
+
+Sample output:
```
staaldraad, line 186: should we say something like "Supabase's attestation will always resolve to repository under the Supabase organisation on GitHub"

staaldraad (dismissed):

docs/NPM_PACKAGE_SECURITY.md
mandarini: Yes

staaldraad (approved):

jgoux (approved):

docs/NPM_PACKAGE_SECURITY.md
```diff
@@ -0,0 +1,327 @@
...
+
+```bash
+bun add @supabase/supabase-js --minimum-release-age 604800
+```
+
+Bun's age gate only affects new resolutions — existing entries in `bun.lock` are unchanged. It also runs a stability check: if multiple versions were published close together just outside your gate, Bun extends the filter to skip those (likely unstable) versions and picks an older, more mature one. Exact-version requests (`pkg@1.1.1`) respect the gate but bypass the stability extension.
+
+### Deno / Edge Functions
+
+Deno has an unstable `--minimum-dependency-age` flag that accepts minutes, an ISO-8601 duration, or an absolute RFC3339 cutoff:
```
laktek, line 140: This wouldn't work with Edge Functions and also an unstable feature. Maybe we should leave this out?

mandarini: Ok, I'll put in a follow up to remove. We do acknowledge at least that it's unstable 😓

PR #2387, opened 2026-05-21
refactor(auth): remove navigator.locks-based mutex; introduce commit guard + dispose()

Bewinxed:
### What changed?

`GoTrueClient` no longer uses a shared mutex to serialize auth operations. `_acquireLock`, its `pendingInLock` queue, and all 14 call-site wrappers are gone. In their place:

- `refreshingDeferred` (already existed) keeps single-flighting in-instance concurrent refreshes.
- A commit guard inside `_callRefreshToken` re-reads storage after the rotated-token response and before `_saveSession`. If the input `refresh_token` isn't in storage anymore (because a concurrent `signOut` ran `_removeSession` between fetch start and continuation), the rotated tokens are discarded and the result resolves to `{ data: null, error: new AuthRefreshDiscardedError() }`. `_recoverAndRefresh` recognises this error and skips its own `_removeSession()` call, so no duplicate `SIGNED_OUT` event fires.
- A new exported error class, `AuthRefreshDiscardedError` (with an `isAuthRefreshDiscardedError` type guard), is what the commit guard returns when it throws away a successfully-rotated token. It signals "the server gave us new tokens but the client decided not to keep them," distinct from `AuthRetryableFetchError` (transient network) and `AuthApiError` (server rejection). Surfaces through `refreshSession()` and `getSession()` results.
- A new `client.dispose()` tears down the auto-refresh interval, the `visibilitychange` listener, the `BroadcastChannel`, and registered `onAuthStateChange` subscribers. Idempotent. Designed for React Strict Mode and HMR cleanup hooks.
- GoTrue handles cross-tab refresh races server-side via the v1 parent-of-active path at `internal/tokens/service.go:376-385`, so the client doesn't need to coordinate. The current default is v1 (`RefreshTokenAlgorithmVersion = 0`); v2 (counter-based) is opt-in via `RefreshTokenUpgradePercentage` and gated rollout. This PR's reasoning relies on v1 behaviour, which is what customers actually run today.
- The `lock` and `lockAcquireTimeout` constructor options are accepted but silently ignored for backwards compatibility. Both are now `@deprecated` in JSDoc, including the `@supabase/supabase-js` re-exported fields.
- `navigatorLock`, `processLock`, `LockAcquireTimeoutError`, `NavigatorLockAcquireTimeoutError`, `ProcessLockAcquireTimeoutError`, and `internals` in `lib/locks.ts` remain exported for one major version. All `@deprecated`.
- Stale JSDoc referencing the lock on `getSession`, `onAuthStateChange`, `_useSession`, `_challengeAndVerify`, and `_listFactors` now matches the new behaviour. The `@deprecated` tag on the async `onAuthStateChange` overload (its only justification was the deadlock that no longer exists) is removed.

The shared mutex has caused seven failure classes on the issue tracker for 18+ months. Each earlier patch fixed one symptom and created another:

- [#1962](https://github.com/supabase/supabase-js/pull/1962) configurable timeout
- [#2106](https://github.com/supabase/supabase-js/pull/2106) steal fallback
- [#2178](https://github.com/supabase/supabase-js/pull/2178) steal-cascade guard

[#2235](https://github.com/supabase/supabase-js/pull/2235) (primitive swap to `processLock`) was an earlier attempt at moving off `navigator.locks` and is no longer being actively pursued. Community PRs [#2016](https://github.com/supabase/supabase-js/pull/2016) and [#2019](https://github.com/supabase/supabase-js/pull/2019) fixed individual symptoms and were closed waiting on a structural fix.

Failure classes resolved by this PR:

1. Strict Mode / HMR orphan deadlocks. Nothing to orphan now that the lock is gone.
2. iOS Safari `INITIAL_SESSION` / `getSession()` hangs with a persisted session ([#936 @bugprone](https://github.com/supabase/supabase-js/issues/936#issuecomment-3691211975)). The client no longer touches `navigator.locks`.
3. Sentry flooding with `AbortError: Lock broken by another request with the 'steal' option` ([#2013 @cpannwitz](https://github.com/supabase/supabase-js/issues/2013#issuecomment-4024042017)). The `{ steal: true }` fallback that produced these errors is gone with the lock.
4. Subscriber re-entry deadlock (the one at the old `GoTrueClient.ts:3824`). The cycle needed the lock; a callback that awaits `getUser()` now just runs.
5. Visibility/tick race ([#936 @JTBrinkmann](https://github.com/supabase/supabase-js/issues/936#issuecomment-3728405087)). The visibility handler and the auto-refresh tick no longer share a lock to fight over.
6. Stale tickers/listeners on un-disposed clients. `dispose()` cleans them up.
7. `signOut` blocked behind in-flight refresh. signOut now runs concurrently with the refresh, and the commit guard prevents the refresh from writing rotated tokens after `_removeSession()` cleared storage.

Why no lock at all: cross-tab races are handled on the server (GoTrue's parent-of-active), in-instance refresh dedup is already handled by `refreshingDeferred`, and the only job left for the lock was serializing subscriber callbacks. That's the deadlock this PR fixes. A narrower lock would still do that one job.

Closes (test target, not for merge yet): #2013, #936, #2111

Related: `lockless_auth_coordination` RFC (to be opened separately).

### `dispose()` in a React app

```ts
import { useEffect } from 'react'

export function useSupabase() {
  useEffect(() => {
    const supabase = createClient(URL, KEY)
    return () => {
      supabase.auth.dispose()
    }
  }, [])
}
```

### Subscriber callbacks can call other auth methods now

```ts
// Previously: deadlocked. Callback held the lock; getUser tried to acquire it.
// Now: works.
supabase.auth.onAuthStateChange(async (event, session) => {
  if (event === 'SIGNED_IN') {
    const { data } = await supabase.auth.getUser()
    // ...
  }
})
```

### `AuthRefreshDiscardedError` for the signOut-during-refresh race

```ts
import { isAuthRefreshDiscardedError } from '@supabase/auth-js'

const { data, error } = await supabase.auth.refreshSession()
if (isAuthRefreshDiscardedError(error)) {
  // A concurrent signOut cleared storage between when this refresh
  // started and when it came back. The rotated tokens were discarded.
  // The SIGNED_OUT event already fired from the concurrent signOut.
}
```

## Breaking changes

### Behaviour changes worth flagging for review

- The JSDoc on `onAuthStateChange` no longer warns against async callbacks. Calling other auth methods inside the callback used to deadlock; now it works. The `@deprecated` marker on the async overload signature is removed (its only justification was the deadlock).
- `lock` and `lockAcquireTimeout` are accepted but ignored. A custom `lock` function passed via constructor is never invoked. Code that depended on a custom lock being called will silently lose that behaviour. The `@deprecated` JSDoc on both options flags this for the developer.
- Subscriber timing: subscribers stay awaited; same as before. What changes is that `signOut` no longer waits for an in-flight refresh's HTTP and continuation to finish before its own fetch goes out. Both fetches now run concurrently, and the commit guard keeps storage consistent.

## Checklist

- [x] I have added tests for new functionality (commit guard, `dispose()`, subscriber re-entry no-deadlock, custom lock no-op)
- [x] I have updated documentation (JSDoc on affected methods, type-level deprecation notices)

### Alternatives considered

**Continuing to patch `navigator.locks`** (better steal recovery, lower timeouts, smarter error filtering). Each prior patch in this direction (#1962, #2106, #2178) fixed one symptom and produced another. The Web Locks API has no recovery for orphaned holders other than `{ steal: true }`, which leaves the previous `fn()` running concurrently with the new holder and creates the steal-cascade error storm. Each fix here swaps one failure for another instead of removing the contention that causes them.

**Swapping `navigatorLock` for `processLock` as the default browser lock** (the direction #2235 explored). Removes the cross-process orphan failure but keeps the same shape of the bug: one primitive serializing operations that don't all need the same synchronization. Subscriber re-entry deadlock, visibility-tick contention, init-tick contention, and in-process orphans of the new primitive all remain.

**Non-blocking subscriber notifications alone** (#2016). Fixes subscriber re-entry. The other six failure classes still bite.

**A "smaller, refactored" lock with narrower scope** (e.g. lock only around storage writes, not around HTTP). Tempting on paper. Cross-tab refresh is already lockless on the server (parent-of-active returns the same rotated token to N tabs), and `refreshingDeferred` already single-flights refreshes inside a tab. The visibility-tick race is the lock fighting itself. The only thing a narrower lock could still do is serialize subscriber callbacks within a tab, which is the deadlock this PR fixes. No smaller version of this PR still fixes the seven failures above.

**Adding an `AbortController` layer for in-flight operation cancellation.** Deferred to a follow-up, not included here. The commit guard catches the one race that affects correctness (signOut overwriting cleared storage with rotated tokens). AbortController would be a UX improvement on top: cancel the in-flight refresh as soon as signOut runs, instead of letting it finish and discarding the result. It also requires plumbing `signal` through `lib/fetch.ts` and a decision on `AbortSignal.any` runtime support (Node ≥ 20.3, Hermes ≥ 0.74, Safari ≥ 17.4). Out of scope for v1.

**Deferring the commit guard and accepting eventual consistency.** Considered and rejected. Without a guard, a refresh that completes after `_removeSession()` cleared storage will write the rotated tokens back. Subscribers then see `SIGNED_OUT` followed by `TOKEN_REFRESHED`, with stale tokens in storage until the next refresh tick (~30s) fails against the server and clears them. The guard is about 15 lines. Better to land it with this PR than to ship the bug and patch it later.

### Server-side context

Why cross-tab is safe: GoTrue's parent-of-active path at `internal/tokens/service.go:376-385` (the v1 branch, `*models.RefreshToken`). When a request arrives with a revoked refresh token whose child is the currently-active token, the server returns the active token instead of rejecting. Two tabs that POST the same refresh token concurrently both receive the same rotated token under DB row locking. This is the production-default behaviour (`RefreshTokenAlgorithmVersion = 0`, v1). v2 (counter-based, gated on `RefreshTokenUpgradePercentage`) is safe under the same N-tab concurrency, covered by `TestConcurrentReuse`.

### Open questions for review

1. **Should `_setSession`'s expired branch also filter `AuthRefreshDiscardedError`?** Currently it propagates the error to the caller, which I think is correct (`setSession` was given new tokens; if a concurrent signOut just ran, the caller should be told). Confirming we're OK with that.
2. **`AbortSignal.any` plumbing in `lib/fetch.ts`** isn't done here (no AbortController layer to plumb yet). If we add the AbortController follow-up later, this becomes a prerequisite. Just so it's on the radar.
3. **`@supabase/supabase-js` types deprecation**: I marked `lock` and `lockAcquireTimeout` `@deprecated` in the supabase-js types file too, since they delegate to the auth client. Confirm that's the right surface to deprecate.

### Test coverage

New tests in `packages/core/auth-js/test/GoTrueClient.test.ts`:

- `'Lockless coordination'` describe: deprecated `lock` and `lockAcquireTimeout` options accepted-but-ignored; subscriber callback may call other auth methods without deadlock.
- `'dispose() lifecycle'` describe: idempotency, subscriber clearing, ticker stopping.
- `'Refresh commit guard (signOut-during-refresh race)'` describe: simulates the race, asserts `AuthRefreshDiscardedError` is returned and storage stays cleared.

New tests in `packages/core/auth-js/test/GoTrueClient.browser.test.ts`:

- `'Lockless coordination: navigator.locks should NOT be invoked'`: confirms the browser API is no longer touched.
- `'Lockless backwards-compatibility: deprecated lock option'`: a custom `lock` function passed via constructor is never invoked.

Removed: `'Lock functionality'`, `'Browser locks functionality'`, `'Lock Mechanism Branches'` (tested a mechanism that no longer exists). Updated the existing `'should handle custom lock implementation'` test to assert `not.toHaveBeenCalled()` instead of `toHaveBeenCalled()`.

### Status

This is on a test branch for now. Ready for review and discussion. Once the design is signed off, I'll open it against `develop`.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -16,11 +16,13 @@ import {
   AuthInvalidTokenResponseError,
   AuthPKCECodeVerifierMissingError,
   AuthPKCEGrantCodeExchangeError,
+  AuthRefreshDiscardedError,
   AuthSessionMissingError,
   AuthUnknownError,
   isAuthApiError,
   isAuthError,
   isAuthImplicitGrantRedirectError,
+  isAuthRefreshDiscardedError,
   isAuthRetryableFetchError,
   isAuthSessionMissingError,
 } from './lib/errors'
@@ -54,7 +56,6 @@ import {
   validateExp,
 } from './lib/helpers'
 import { memoryLocalStorageAdapter } from './lib/local-storage'
-import { LockAcquireTimeoutError, navigatorLock } from './lib/locks'
 import { polyfillGlobalThis } from './lib/polyfills'
 import { version } from './lib/version'
 
@@ -91,7 +92,6 @@ import type {
   JWK,
   JwtHeader,
   JwtPayload,
-  LockFunc,
   MFAChallengeAndVerifyParams,
   MFAChallengeParams,
   MFAChallengePhoneParams,
@@ -183,7 +183,7 @@ polyfillGlobalThis() // Make "globalThis" available
 
 const DEFAULT_OPTIONS: Omit<
   Required<GoTrueClientOptions>,
-  'fetch' | 'storage' | 'userStorage' | 'lock'
+  'fetch' | 'storage' | 'userStorage' | 'lock' | 'lockAcquireTimeout'
 > = {
   url: GOTRUE_URL,
   storageKey: STORAGE_KEY,
@@ -195,13 +195,8 @@ const DEFAULT_OPTIONS: Omit<
   debug: false,
   hasCustomAuthorizationHeader: false,
   throwOnError: false,
-  lockAcquireTimeout: 5000, // 5 seconds
   skipAutoInitialize: false,
   experimental: {},
-}
-
-async function lockNoOp<R>(name: string, acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
-  return await fn()
 }
 
 /**
@@ -297,11 +292,7 @@ export default class GoTrueClient {
   protected hasCustomAuthorizationHeader = false
   protected suppressGetSessionWarning = false
   protected fetch: Fetch
-  protected lock: LockFunc
-  protected lockAcquired = false
-  protected pendingInLock: Promise<any>[] = []
   protected throwOnError: boolean
-  protected lockAcquireTimeout: number
   /**
    * Opt-in flags for experimental features. Defaults to an empty object.
    * See `GoTrueClientOptions.experimental`.
@@ -371,20 +362,13 @@ export default class GoTrueClient {
     this.url = settings.url
     this.headers = settings.headers
     this.fetch = resolveFetch(settings.fetch)
-    this.lock = settings.lock || lockNoOp
     this.detectSessionInUrl = settings.detectSessionInUrl
     this.flowType = settings.flowType
     this.hasCustomAuthorizationHeader = settings.hasCustomAuthorizationHeader
     this.throwOnError = settings.throwOnError
-    this.lockAcquireTimeout = settings.lockAcquireTimeout
 
-    if (settings.lock) {
-      this.lock = settings.lock
-    } else if (this.persistSession && isBrowser() && globalThis?.navigator?.locks) {
-      this.lock = navigatorLock
-    } else {
-      this.lock = lockNoOp
-    }
+    // settings.lock and settings.lockAcquireTimeout are typed for backwards
+    // compatibility and have no effect; deliberately not read here.
 
     if (!this.jwks) {
       this.jwks = { keys: [] }
@@ -518,9 +502,7 @@ export default class GoTrueClient {
     }
 
     this.initializePromise = (async () => {
-      return await this._acquireLock(this.lockAcquireTimeout, async () => {
-        return await this._initialize()
-      })
+      return await this._initialize()
     })()
 
     return await this.initializePromise
@@ -1405,9 +1387,7 @@ export default class GoTrueClient {
   async exchangeCodeForSession(authCode: string): Promise<AuthTokenResponse> {
     await this.initializePromise
 
-    return this._acquireLock(this.lockAcquireTimeout, async () => {
-      return this._exchangeCodeForSession(authCode)
-    })
+    return this._exchangeCodeForSession(authCode)
   }
 
   /**
@@ -2444,9 +2424,7 @@ export default class GoTrueClient {
   async reauthenticate(): Promise<AuthResponse> {
     await this.initializePromise
 
-    return await this._acquireLock(this.lockAcquireTimeout, async () => {
-      return await this._reauthenticate()
-    })
+    return await this._reauthenticate()
   }
 
   private async _reauthenticate(): Promise<AuthResponse> {
@@ -2593,7 +2571,7 @@ export default class GoTrueClient {
    * - If the session's access token is expired or is about to expire, this method will use the refresh token to refresh the session.
    * - When using in a browser, or you've called `startAutoRefresh()` in your environment (React Native, etc.) this function always returns a valid access token without refreshing the session itself, as this is done in the background. This function returns very fast.
    * - **IMPORTANT SECURITY NOTICE:** If using an insecure storage medium, such as cookies or request headers, the user object returned by this function **must not be trusted**. Always verify the JWT using `getClaims()` or your own JWT verification library to securely establish the user's identity and access. You can also use `getUser()` to fetch the user object directly from the Auth server for this purpose.
-   * - When using in a browser, this function is synchronized across all tabs using the [LockManager](https://developer.mozilla.org/en-US/docs/Web/API/LockManager) API. In other environments make sure you've defined a proper `lock` property, if necessary, to make sure there are no race conditions while the session is being refreshed.
+   * - Cross-tab refresh races are handled by the GoTrue server (the rotated token from the first tab is returned to subsequent tabs via the parent-of-active mechanism), so no client-side serialization is needed.
    *
    * @example Get the session data
    * ```js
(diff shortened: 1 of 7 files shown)


packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -4678,6 +4555,29 @@ export default class GoTrueClient {
       if (error) throw error
       if (!data.session) throw new AuthSessionMissingError()
 
+      // Commit guard: storage may have changed under us between fetch start
+      // and now (usually because a concurrent signOut ran `_removeSession`).
+      // If the refresh token we just rotated isn't in storage anymore,
+      // discard the rotated tokens. Writing them back would undo what
+      // signOut just cleared.
+      const currentStored = (await getItemAsync(this.storage, this.storageKey)) as Session | null
+      if (!currentStored || currentStored.refresh_token !== refreshToken) {
```
mandarini, line 4564: Commit guard fires false-positives on empty storage.

To repro: fresh client, empty `memoryLocalStorageAdapter`, no signOut anywhere, mocked `_refreshAccessToken` returning rotated tokens without touching storage:

```ts
await client.setSession({ access_token: expiredAT, refresh_token: 'R-external' })
// returns: { data: { session: null }, error: AuthRefreshDiscardedError }
// error.message: "Refresh result discarded: session state changed mid-flight (e.g., concurrent signOut)"
```

Breaks `setSession` and `refreshSession({ refresh_token })` for any caller supplying an expired AT plus a refresh token that isn't already in storage. That's the entire SSR cookie-handoff / external-token-hydration flow.

Fix: snapshot storage *before* the fetch, only discard when a non-null snapshot got cleared.

```ts
const storedAtStart = await getItemAsync(this.storage, this.storageKey)
// ... _refreshAccessToken ...
const storedAfter = await getItemAsync(this.storage, this.storageKey)
const wasClearedMidFlight =
  storedAtStart !== null &&
  (storedAfter === null || storedAfter.refresh_token !== refreshToken)
```

Add a test for the empty-storage path. Your existing `'discards rotated tokens when storage was cleared mid-flight'` test plants R1 first, so it doesn't catch this.

packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -4978,58 +4878,83 @@ export default class GoTrueClient {
...
   }
 
+  /**
+   * Tears down the client's background work: stops the auto-refresh interval,
+   * removes the `visibilitychange` listener, closes the cross-tab
+   * `BroadcastChannel`, and clears registered `onAuthStateChange` subscribers.
+   *
+   * Call this from cleanup hooks when the client is being replaced before
+   * its JS realm is destroyed. React Strict Mode and HMR are the common
+   * cases. After dispose, network operations will reject. Storage reads
```
mandarini, line 4888: The `dispose()` docstring claims behavior the code doesn't have. `dispose()` stops the ticker, closes the broadcast channel, removes the visibility listener, clears subscribers. In-flight fetches still complete and `_saveSession` still writes. New `getSession`/`getUser` calls after `dispose()` work fine. Per our earlier alignment to defer AbortController plumbing, delete the sentence.

packages/core/auth-js/src/GoTrueClient.ts
```diff
@@ -3832,16 +3720,11 @@ export default class GoTrueClient {
...
-   *
-   * Receive a notification every time an auth event happens.
+   * Receive a notification every time an auth event happens. Callbacks can be
+   * `async` and can safely call other Supabase auth methods (`getUser`,
+   * `setSession`, etc.) from inside the callback.
    *
    * @param callback A callback function to be invoked when an auth event happens.
-   * @deprecated Due to the possibility of deadlocks with async functions as callbacks, use the version without an async function.
    */
   onAuthStateChange(callback: (event: AuthChangeEvent, session: Session | null) => Promise<void>): {
```
mandarini, line 3846: I think removing `@deprecated` on the async `onAuthStateChange` overload is premature. One reentry deadlock remains: a subscriber that calls `refreshSession` (or anything routing through `_callRefreshToken`) from inside a `TOKEN_REFRESHED` handler. `refreshingDeferred.resolve` happens *after* `_notifyAllSubscribers` returns:

```ts
await this._notifyAllSubscribers('TOKEN_REFRESHED', data.session)  // subscriber runs here
// ...
this.refreshingDeferred.resolve(result)  // resolved AFTER
```

Subscriber's inner `refreshSession` dedups onto `refreshingDeferred`, which won't resolve until the outer finishes awaiting subscribers. Deadlock.

Let's keep the `@deprecated` on the async overload, rewrite the reason to point at residual `refreshingDeferred` reentry instead of the old lock. The common cases (`getUser`, `setSession`, reading session from a callback) are genuinely fixed and worth saying.

mandarini (changes requested):
Left some comments in the code, and also:

1.  "No breaking changes" checkbox is incorrect. There are two real breaking changes.

- Custom `lock` functions are silently ignored. React Native users with `processLock` and Node multi-process setups with custom locks now race silently with no warning. Add a `console.warn` (gated on an opt-out) at construction when a non-null `lock` is supplied.
- `setSession`/`refreshSession({ refresh_token })` with externally-supplied tokens against empty storage now error. Goes away once #1 is fixed.

Let's update the PR description.

2. `_autoRefreshTokenTick` behavior change not flagged.

Old code: `_acquireLock(0, ...)` made the tick skip whenever *any* auth op held the lock. New code: only skips when `refreshingDeferred` is set. The tick can now fire concurrently with signOut/setSession/getUser/etc. for the first time. Add to "Behaviour changes worth flagging for review".

Side note: the early-return `if (this.refreshingDeferred !== null)` at the top of the tick duplicates dedup that `_callRefreshToken` already does. Keep as perf opt or drop, your call.

3. RFC references in code.

Code comments and test names reference `lockless_auth_coordination`. If the RFC isn't landing as a separate doc, scrub the references; otherwise link it.


Thank you so much @Bewinxed !!! :D

Bewinxed:
Addressed in #2392 — see [the summary comment](https://github.com/supabase/supabase-js/pull/2392) for the per-point response.

Bewinxed:
Closed and continued in #2392 — rebased onto `master`.

PR #2392, opened 2026-05-21
feat(auth): remove navigator.locks-based mutex; introduce commit guard + dispose()

### LINKED ISSUE #879: TypeError: Network request failed on sign up / sign in on Android
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Signing in or up on Android emulator locally results in `TypeError: Network request failed`. This error only shows up on the Android emulator and not the iOS one

## To Reproduce

Minimal reproducible example: https://github.com/maxwowo/sbrepro

Environment:
* Java: `openjdk 11.0.19 2023-04-18 LTS`
* Node: `v16.20.2`

1. Clone repo
2. `yarn install`
3. `yarn android`
4. Click on either `Sign up` or `Sign in`

## Expected behavior

Should sign up or sign in without error

## Screenshots

https://github.com/supabase/supabase-js/assets/40416283/b85f24b7-3c1f-4f0a-ad56-484ce0fd4964

## System information

- OS: macOS Sonoma 14.0
- Browser (if applies) N/A
- Version of supabase-js: `2.38.1`
- Version of Node.js: `v16.20.2`

### LINKED ISSUE #1618: onAuthStateChange not triggered when using server-side signIn/signOut in Next.js with Supabase
# Bug report

## Describe the bug

When using the Supabase **server client** (from `@supabase/ssr`) to call `signIn` or `signOut` inside a Next.js `server action`, the **auth state changes are not detected** on the **browser client** (`createBrowserClient()`).

Specifically, `onAuthStateChange()` on the browser client does **not fire** when the sign-in or sign-out is done from a server action.

This leads to an inconsistent auth state in the frontend after using server-side auth methods.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Set up Supabase in a Next.js app using both `createServerClient()` and `createBrowserClient()`
2. Set up `onAuthStateChange()` in a layout or component using the browser client.
3. Trigger `signIn()` or `signOut()` using a **server action** via the server client.
4. Notice that the browser does not detect any change in auth state.

## Expected behavior

Calling `signIn()` or `signOut()` on the server should also update the browser-side auth state or trigger `onAuthStateChange()` accordingly, assuming the cookies are properly shared.

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

## Additional context

This behavior breaks real-time auth-dependent UI updates and makes sign-in/sign-out UX inconsistent.

### LINKED ISSUE #1620: Supabase Query Builder Hangs Indefinitely on Windows 11 + Vite + React 19
### Describe the bug

Supabase JS client query builder creates query objects successfully but the promise never resolves when awaited. No HTTP request is sent, and the query hangs indefinitely (no timeout, just infinite wait).

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

1. Create Supabase client:
```typescript
import { createClient } from '@supabase/supabase-js';

const supabaseUrl = import.meta.env.VITE_SUPABASE_URL;
const supabaseAnonKey = import.meta.env.VITE_SUPABASE_ANON_KEY;

export const supabase = createClient(supabaseUrl, supabaseAnonKey);
```

2. Attempt to query using standard pattern:
```typescript
const { data, error, count } = await supabase
  .from('events')
  .select('*', { count: 'exact' })
  .order('start_date', { ascending: true })
  .range(0, 19);
```

3. **Result**: Promise never resolves, query hangs indefinitely

### System Info

```shell
### Operating System
- **OS**: Microsoft Windows 11 Home
- **Version**: 10.0.26100

### Package Versions
- **@supabase/supabase-js**: 2.58.0
- **React**: 19.2.0
- **React DOM**: 19.2.0
- **Vite**: 7.1.9
- **Node.js**: v22.18.0
- **@vitejs/plugin-react**: 5.0.4

### Build Tool
- Vite 7.1.9 with TypeScript
- Target: ESNext
```

### Used Package Manager

npm

### What Happens
1. ✅ Query object builds successfully
2. ✅ Query object has correct URL property: `https://[project].supabase.co/rest/v1/events?select=*&order=start_date.asc&offset=0&limit=20`
3. ✅ Query object has correct method: `GET`
4. ❌ **Headers object is EMPTY**: `Headers {}`
5. ❌ `await query` hangs forever (no timeout, no error, no response)
6. ❌ **Network tab shows NO HTTP request is sent**
7. ❌ No errors in console
8. ❌ No network errors

### What Works
Direct `fetch()` to the EXACT same URL works perfectly:
```typescript
const directUrl = query.url?.toString();
const directResponse = await fetch(directUrl, {
  headers: {
    'apikey': import.meta.env.VITE_SUPABASE_ANON_KEY,
    'Authorization': `Bearer ${import.meta.env.VITE_SUPABASE_ANON_KEY}`,
    'Prefer': 'count=exact'
  }
});
const data = await directResponse.json(); // ✅ Works perfectly, returns data
```

### 1. Headers Object is Empty
When inspecting the query object before execution:
```typescript
console.log('Query headers:', query.headers);
// Output: Headers {}  <-- EMPTY, should contain apikey and authorization
```

### 2. Custom Fetch Test
Added custom fetch to client config to debug:
```typescript
const customFetch = (url, options) => {
  console.log('[Supabase] Custom fetch called:', url);
  return fetch(url, options);
};

export const supabase = createClient(supabaseUrl, supabaseAnonKey, {
  global: {
    fetch: customFetch,
  },
  auth: {
    autoRefreshToken: true,
    persistSession: true,
    detectSessionInUrl: true,
  },
});
```

**Result:**
- ✅ Custom fetch IS called for auth: `[Supabase] Custom fetch called: https://[project].supabase.co/auth/v1/token?grant_type=refresh_token`
- ❌ Custom fetch is NEVER called for queries (no log appears for `/rest/v1/events`)

**Conclusion:** The query builder fails during promise construction, BEFORE it attempts to call fetch.

### 1. Reinstalling Packages
```bash
npm uninstall @supabase/supabase-js
npm install @supabase/supabase-js@latest
# Deleted node_modules and package-lock.json
npm install
```
**Result:** No change

### 2. Minimal Client Config
```typescript
export const supabase = createClient(supabaseUrl, supabaseAnonKey);
```
**Result:** No change

### 3. Custom Fetch Implementation
```typescript
export const supabase = createClient(supabaseUrl, supabaseAnonKey, {
  global: {
    fetch: customFetch,
  }
});
```
**Result:** Fetch is called for auth but NOT for queries

### 4. Vite Configuration
```typescript
export default defineConfig({
  plugins: [react()],
  optimizeDeps: {
    include: ['@supabase/supabase-js'],
    esbuildOptions: {
      target: 'esnext'
    }
  },
  build: {
    target: 'esnext'
  }
})
```
**Result:** No change

### 5. Different Query Patterns
Tried:
- Simple select: `supabase.from('events').select('*')`
- With filters: `supabase.from('events').select('*').eq('id', '123')`
- With single: `supabase.from('events').select('*').single()`
- Different tables
**Result:** ALL queries hang

## Expected Behavior
The query promise should resolve with `{ data, error, count }` within a reasonable timeframe, just like the direct fetch does.

## Actual Behavior
The query promise never resolves. No HTTP request is sent. The application hangs waiting for the promise indefinitely.

## Impact
- **Severity**: Critical - All database queries are unusable
- **Workaround**: Using direct fetch with query.url (not ideal, requires accessing protected properties)
- **Scope**: Affects all SELECT queries on Windows 11 with Vite + React 19

### Browser Testing
Tested in:
- Chrome (latest)
- Edge (latest)
- Firefox (latest)

**Result:** Same issue in all browsers

### React Query Integration
Using @tanstack/react-query 5.90.2:
```typescript
const { data, isLoading, error } = useQuery({
  queryKey: ['events'],
  queryFn: () => eventsService.getEvents(),
});
```
Query remains in `isLoading: true` state forever.

### Network Inspection
Using Chrome DevTools Network tab:
- No request to `/rest/v1/events` appears
- Only auth-related requests appear (`/auth/v1/token`)

### RLS Policies
All RLS policies are correctly configured (verified by running direct SQL queries in Supabase dashboard, which return data successfully).

## Workaround Currently Using

```typescript
// Build query to get the URL and parameters
const query = supabase
  .from('events')
  .select('*', { count: 'exact' })
  .order('start_date', { ascending: true })
  .range(0, 19);

// Access protected URL property
// @ts-ignore
const directUrl = query.url?.toString();

// Execute with direct fetch
const directResponse = await fetch(directUrl, {
  headers: {
    'apikey': import.meta.env.VITE_SUPABASE_ANON_KEY,
    'Authorization': `Bearer ${import.meta.env.VITE_SUPABASE_ANON_KEY}`,
    'Prefer': 'count=exact'
  }
});

const data = await directResponse.json();
```

This works but:
1. Requires accessing protected `url` property
2. Bypasses Supabase's built-in transformations
3. Returns snake_case instead of camelCase
4. Not sustainable for production

## Questions

1. Why would the Headers object be empty when the query is built?
2. Why does custom fetch work for auth but not for queries?
3. Is there a specific Windows/Vite/React 19 compatibility issue?
4. Is there additional debugging we can enable to see why the promise never resolves?

## Repository
Full reproduction available at: https://github.com/[your-username]/[repo-name] (if you want to create one)

## Request
Please investigate why the query builder promise construction fails on Windows 11 with Vite + React 19. This appears to be an environment-specific issue that makes the Supabase JS client completely unusable without workarounds.

### Validations

### LINKED ISSUE #2013: Supabase Auth-JS Deadlock Issue

### LINKED ISSUE #2126: TOKEN_REFRESHED loop on init with ECC P-256 signing key - SIGNED_OUT after 30 rapid refreshes
### Describe the bug

On every page load, onAuthStateChange receives 30+ TOKEN_REFRESHED events in rapid succession (~100ms apart), followed by a SIGNED_OUT event that kills the session. This happens with zero Realtime channels active, with a clean localStorage, and regardless of supabase-js version (tested on 2.46.0 and 2.94.1).

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

1.Have a Supabase project with ECC (P-256) as the active JWT signing key (migrated automatically by Supabase)
2.Create a React app with a standard AuthProvider using onAuthStateChange
3.Load the app with a valid session in localStorage
4.Observe console logs

1 INITIAL_SESSION event, then normal session behavior with TOKEN_REFRESHED firing at most once per hour.
Actual behavior
INITIAL_SESSION  ← normal
TOKEN_REFRESHED  ← fires BEFORE INITIAL_SESSION on some reloads
TOKEN_REFRESHED
SIGNED_OUT       ← 5-6 seconds after INITIAL_SESSION
SIGNED_IN        ← session re-established
TOKEN_REFRESHED  ← rapid loop starts
TOKEN_REFRESHED  ← ~100ms between each
TOKEN_REFRESHED  ← 30+ times
...
SIGNED_OUT       ← session killed by "refresh token already used"
Key observations

The loop starts before INITIAL_SESSION is processed — it originates inside the Supabase client initialization, not in application code
The loop occurs with zero Realtime channels active
The loop occurs with clean localStorage (all sb-* keys manually deleted)
A minimal test component with only 2 Realtime channels and no AuthProvider logic works perfectly — no TOKEN_REFRESHED loop
The project recently had its JWT signing key automatically migrated from Legacy HS256 to ECC (P-256) by Supabase
Supabase dashboard settings: Refresh token reuse interval: 30s, Detect and revoke compromised tokens: enabled
The SIGNED_OUT is caused by "refresh token already used" — multiple parallel refresh attempts consuming the same token within the reuse interval

### System Info

```shell
@supabase/supabase-js: tested on 2.46.0 and 2.94.1 — same behavior on both
@supabase/auth-js: 2.94.1
@supabase/realtime-js: 2.94.1
Browser: Chrome (Windows)
JWT Signing Key: ECC (P-256) (migrated automatically, Legacy HS256 still present as "Previously used key")
React 18, Vite
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

### LINKED ISSUE #2145: Bug: _callRefreshToken permanently deletes session on non-retryable refresh failure, even when access token is still valid
# Bug: `_callRefreshToken` permanently deletes session on non-retryable refresh failure, even when access token is still valid

## Bug report

- [x] I confirm I have searched the Docs, Discussions, and Discord before opening this.

## Describe the bug

When `getSession()` proactively refreshes a token that's within the `EXPIRY_MARGIN_MS` (90s) window, and the refresh fails with a non-retryable error (e.g. 400 `invalid_grant`), `_callRefreshToken` calls `_removeSession()` which **permanently deletes the session from storage** — even though the access token is still valid for up to 90 more seconds.

After deletion, every subsequent `getSession()` call returns `{ session: null, error: null }` forever. The user is silently logged out with no error reported to the caller.

## The problematic code

[`GoTrueClient.ts` — `_callRefreshToken` method](https://github.com/supabase/supabase-js/blob/master/packages/core/auth-js/src/GoTrueClient.ts) (still present in v2.98.0, line ~1997 in compiled output):

```typescript
catch (error) {
    if (isAuthError(error)) {
        const result = { data: null, error };
        if (!isAuthRetryableFetchError(error)) {
            await this._removeSession();  // 💀 Permanently deletes session from storage
        }
        this.refreshingDeferred?.resolve(result);
        return result;
    }
}
```

The refresh is triggered proactively by `getSession()` → `__loadSession()`:

```typescript
const hasExpired = currentSession.expires_at
    ? currentSession.expires_at * 1000 - Date.now() < EXPIRY_MARGIN_MS  // 90,000ms
    : false;

if (hasExpired) {
    // Token still valid for up to 90s, but SDK considers it "expired"
    const { data: session, error } = await this._callRefreshToken(currentSession.refresh_token);
    // If this fails → session PERMANENTLY DELETED
}
```

## Why this is wrong

1. **The access token is still valid.** The `EXPIRY_MARGIN_MS` (90s) is a proactive optimization — the token hasn't actually expired. Destroying a valid token because an optimization failed is incorrect.

2. **The refresh failure is often transient.** Common causes include refresh token rotation races (two tabs refreshing simultaneously), brief Supabase service issues returning 400 instead of 5xx, and mobile browser tab lifecycle quirks.

3. **There is no recovery path.** Once `_removeSession()` fires, the session is gone from storage. `getSession()` returns `{ session: null, error: null }` — the `error: null` means callers have no way to distinguish "user is not logged in" from "SDK destroyed your valid session".

4. **`_removeSession()` fires `SIGNED_OUT` event.** This is misleading — the user never signed out. Listeners that clean up state on `SIGNED_OUT` will tear down the application unnecessarily.

### Automated test (runs against real GoTrueClient)

```typescript
import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
import { GoTrueClient } from "@supabase/auth-js";

describe("getSession permanently null after failed refresh", () => {
  let authClient: GoTrueClient;
  let memoryStore: Record<string, string>;

  const createExpiringSession = (secondsUntilExpiry: number) => ({
    access_token: "still-valid-access-token",
    refresh_token: "refresh-token-abc123",
    expires_in: secondsUntilExpiry,
    expires_at: Math.floor(Date.now() / 1000) + secondsUntilExpiry,
    token_type: "bearer",
    user: {
      id: "user-123",
      email: "test@example.com",
      aud: "authenticated",
      role: "authenticated",
      app_metadata: {},
      user_metadata: {},
      created_at: new Date().toISOString(),
    },
  });

  beforeEach(() => {
    vi.useFakeTimers({ shouldAdvanceTime: false });
    memoryStore = {};

    // Mock fetch: returns 400 for refresh (simulates token rotation race)
    const mockFetch = vi.fn().mockImplementation(async (url: string) => {
      if (url.includes("/token?grant_type=refresh_token")) {
        return {
          ok: false,
          status: 400,
          json: async () => ({
            error: "invalid_grant",
            error_description: "Invalid Refresh Token: Already Used",
          }),
          headers: new Headers({ "Content-Type": "application/json" }),
        };
      }
      return { ok: true, status: 200, json: async () => ({}), headers: new Headers() };
    });

    authClient = new GoTrueClient({
      url: "https://fake.supabase.co/auth/v1",
      headers: { apikey: "fake-anon-key" },
      storageKey: "sb-test-auth-token",
      autoRefreshToken: false,
      persistSession: true,
      detectSessionInUrl: false,
      storage: {
        getItem: (key: string) => memoryStore[key] ?? null,
        setItem: (key: string, value: string) => { memoryStore[key] = value; },
        removeItem: (key: string) => { delete memoryStore[key]; },
      },
      fetch: mockFetch,
    });
  });

  afterEach(() => vi.useRealTimers());

  it("BUG: session within 90s margin → refresh 400 → session PERMANENTLY deleted", async () => {
    // Token expires in 60s — within the 90s EXPIRY_MARGIN but STILL VALID
    const session = createExpiringSession(60);
    memoryStore["sb-test-auth-token"] = JSON.stringify(session);

    expect(memoryStore["sb-test-auth-token"]).toBeDefined(); // Session IS in storage

    const result1 = await authClient.getSession();

    // SDK tried to refresh, got 400, called _removeSession()
    expect(result1.data.session).toBeNull();
    expect(memoryStore["sb-test-auth-token"]).toBeUndefined(); // 💀 DELETED from storage

    // All subsequent calls return null forever — no recovery
    const result2 = await authClient.getSession();
    expect(result2.data.session).toBeNull();
    expect(result2.error).toBeNull(); // No error reported — silent failure
  });
});
```

### Browser console reproduction (against a real Supabase instance)

```javascript
// Run in browser console with any Supabase project
const { createClient } = await import("https://esm.sh/@supabase/supabase-js@2");

const SUPABASE_URL = "https://YOUR_PROJECT.supabase.co";
const ANON_KEY = "YOUR_ANON_KEY";
const REPRO_KEY = "sb-repro-bug-auth-token";

// 1. Inject a session within the 90s danger zone
const nowSec = Math.floor(Date.now() / 1000);
localStorage.setItem(REPRO_KEY, JSON.stringify({
  access_token: "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0.fake",
  refresh_token: "fake-refresh-token-will-get-400",
  expires_in: 60,
  expires_at: nowSec + 60,  // 60s remaining — within 90s EXPIRY_MARGIN
  token_type: "bearer",
  user: { id: "x", email: "x@x.com", aud: "authenticated", role: "authenticated",
          app_metadata: {}, user_metadata: {}, created_at: new Date().toISOString() }
}));

// 2. Wrap storage to trace operations
const trace = [];
const tracingStorage = {
  getItem: (key) => { const v = localStorage.getItem(key); trace.push({ op: "GET", key, found: v !== null }); return v; },
  setItem: (key, value) => { trace.push({ op: "SET", key }); localStorage.setItem(key, value); },
  removeItem: (key) => { trace.push({ op: "DELETE", key }); localStorage.removeItem(key); },
};

// 3. Create client — initialization triggers _recoverAndRefresh → _callRefreshToken → _removeSession
const supabase = createClient(SUPABASE_URL, ANON_KEY, {
  auth: { autoRefreshToken: false, persistSession: true, detectSessionInUrl: false,
          storageKey: REPRO_KEY, storage: tracingStorage }
});

await new Promise(r => setTimeout(r, 3000));

// 4. Observe the damage
console.table(trace);
// Output:
//   GET  sb-repro-bug-auth-token-code-verifier  → not found
//   GET  sb-repro-bug-auth-token                → found ✅
//   GET  sb-repro-bug-auth-token                → found ✅
//   DELETE sb-repro-bug-auth-token              → 💀 DELETED
//   GET  sb-repro-bug-auth-token                → not found
//   GET  sb-repro-bug-auth-token                → not found

const result = await supabase.auth.getSession();
console.log("session:", result.data.session);  // null
console.log("error:", result.error);            // null — SILENT FAILURE

localStorage.removeItem(REPRO_KEY);
```

### Verified live

We ran this reproduction against a real Supabase instance (`sfkvqohjfyohvhgsgybk.supabase.co`) with auth-js v2.88.0 and confirmed the storage trace shows the exact sequence: `GET (found) → GET (found) → DELETE → GET (not found)`.

We also confirmed the bug is **still present in v2.98.0** — the `_removeSession()` call inside `_callRefreshToken`'s catch block for non-retryable errors remains at line 1997 of the compiled output.

## Real-world impact

We're seeing this in production affecting real users:
- **19 occurrences, 5 users** reported via Sentry
- Primarily affects **Mobile Safari (iOS 18.7)** and **Chrome on Windows/iOS**
- Manifests as HTTP 401 errors during long-running operations (SSE streams, long http calls) that start during the ~59 minute mark of a session
- Users have no way to recover without a full page reload and re-login

## Suggested fix

`_callRefreshToken` should **not** call `_removeSession()` when the access token is still within its valid expiry window. A proactive refresh failure should degrade gracefully — return the existing (still-valid) session, not destroy it.

```typescript
// Instead of unconditionally deleting:
if (!isAuthRetryableFetchError(error)) {
    await this._removeSession();
}

// Consider: only remove if the token has ACTUALLY expired
if (!isAuthRetryableFetchError(error)) {
    const currentSession = await this._getSessionFromStorage();
    const actuallyExpired = currentSession?.expires_at
        ? currentSession.expires_at * 1000 < Date.now()
        : true;
    if (actuallyExpired) {
        await this._removeSession();
    } else {
        // Token still valid — keep it, log the refresh failure, try again later
        this._debug('_callRefreshToken', 'refresh failed but token still valid, keeping session');
    }
}
```

Alternatively, `getSession()` could simply return the existing session when the proactive refresh fails, since the token hasn't actually expired yet.

## Related issues

**Note:** `auth-js` has moved into this monorepo at `packages/core/auth-js/`. The old issues are archived at `supabase/auth-js`:

- supabase/auth-js#904 — "Current session lost when auth function call fails" (closed by supabase/auth-js#915, but that PR only fixed sign-in/sign-up flows, NOT `_callRefreshToken`)
- supabase/auth-js#141 — "Network error removes session data" (closed, partially fixed — retryable errors now preserved, but non-retryable auth errors still nuke session)
- #1560 — "iOS Safari/Chrome: getSession() intermittently returns null" (still open — likely same root cause)

## Environment

- auth-js version: 2.88.0 (verified still present in 2.98.0)
- Browser: Mobile Safari 26.2 (iOS 18.7), Chrome 144 (Windows), Chrome Mobile iOS 145
- Framework: React (Vite)

### LINKED ISSUE #2344: auth-js: _initialize deadlocks when onAuthStateChange registers during init while session is within 90s of expiry
## Bug report

> Filing here since `supabase/auth-js` is archived. Affects the auth-js package in `packages/core/auth-js/`.

### Summary

When an app registers `onAuthStateChange` synchronously during `_initialize`'s lock callback **and** the persisted session is within the 90s expiry threshold, the auth client deadlocks. `_initialize` never resolves, the navigator Web Lock named `lock:sb-<ref>-auth-token` is held forever, and every subsequent `getSession()` / `getUser()` / `from()` call queues behind it.

### Affected version

Reproduced on `@supabase/supabase-js@2.105.4` (which bundles `@supabase/auth-js`) and `@supabase/ssr@0.10.3`.

### Reproduction

```ts
useEffect(() => {
  const supabase = createBrowserClient(URL, ANON);
  // Both calls fire synchronously in the same tick:
  supabase.auth.getUser().then(...);
  supabase.auth.onAuthStateChange(cb);   // ← registers DURING init
}, []);
```

Conditions for the hang:
1. A persisted session in storage with `expires_at` within 90 seconds of now (e.g. happens on reload near token-refresh boundary).
2. App registers `onAuthStateChange` in the same useEffect tick as init kicks off.

Symptoms (introspected via the singleton client exposed for diagnostics):

| Property | Value |
|---|---|
| `auth.lockAcquired` | `true` (held forever) |
| `auth.pendingInLock.length` | `1+` (never drains) |
| `auth._initialized` | `false` |
| `auth.currentSession` | `undefined` |
| `await navigator.locks.query()` | one held lock named `lock:sb-<ref>-auth-token` |
| `auth._refreshAccessToken(refreshToken)` called directly | resolves in <1s with a fresh session |

### Mechanism

1. `_initialize()` acquires the Web Lock and runs `_recoverAndRefresh()`.
2. Session is < 90s from expiry, so `_recoverAndRefresh()` calls `_callRefreshToken()`. This sets `refreshingDeferred` and starts the refresh.
3. Concurrently, the app's `useEffect` calls `supabase.auth.onAuthStateChange(cb)`. The listener registration calls `_emitInitialSession(id)` for the new listener.
4. `_emitInitialSession` → `_useSession` → `__loadSession`. **`__loadSession` itself contains a refresh path**:
   ```js
   let s = !!t.expires_at && t.expires_at*1000 - Date.now() < 9e4;
   if (s) await this._callRefreshToken(t.refresh_token);
   ```
   The SDK source already warns about this at the top of `__loadSession`:
   ```js
   this.lockAcquired || this._debug('#__loadSession()', 'used outside of an acquired lock!', Error().stack);
   ```
5. `_callRefreshToken` is idempotent via `refreshingDeferred`, so the second call awaits the same in-flight promise. The refresh itself completes — the persisted cookie ends up with fresh tokens.
6. However, somewhere in the resulting Promise composition (the `_callRefreshToken` `.then` chain inside `_emitInitialSession`'s `_useSession`), an entry is left in `pendingInLock` that never resolves.
7. `_acquireLock`'s drain loop blocks forever:
   ```js
   while (this.pendingInLock.length) {
     let arr = [...this.pendingInLock];
     await Promise.all(arr);
     this.pendingInLock.splice(0, arr.length);
   }
   ```
   Drain never completes → Web Lock never released → `_initialized` never set → all later operations stuck.

### What works vs what hangs

- `auth.storage.getItem(storageKey)` directly: returns in 1ms with the parsed session string.
- `auth._refreshAccessToken(refreshToken)` directly: resolves in ~850ms with a fresh session.
- `auth.getSession()` / `auth.getUser()` / any `from(...).select(...)`: hang indefinitely.
- Direct `fetch()` to `/rest/v1/<table>` with the cookie's bearer token: returns in <1s. Only the SDK wrapper is stuck.

### Why first-load post-deploy "works" but reload hangs

- First load after fresh sign-in: session is far from the 90s threshold; `_recoverAndRefresh` doesn't call `_callRefreshToken`; no race.
- Reload near expiry: refresh path fires; race triggers.

This pattern likely explains other intermittent / "sometimes hangs on reload" reports.

### App-side workaround

In all hooks that call `onAuthStateChange`, defer the registration until `getUser()` (or `getSession()`) resolves:

```ts
useEffect(() => {
  const supabase = createBrowserClient(URL, ANON);
  let mounted = true;
  let unsubscribe: (() => void) | undefined;

  const setup = async () => {
    try {
      const { data: { user } } = await supabase.auth.getUser();
      if (!mounted) return;
      // ... use user
    } catch { /* ignore */ }

    if (!mounted) return;
    const { data: { subscription } } = supabase.auth.onAuthStateChange(cb);
    unsubscribe = () => subscription.unsubscribe();
  };

  setup();
  return () => { mounted = false; unsubscribe?.(); };
}, []);
```

`getUser()` awaits `initializePromise` cleanly via `_acquireLock`'s blessed path. By the time the listener subscribes, `_initialize` has fully resolved and the lock has been released, so `_emitInitialSession` runs in the post-init state where it can't race.

This fully eliminated the hang for us in production. Verified by manually setting cookie `expires_at` to `now+85s` and reloading: pre-fix → deadlock every time; post-fix → refresh fires inside the lock, completes, drains, page renders normally.

### Suggested upstream fix (any of)

1. **Make `__loadSession` skip the in-line refresh when called outside the lock.** When `lockAcquired` is false, return the (possibly-soon-to-be-stale) session from storage and let the lock holder drive the refresh. The pre-existing `outside of an acquired lock!` warning hints this path was already known to be dangerous.
2. **Defer `_emitInitialSession` until after `initializePromise` resolves.** A natural `await this.initializePromise` at the top of `_emitInitialSession` would prevent it from running concurrently with the lock callback.
3. **Add an upper time bound to `_acquireLock`'s drain loop** so a buggy queue entry doesn't permanently brick the client (defense in depth).

### Environment

- `@supabase/supabase-js@2.105.4`, `@supabase/ssr@0.10.3`
- Browser: Chromium (verified in Playwright Chromium and real Chrome)
- App stack: Next.js 16 App Router, React 19
- Reproduces against a vanilla `createBrowserClient` setup with a session that happens to fall within 90s of expiry on reload

### Diagnostic harness used

A `?<flag>=1`-gated probe in our `lib/supabase/client.ts` exposed `window.__supabase`, logged `getSession()` resolution timing, and ticked `_initialized` / `currentSession` state every second. Combined with `navigator.locks.query()` and `Function.prototype.toString` reading of the deminified bundled source, this localized the bug to the exact `_acquireLock` drain step. Happy to share the harness if useful.

### LINKED ISSUE #2376: Intermittent RPC hang with lock: No-Op config (not covered by PR #2106)
**Bug Summary**

`supabase.rpc()` calls hang indefinitely (~5–10% of calls) when `createClient` is configured with `lock: async (_n, _t, fn) => fn()` (No-Op lock for multi-tab scenarios). The Promise neither resolves nor rejects. PR #2106's `steal: true` retry fix doesn't apply because No-Op lock bypasses the `navigator.locks` code path entirely.

**Environment**

- `@supabase/supabase-js` `2.105.4` (latest stable as of 2026-05-08)
- `@supabase/auth-js` `2.105.4` (bundled, contains PR #2106 fix already)
- React `19.2.6`
- Vite `8.0.12` (dev server, React Strict Mode **off**)
- Chrome `130+` on macOS `14`
- Single tab, single authenticated user session (no multi-tab scenario at repro time)

**Setup (relevant `createClient` config)**

```typescript

const FETCH_TIMEOUT_MS = 15000

const fetchWithTimeout: typeof fetch = (input, init) => {
  const timeoutId = setTimeout(
    () => controller.abort(new DOMException(`Timeout ${FETCH_TIMEOUT_MS}ms`, 'TimeoutError')),
    FETCH_TIMEOUT_MS
  )
  return fetch(input, { ...init, signal: controller.signal })
    .finally(() => clearTimeout(timeoutId))
}

export const supabase = createClient(SUPABASE_URL, SUPABASE_ANON_KEY, {
  auth: {
    // No-Op lock as a workaround for an earlier multi-tab navigator.locks hang
    // (orphaned-lock cluster). Replaces the default navigator.locks usage.
    lock: async (_name, _acquireTimeout, fn) => fn(),
  },
  global: {
    fetch: fetchWithTimeout,
  },
})
```

**Reproduction steps**

1. Use the config above.
2. Log in via `supabase.auth.signInWithPassword(...)`.
3. Call any RPC repeatedly:

```typescript
for (let i = 0; i < 20; i++) {
  const start = performance.now()
  const { data, error } = await supabase.rpc('your_rpc', { /* args */ })
  console.log(`call ${i}: ${(performance.now() - start).toFixed(0)}ms`, { data, error })
}
```

4. Observe: 1–2 of the 20 calls hang. They eventually error with our `fetchWithTimeout`'s `TimeoutError` after 15s. Without the timeout wrapper, the Promise stays pending forever.

**Expected behavior**

All 20 RPC calls resolve with HTTP 200 within ~100ms each.

**Actual behavior**

5–10% of calls hang. They are not aborted by the client, they don't time out at the PostgREST layer (the request never reaches the server — see evidence below). Without an external timeout wrapper, they remain pending indefinitely.

**Evidence — server is fine, client is the culprit**

Equivalent direct `fetch()` (bypassing the SDK entirely, same auth token from `localStorage['sb-<project-ref>-auth-token']`, same endpoint, same headers):

```typescript
const SUPABASE_URL = import.meta.env.VITE_SUPABASE_URL
const SUPABASE_KEY = import.meta.env.VITE_SUPABASE_ANON_KEY

function extractAccessToken(): string | null {
  for (const key of Object.keys(localStorage)) {
    if (key.startsWith('sb-') && key.endsWith('-auth-token')) {
      const parsed = JSON.parse(localStorage.getItem(key)!)
      return parsed?.access_token ?? null
    }
  }
  return null
}

async function directRpc<T>(rpcName: string, args: Record<string, unknown>): Promise<T> {
  const token = extractAccessToken()!
  const r = await fetch(`${SUPABASE_URL}/rest/v1/rpc/${rpcName}`, {
    method: 'POST',
    headers: {
      apikey: SUPABASE_KEY,
      Authorization: `Bearer ${token}`,
    },
    body: JSON.stringify(args),
  })
  if (!r.ok) throw new Error(`HTTP ${r.status}: ${await r.text()}`)
  return await r.json() as T
}

// Stress test:
for (let i = 0; i < 10; i++) {
  const start = performance.now()
  await directRpc<unknown>('your_rpc', { /* same args */ })
  console.log(`direct ${i}: ${(performance.now() - start).toFixed(0)}ms`)
}
```

Result on our environment: **10/10 HTTP 200, avg 93ms, max 322ms.** Same RPC, same user, same session, same browser — no hangs.

**Why this is NOT issue #2111 / fixed by PR #2106**

[#2111](https://github.com/supabase/supabase-js/issues/2111) describes orphaned `navigator.locks` leading to indefinite auth hangs. [PR #2106](https://github.com/supabase/supabase-js/pull/2106) fixes that by retrying lock acquisition with `{ steal: true }` after a 5s timeout.

Our config explicitly **bypasses** `navigator.locks` via `lock: async (_n, _t, fn) => fn()`. The `steal: true` retry path is never taken because the lock callback resolves immediately. We're hitting a different race.

**Hypothesis**

Without lock-based serialization, the auth-refresh background cycle (or some other internal queue) can interleave with concurrent RPC requests in a way that leaves a fetch in a non-resolvable state. PostgREST client internals (auth-token refresh handler? request queue?) seem to assume the lock semantics are present.

This is consistent with:

- Bug only manifests post-login (anonymous-only calls don't reproduce reliably for us).
- Bug rate is ~5–10%, suggesting timing-dependent race.
- Direct fetch (which bypasses the entire SDK's request pipeline) is 100% reliable.

**Asks**

1. Is `lock: async (_n, _t, fn) => fn()` (No-Op lock) a **supported** configuration? The TypeScript signature allows it, but if the SDK assumes lock semantics elsewhere, the No-Op breaks invariants.
2. If supported: is there a known race condition between the No-Op lock path and auth-refresh / request queueing that could leave fetches in a pending-forever state?
3. If not supported: what's the recommended pattern for **opting out of multi-tab lock serialization** without losing the SDK's request reliability? (Background: many internal-tool SPAs don't need cross-tab serialization but were hit by the original navigator.locks orphan cluster.)

**Workaround currently in production**

We bypass the SDK for the affected RPCs entirely (direct `fetch()` with manual token extraction). Plus a global `fetchWithTimeout` wrapper as a safety net for any other SDK call that might hang. Functional but introduces technical debt (duplicate request path, manual auth-storage parsing).

Happy to provide a minimal runnable StackBlitz repro if useful — let me know.

---

*Filed from an internal SaaS team running supabase-js in a small React admin app — happy to dig deeper if useful.*

Bewinxed:
### What changed?

`GoTrueClient` now defaults to lockless coordination. The `navigator.locks`-based mutex no longer runs by default — most users get the lockless path. Callers who explicitly pass a custom `lock` (typically React Native `processLock` or Node multi-process setups) keep the old behavior on an opt-in legacy path so the change is backwards-compatible.

The lockless default uses:

- `refreshingDeferred` (already existed) to single-flight in-instance concurrent refreshes.
- A two-leg commit guard inside `_callRefreshToken`: **(1)** snapshots storage before the rotated-token fetch and re-reads after, discarding rotated tokens if a non-null snapshot was cleared between the two reads (typical case: a concurrent `signOut` ran `_removeSession`); **(2)** captures a session-removal epoch counter before `_saveSession` and re-checks after, so a `signOut` that interleaves inside `_saveSession`'s storage-write awaits is also caught. Either leg returns `{ data: null, error: new AuthRefreshDiscardedError() }`. `_recoverAndRefresh` recognises this error and skips its own `_removeSession()` call so no duplicate `SIGNED_OUT` event fires.
- A new exported error class, `AuthRefreshDiscardedError` (with an `isAuthRefreshDiscardedError` type guard), is what the commit guard returns when it throws away a successfully-rotated token. Distinct from `AuthRetryableFetchError` (transient network) and `AuthApiError` (server rejection). Surfaces through `refreshSession()` and `getSession()` results.
- A new `client.auth.dispose()` tears down the auto-refresh interval, the `visibilitychange` listener, the `BroadcastChannel`, and registered `onAuthStateChange` subscribers. Idempotent. Designed for React Strict Mode and HMR cleanup hooks. In-flight `fetch` calls are not aborted — they run to completion.
- GoTrue handles cross-tab refresh races server-side via the v1 parent-of-active path at `internal/tokens/service.go:376-385`, so the client doesn't need to coordinate. The current default is v1 (`RefreshTokenAlgorithmVersion = 0`); v2 (counter-based) is opt-in via `RefreshTokenUpgradePercentage` and gated rollout. This PR's reasoning relies on v1 behaviour, which is what customers actually run today.

The legacy opt-in path (`settings.lock != null`):

- `_acquireLock`, `pendingInLock`, `lockAcquired`, `lockAcquireTimeout`, and all 14 call-site wrappers are preserved verbatim from `master` and gated on `this.lock != null`. The default path is untouched by them.
- Every gated site is marked `// TODO(v3): remove legacy lock path` so the eventual v3 cleanup is a mechanical search-and-delete. A separate v3 follow-up tracks that work.
- The `lock` and `lockAcquireTimeout` constructor options are accepted and honored when supplied; both are `@deprecated` in JSDoc with "Custom locks still work in v2.x for backwards compatibility. Will be removed in v3."
- `navigatorLock`, `processLock`, `LockAcquireTimeoutError`, `NavigatorLockAcquireTimeoutError`, `ProcessLockAcquireTimeoutError`, and `internals` in `lib/locks.ts` remain exported. Still `@deprecated`.

Stale JSDoc referencing the lock on `getSession`, `onAuthStateChange`, `_useSession`, `_challengeAndVerify`, and `_listFactors` now matches the new default behaviour. The `@deprecated` tag on the async `onAuthStateChange` overload is **kept**, with its reason updated to point at the one residual reentry hazard (`refreshSession` called from inside a `TOKEN_REFRESHED` handler — `refreshingDeferred.resolve` happens after `_notifyAllSubscribers` returns).

The shared mutex has caused seven failure classes on the issue tracker for 18+ months. Each earlier patch fixed one symptom and created another:

- [#1962](https://github.com/supabase/supabase-js/pull/1962) configurable timeout
- [#2106](https://github.com/supabase/supabase-js/pull/2106) steal fallback
- [#2178](https://github.com/supabase/supabase-js/pull/2178) steal-cascade guard

[#2235](https://github.com/supabase/supabase-js/pull/2235) (primitive swap to `processLock`) was an earlier attempt at moving off `navigator.locks` and is no longer being actively pursued. Community PRs [#2016](https://github.com/supabase/supabase-js/pull/2016) and [#2019](https://github.com/supabase/supabase-js/pull/2019) fixed individual symptoms and were closed waiting on a structural fix.

Failure classes resolved by this PR's default (lockless) path:

1. Strict Mode / HMR orphan deadlocks. Nothing to orphan on the default path.
2. iOS Safari `INITIAL_SESSION` / `getSession()` hangs with a persisted session ([#936 @bugprone](https://github.com/supabase/supabase-js/issues/936#issuecomment-3691211975)). The default path no longer touches `navigator.locks`.
3. Sentry flooding with `AbortError: Lock broken by another request with the 'steal' option` ([#2013 @cpannwitz](https://github.com/supabase/supabase-js/issues/2013#issuecomment-4024042017)). The `{ steal: true }` fallback that produced these errors is bypassed on the default path.
4. Subscriber re-entry deadlock (the one at the old `GoTrueClient.ts:3824`). The cycle needed the lock; a callback that awaits `getUser()` now just runs on the default path.
5. Visibility/tick race ([#936 @JTBrinkmann](https://github.com/supabase/supabase-js/issues/936#issuecomment-3728405087)). The visibility handler and the auto-refresh tick no longer share a lock to fight over on the default path.
6. Stale tickers/listeners on un-disposed clients. `dispose()` cleans them up.
7. `signOut` blocked behind in-flight refresh. `signOut` now runs concurrently with the refresh on the default path, and the two-leg commit guard prevents the refresh from writing rotated tokens after `_removeSession()` cleared storage — whether the clear happened before the post-fetch snapshot read or inside `_saveSession`'s storage writes.

Callers on the legacy opt-in path (explicit `settings.lock`) keep the old serialization semantics and the failure modes that come with them. They accepted those when they opted in; they can drop the option to migrate to the lockless default at any time.

Why default lockless: cross-tab races are handled on the server (GoTrue's parent-of-active), in-instance refresh dedup is already handled by `refreshingDeferred`, and the only job the lock did beyond that was serializing subscriber callbacks — which is the deadlock we're fixing.

Closes (test target): #2013, #936, #2111

### `dispose()` in a React app

```ts
import { useEffect } from 'react'

export function useSupabase() {
  useEffect(() => {
    const supabase = createClient(URL, KEY)
    return () => {
      supabase.auth.dispose()
    }
  }, [])
}
```

### Subscriber callbacks can call other auth methods now (default path)

```ts
// Previously: deadlocked. Callback held the lock; getUser tried to acquire it.
// Now (default path): works.
supabase.auth.onAuthStateChange(async (event, session) => {
  if (event === 'SIGNED_IN') {
    const { data } = await supabase.auth.getUser()
    // ...
  }
})
```

One residual hazard remains: calling `refreshSession` from inside a `TOKEN_REFRESHED` handler still deadlocks via `refreshingDeferred`. The async overload of `onAuthStateChange` keeps its `@deprecated` marker for that reason.

### `AuthRefreshDiscardedError` for the signOut-during-refresh race

```ts
import { isAuthRefreshDiscardedError } from '@supabase/auth-js'

const { data, error } = await supabase.auth.refreshSession()
if (isAuthRefreshDiscardedError(error)) {
  // A concurrent signOut cleared storage between when this refresh
  // started and when it came back. The rotated tokens were discarded.
  // The SIGNED_OUT event already fired from the concurrent signOut.
}
```

### Legacy lock opt-in (existing custom-lock users)

```ts
// React Native processLock setup or Node multi-process setup
import { processLock } from '@supabase/auth-js'

const supabase = createClient(URL, KEY, {
  auth: { lock: processLock }, // legacy path, still works, deprecated for v3
})
```

## Breaking changes

The earlier draft of this PR silently dropped custom `lock` functions, which was behaviorally breaking for opt-in users. The current design preserves that behavior on a gated legacy path so the change ships as a v2 minor. Custom-lock users have time to migrate before v3 removes the option entirely.

### Behaviour changes worth flagging for review

- **Default coordination is lockless.** A client constructed without a `lock` option no longer acquires `navigator.locks` or any in-process lock. For most users this is invisible (the lock was internal coordination). For users who depended on observing the lock indirectly (rare), the change is observable.
- **`_autoRefreshTokenTick` may now run concurrently with `signOut` / `setSession` / `getUser` on the default path.** Previously `_acquireLock(0, ...)` made the tick skip whenever **any** auth op held the lock. The lockless equivalent only skips when `refreshingDeferred` is set. The commit guard keeps storage consistent under the new concurrency. The legacy lock opt-in path retains the old skip-on-any-lock behavior.
- **Subscriber timing on the default path:** subscribers stay awaited; same as before. What changes is that `signOut` no longer waits for an in-flight refresh's HTTP and continuation to finish before its own fetch goes out. Both fetches now run concurrently, and the commit guard keeps storage consistent.
- **New `@deprecated` warnings on `lock` and `lockAcquireTimeout`.** Both options now carry `@deprecated` JSDoc on `@supabase/auth-js` and `@supabase/supabase-js` types. Consumers with strict deprecation lint rules (e.g. `deprecation/deprecation` as error) will see new warnings after upgrading. The options continue to work in v2.x; removal is slated for v3.

## Checklist

- [x] I have added tests for new functionality (commit guard, `dispose()`, subscriber re-entry no-deadlock, default vs. legacy path)
- [x] I have updated documentation (JSDoc on affected methods, type-level deprecation notices)

### Alternatives considered

**Continuing to patch `navigator.locks`** (better steal recovery, lower timeouts, smarter error filtering). Each prior patch in this direction (#1962, #2106, #2178) fixed one symptom and produced another. The Web Locks API has no recovery for orphaned holders other than `{ steal: true }`, which leaves the previous `fn()` running concurrently with the new holder and creates the steal-cascade error storm. Each fix here swaps one failure for another instead of removing the contention that causes them.

**Swapping `navigatorLock` for `processLock` as the default browser lock** (the direction #2235 explored). Removes the cross-process orphan failure but keeps the same shape of the bug: one primitive serializing operations that don't all need the same synchronization.

**Non-blocking subscriber notifications alone** (#2016). Fixes subscriber re-entry. The other six failure classes still bite.

**Removing the lock entirely (no legacy opt-in).** The earlier draft of this PR did this. It silently dropped custom locks supplied via `settings.lock`, which broke React Native `processLock` and Node multi-process setups. The current design preserves those callers on a gated legacy path so the change is non-breaking; the legacy path is slated for removal in v3.

**Adding an `AbortController` layer for in-flight operation cancellation.** Deferred to a follow-up, not included here. The commit guard catches the races that affect correctness (signOut overwriting cleared storage with rotated tokens, in both the fetch and storage-write phases). AbortController would be a UX improvement on top: cancel the in-flight refresh as soon as signOut runs, instead of letting it finish and discarding the result. Out of scope for this PR.

**Deferring the commit guard and accepting eventual consistency.** Considered and rejected. Without a guard, a refresh that completes after `_removeSession()` cleared storage will write the rotated tokens back. Subscribers then see `SIGNED_OUT` followed by `TOKEN_REFRESHED`, with stale tokens in storage until the next refresh tick (~30s) fails against the server and clears them. The guard is about 30 lines (both legs combined). Better to land it with this PR than to ship the bug and patch it later.

### Server-side context

Why cross-tab is safe on the default path: GoTrue's parent-of-active path at `internal/tokens/service.go:376-385` (the v1 branch, `*models.RefreshToken`). When a request arrives with a revoked refresh token whose child is the currently-active token, the server returns the active token instead of rejecting. Two tabs that POST the same refresh token concurrently both receive the same rotated token under DB row locking. This is the production-default behaviour (`RefreshTokenAlgorithmVersion = 0`, v1). v2 (counter-based, gated on `RefreshTokenUpgradePercentage`) is safe under the same N-tab concurrency, covered by `TestConcurrentReuse`.

### Test coverage

Updated tests in `packages/core/auth-js/test/GoTrueClient.test.ts`:

- `'Lockless coordination (default) and legacy lock opt-in'` describe:
  - default path: client with no `lock` option, assert `_acquireLock` is never invoked
  - legacy path: client with custom `lock`, assert `lock` IS invoked (`.toHaveBeenCalled()`)
  - `lockAcquireTimeout` accepted on both paths (lockless ignores it at runtime; legacy uses it)
  - subscriber callback may call other auth methods from inside `onAuthStateChange` without deadlock
- `'dispose() lifecycle'` describe: idempotency, subscriber clearing, ticker stopping.
- `'Refresh commit guard (signOut-during-refresh race)'` describe (4 tests): cleared-mid-flight discard, empty-storage acceptance, different-session acceptance, different-session-written-mid-flight discard.

Updated tests in `packages/core/auth-js/test/GoTrueClient.browser.test.ts`:

- `'Lockless coordination: navigator.locks should NOT be invoked'`: default path no longer touches the browser API.
- `'Legacy lock opt-in: custom `lock` function is invoked when supplied'`: custom lock IS called on the legacy path.

Updated tests in `packages/core/supabase-js/test/unit/SupabaseAuthClient.test.ts`:

- `'should pass through lockAcquireTimeout option'` and `'should accept auth.lockAcquireTimeout and wire it to auth client'`: assert the option flows from `createClient` through to the auth client instance (via a targeted `as unknown as { lockAcquireTimeout: number }` cast). Comment in-source explains the cast is greppable for the v3 cleanup.

### AI review responses

Addresses inline AI review feedback on this PR:

- **TOCTOU window inside `_saveSession`** (MEDIUM): closed via a new `_sessionRemovalEpoch` counter, incremented synchronously at the top of `_removeSession` before any `await`, captured in `_callRefreshToken` immediately before `_saveSession`, and re-checked after. If the epoch advances during the save, the rotated tokens are undone directly with `removeItemAsync` (not via `_removeSession`, which would emit a duplicate `SIGNED_OUT`).
- **Refresh-token fragments in debug logs** (LOW): removed `substring(0, 5)` of `refresh_token` from all four `_debug()` paths (`_callRefreshToken` / `_refreshAccessToken` `debugName` construction, plus the two fields in the commit-guard discard log). Discard logs now use presence indicators (`'present'` / `'replaced'` / `'cleared'`) instead of credential fragments.

### Resolved during review

1. **`AbortSignal.any` plumbing in `lib/fetch.ts`** — not done here and not blocking. Deferred to the AbortController follow-up, which gates on `AbortSignal.any` (Node 20.3+ floor or polyfill). Worth noting the follow-up has rate-limit-posture implications, not just UX: cancelling refreshes that the commit guard is about to discard reduces project-level `/token` volume.
2. **`@supabase/supabase-js` types deprecation** — confirmed. `lock` and `lockAcquireTimeout` are `@deprecated` in both `@supabase/auth-js` and `@supabase/supabase-js` types. The supabase-js surface is the one most users see; deprecating only in auth-js would miss the JSDoc tooltip for the passthrough. Flagged separately under "Behaviour changes worth flagging" so the new lint warnings don't surprise downstream consumers.

### v3 cleanup

A separate v3 follow-up tracks the mechanical removal of the legacy lock path. Every gated site in `GoTrueClient.ts` has a `// TODO(v3): remove legacy lock path` marker; the follow-up lists the full file-by-file scope (covering both `@supabase/auth-js` and `@supabase/supabase-js`).

Replaces #2387 (which was opened against `develop` before the v3 branching update).

### DIFF
--- a/packages/core/auth-js/migrations/lockless-coordination.md
+++ b/packages/core/auth-js/migrations/lockless-coordination.md
@@ -1,0 +1,89 @@
+# Lockless auth coordination
+
+**Since:** v2.X.Y (update at release time)
+**Action required by:** v3.0.0
+
+`@supabase/auth-js` now coordinates session refreshes without a shared mutex by default. The legacy `lock` option is still honored when supplied, but it is deprecated and will be removed in v3.
+
+## What changed
+
+- **Default coordination is lockless.** A client constructed without a `lock` option no longer acquires `navigator.locks` or any in-process lock. In-tab concurrent refreshes are deduplicated via the pre-existing single-flight (`refreshingDeferred`); cross-tab refresh races are resolved by GoTrue's server-side parent-of-active mechanism on the v1 refresh-token path.
+- **Commit guard inside `_callRefreshToken`.** The client snapshots storage before the rotated-token fetch and re-reads after. If a non-null pre-fetch snapshot was cleared between the two reads (typical case: a concurrent `signOut` ran `_removeSession`), the rotated tokens are discarded instead of being written back over the cleared storage. The discarded result resolves with `{ data: null, error: new AuthRefreshDiscardedError() }`.
+- **New `AuthRefreshDiscardedError`** (with `isAuthRefreshDiscardedError` type guard). Surfaces through `refreshSession()` and `getSession()` results when the commit guard fires. Distinct from `AuthRetryableFetchError` (transient network) and `AuthApiError` (server rejection).
+- **New `client.auth.dispose()`.** Tears down the auto-refresh interval, the `visibilitychange` listener, the `BroadcastChannel`, and registered `onAuthStateChange` subscribers. Idempotent. Designed for React Strict Mode and HMR cleanup hooks. In-flight `fetch` calls are not aborted — they run to completion.
+- **`lock` and `lockAcquireTimeout` options.** Accepted and honored when supplied (legacy opt-in path); both are `@deprecated` and will be removed in v3.
+
+## Who is affected
+
+**Most callers: nothing to do.** If you do not pass a `lock` option to `createClient` / `GoTrueClient`, you are already on the lockless default path and will get the bug fixes and new APIs without any code change.
+
+**Callers passing a custom `lock`** (typically React Native `processLock`, Node multi-process setups with shared AsyncStorage, or a custom lock implementation):
+
+- v2.x (this release): your custom `lock` is still invoked exactly as before. The legacy `_acquireLock` machinery is preserved on an opt-in path gated by `settings.lock != null`. No code change required.
+- v3.0.0 (planned): the `lock` and `lockAcquireTimeout` options will be removed entirely. Drop them from your client options before upgrading to v3.
+
+## New APIs worth knowing
+
+### `client.auth.dispose()`
+
+Tears down the client's background work in one call. Safe to call repeatedly.
+
+```ts
+useEffect(() => {
+  const supabase = createClient(URL, KEY)
+  return () => {
+    supabase.auth.dispose()
+  }
+}, [])
+```
+
+### `AuthRefreshDiscardedError`
+
+Returned from `refreshSession()` / `getSession()` when the commit guard discards a successfully-rotated session.
+
+```ts
+import { isAuthRefreshDiscardedError } from '@supabase/auth-js'
+
+const { data, error } = await supabase.auth.refreshSession()
+if (isAuthRefreshDiscardedError(error)) {
+  // A concurrent signOut cleared storage between fetch start and now.
+  // The rotated tokens were discarded; the SIGNED_OUT event already fired.
+  // Treat as a successful no-op — no need to retry.
+}
+```
+
+## Behavior changes worth flagging
+
+- **`_autoRefreshTokenTick` may run concurrently with `signOut` / `setSession` / `getUser`** on the lockless default path. Previously the tick used `_acquireLock(0, ...)` which skipped whenever any auth op held the lock. The lockless equivalent only skips when `refreshingDeferred` is set. The commit guard keeps storage consistent under the new concurrency. The legacy lock opt-in path retains the old skip-on-any-lock behavior.
+- **`onAuthStateChange` async callbacks** that call `getUser`, `setSession`, or read the session from inside the callback are now safe on the default path (previously deadlocked through the lock). One residual hazard remains: calling `refreshSession` (or anything routing through `_callRefreshToken`) from inside a `TOKEN_REFRESHED` handler still deadlocks via `refreshingDeferred`. The `@deprecated` marker on the async overload is kept with its reason updated to point at this specific case.
+- **Subscriber timing on the default path:** subscribers stay awaited; same as before. What changes is that `signOut` no longer waits for an in-flight refresh's HTTP and continuation to finish before its own fetch goes out. Both fetches now run concurrently, and the commit guard keeps storage consistent.
+
+## Migration steps
+
+### If you do not pass a custom `lock` (most users)
+
+No action required for v2. No action required for v3.
+
+### If you pass a custom `lock` (e.g., React Native `processLock`)
+
+No action required for v2 — your lock continues to work.
+
+For v3 readiness:
+
+1. Remove `lock` and `lockAcquireTimeout` from your `createClient` / `GoTrueClient` options before upgrading to v3.
+2. If you depended on cross-process serialization (e.g., Node multi-process with shared AsyncStorage), validate that the lockless coordination (in-tab single-flight + server parent-of-active) is sufficient for your runtime. The default is safe for the cases the lock was originally added to handle (cross-tab refresh races), since the server resolves them.
+
+```ts
+// Before (v2.x, still works):
+const supabase = createClient(URL, KEY, {
+  auth: { lock: processLock, lockAcquireTimeout: 5000 },
+})
+
+// After (v3-ready):
+const supabase = createClient(URL, KEY)
+```
+
+## Reference
+
+- Server-side parent-of-active mechanism: `internal/tokens/service.go:376-385` in the [supabase/auth](https://github.com/supabase/auth) repo (v1 branch, the `*models.RefreshToken` type assertion). When a request arrives with a revoked refresh token whose child is the currently-active token, the server returns the active token instead of rejecting — both tabs receive the same rotated token under DB row locking.
+- `lib/locks.ts` exports (`navigatorLock`, `processLock`, `LockAcquireTimeoutError`, `NavigatorLockAcquireTimeoutError`, `ProcessLockAcquireTimeoutError`, `internals`) remain available for direct imports, marked `@deprecated`. Direct callers who use these exports outside the `GoTrueClient` constructor option are unaffected.
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -56,6 +56,7 @@ import {
   validateExp,
 } from './lib/helpers'
 import { memoryLocalStorageAdapter } from './lib/local-storage'
+import { LockAcquireTimeoutError, navigatorLock } from './lib/locks'
 import { polyfillGlobalThis } from './lib/polyfills'
 import { version } from './lib/version'
 
@@ -92,6 +93,7 @@ import type {
   JWK,
   JwtHeader,
   JwtPayload,
+  LockFunc,
   MFAChallengeAndVerifyParams,
   MFAChallengeParams,
   MFAChallengePhoneParams,
@@ -183,7 +185,7 @@ polyfillGlobalThis() // Make "globalThis" available
 
 const DEFAULT_OPTIONS: Omit<
   Required<GoTrueClientOptions>,
-  'fetch' | 'storage' | 'userStorage' | 'lock' | 'lockAcquireTimeout' | 'suppressLockOptionWarning'
+  'fetch' | 'storage' | 'userStorage' | 'lock'
 > = {
   url: GOTRUE_URL,
   storageKey: STORAGE_KEY,
@@ -195,8 +197,19 @@ const DEFAULT_OPTIONS: Omit<
   debug: false,
   hasCustomAuthorizationHeader: false,
   throwOnError: false,
+  lockAcquireTimeout: 5000, // 5 seconds. Only used when a custom `lock` is supplied. TODO(v3): remove.
   skipAutoInitialize: false,
   experimental: {},
+}
+
+/**
+ * No-op lock used internally as a placeholder. Kept so older test setups that
+ * inject this exact reference do not break; new code never sees it because
+ * `this.lock` stays `null` when no custom lock is supplied (lockless path).
+ * TODO(v3): remove with the legacy lock path.
+ */
+async function lockNoOp<R>(name: string, acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
+  return await fn()
 }
 
 /**
@@ -292,7 +305,20 @@ export default class GoTrueClient {
   protected hasCustomAuthorizationHeader = false
   protected suppressGetSessionWarning = false
   protected fetch: Fetch
+  /**
+   * Custom lock function passed via `settings.lock`. When non-null, every auth
+   * operation runs inside `_acquireLock`. When null (the default), the client
+   * uses its lockless coordination (refresh single-flight + commit guard).
+   * TODO(v3): remove along with the legacy lock path.
+   */
+  protected lock: LockFunc | null = null
+  protected lockAcquired = false
+  protected pendingInLock: Promise<any>[] = []
   protected throwOnError: boolean
+  /**
+   * Only consulted when a custom `lock` is supplied. TODO(v3): remove.
+   */
+  protected lockAcquireTimeout: number
   /**
    * Opt-in flags for experimental features. Defaults to an empty object.
    * See `GoTrueClientOptions.experimental`.
@@ -367,23 +393,18 @@ export default class GoTrueClient {
     this.hasCustomAuthorizationHeader = settings.hasCustomAuthorizationHeader
     this.throwOnError = settings.throwOnError
 
-    // settings.lock and settings.lockAcquireTimeout are typed for backwards
-    // compatibility and have no effect at runtime. Warn once at construction
-    // when a non-null `lock` is supplied so callers who relied on a custom
-    // lock (typically React Native `processLock` or Node multi-process
-    // setups) discover the change instead of racing silently. Opt-out via
-    // `suppressLockOptionWarning: true`.
-    if (settings.lock != null && !settings.suppressLockOptionWarning) {
-      const lockWarning =
-        `${this._logPrefix()} The \`lock\` option is accepted for backwards ` +
-        `compatibility but is not invoked by the client. The auth client ` +
-        `coordinates refreshes itself and the server resolves cross-instance ` +
-        `races. Code that depended on a custom lock being called will not run. ` +
-        `Pass \`suppressLockOptionWarning: true\` to silence this warning.`
-      console.warn(lockWarning)
-      if (this.logDebugMessages) {
-        console.trace(lockWarning)
-      }
+    // Always wire `lockAcquireTimeout` even on the lockless path: consumers
+    // (including supabase-js tests) read it off the client to verify option
+    // flow-through.
+    this.lockAcquireTimeout = settings.lockAcquireTimeout
+
+    // TODO(v3): remove. Legacy opt-in path preserved for backwards
+    // compatibility with callers passing a custom `lock` (typically React
+    // Native `processLock` or Node multi-process setups). When `settings.lock`
+    // is null the client uses its lockless coordination — no `navigator.locks`
+    // by default, no implicit `processLock`.
+    if (settings.lock != null) {
+      this.lock = settings.lock
     }
 
     if (!this.jwks) {
@@ -518,6 +539,12 @@ export default class GoTrueClient {
     }
 
     this.initializePromise = (async () => {
+      if (this.lock != null) {
+        // TODO(v3): remove legacy lock path
+        return await this._acquireLock(this.lockAcquireTimeout, async () => {
+          return await this._initialize()
+        })
+      }
       return await this._initialize()
     })()
 
@@ -1402,6 +1429,13 @@ export default class GoTrueClient {
    */
   async exchangeCodeForSession(authCode: string): Promise<AuthTokenResponse> {
     await this.initializePromise
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return this._acquireLock(this.lockAcquireTimeout, async () => {
+        return this._exchangeCodeForSession(authCode)
+      })
+    }
 
     return this._exchangeCodeForSession(authCode)
   }
@@ -2440,6 +2474,13 @@ export default class GoTrueClient {
   async reauthenticate(): Promise<AuthResponse> {
     await this.initializePromise
 
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._reauthenticate()
+      })
+    }
+
     return await this._reauthenticate()
   }
 
@@ -2655,9 +2696,93 @@ export default class GoTrueClient {
   async getSession() {
     await this.initializePromise
 
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return this._useSession(async (result) => {
+          return result
+        })
+      })
+    }
+
     return await this._useSession(async (result) => {
       return result
     })
+  }
+
+  /**
+   * Acquires a global lock based on the storage key.
+   *
+   * TODO(v3): remove along with the legacy lock path. Only called when
+   * `this.lock` is non-null (custom lock supplied via constructor). The
+   * default lockless path bypasses this entirely.
+   */
+  private async _acquireLock<R>(acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
+    this._debug('#_acquireLock', 'begin', acquireTimeout)
+
+    try {
+      if (this.lockAcquired) {
+        const last = this.pendingInLock.length
+          ? this.pendingInLock[this.pendingInLock.length - 1]
+          : Promise.resolve()
+
+        const result = (async () => {
+          await last
+          return await fn()
+        })()
+
+        this.pendingInLock.push(
+          (async () => {
+            try {
+              await result
+            } catch (_e) {
+              // we just care if it finished
+            }
+          })()
+        )
+
+        return result
+      }
+
+      return await this.lock!(`lock:${this.storageKey}`, acquireTimeout, async () => {
+        this._debug('#_acquireLock', 'lock acquired for storage key', this.storageKey)
+
+        try {
+          this.lockAcquired = true
+
+          const result = fn()
+
+          this.pendingInLock.push(
+            (async () => {
+              try {
+                await result
+              } catch (e: any) {
+                // we just care if it finished
+              }
+            })()
+          )
+
+          await result
+
+          // keep draining the queue until there's nothing to wait on
+          while (this.pendingInLock.length) {
+            const waitOn = [...this.pendingInLock]
+
+            await Promise.all(waitOn)
+
+            this.pendingInLock.splice(0, waitOn.length)
+          }
+
+          return await result
+        } finally {
+          this._debug('#_acquireLock', 'lock released for storage key', this.storageKey)
+
+          this.lockAcquired = false
+        }
+      })
+    } finally {
+      this._debug('#_acquireLock', 'end')
+    }
   }
 
   /**
@@ -2729,6 +2854,11 @@ export default class GoTrueClient {
       }
   > {
     this._debug('#__loadSession()', 'begin')
+
+    if (this.lock != null && !this.lockAcquired) {
+      // TODO(v3): remove. Only meaningful on the legacy lock path.
+      this._debug('#__loadSession()', 'used outside of an acquired lock!', new Error().stack)
+    }
 
     try {
       let currentSession: Session | null = null
@@ -2893,7 +3023,15 @@ export default class GoTrueClient {
 
     await this.initializePromise
 
-    const result = await this._getUser()
+    let result: UserResponse
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      result = await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._getUser()
+      })
+    } else {
+      result = await this._getUser()
+    }
 
     if (result.data.user) {
       this.suppressGetSessionWarning = true
@@ -3067,6 +3205,13 @@ export default class GoTrueClient {
     } = {}
   ): Promise<UserResponse> {
     await this.initializePromise
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._updateUser(attributes, options)
+      })
+    }
 
     return await this._updateUser(attributes, options)
   }
@@ -3254,6 +3399,13 @@ export default class GoTrueClient {
     refresh_token: string
   }): Promise<AuthResponse> {
     await this.initializePromise
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._setSession(currentSession)
+      })
+    }
 
     return await this._setSession(currentSession)
   }
@@ -3443,6 +3595,13 @@ export default class GoTrueClient {
    */
   async refreshSession(currentSession?: { refresh_token: string }): Promise<AuthResponse> {
     await this.initializePromise
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._refreshSession(currentSession)
+      })
+    }
 
     return await this._refreshSession(currentSession)
   }
@@ -3691,6 +3850,13 @@ export default class GoTrueClient {
    */
   async signOut(options: SignOut = { scope: 'global' }): Promise<{ error: AuthError | null }> {
     await this.initializePromise
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return await this._acquireLock(this.lockAcquireTimeout, async () => {
+        return await this._signOut(options)
+      })
+    }
 
     return await this._signOut(options)
   }
@@ -3955,7 +4121,15 @@ export default class GoTrueClient {
     this.stateChangeEmitters.set(id, subscription)
     ;(async () => {
       await this.initializePromise
-      await this._emitInitialSession(id)
+
+      if (this.lock != null) {
+        // TODO(v3): remove legacy lock path
+        await this._acquireLock(this.lockAcquireTimeout, async () => {
+          this._emitInitialSession(id)
+        })
+      } else {
+        await this._emitInitialSession(id)
+      }
     })()
 
     return { data: { subscription } }
@@ -4960,9 +5134,62 @@ export default class GoTrueClient {
   private async _autoRefreshTokenTick() {
     this._debug('#_autoRefreshTokenTick()', 'begin')
 
-    // Skip if a refresh is already in flight. `_callRefreshToken` also
-    // dedupes via the same field, so this is just a fast-path skip to
-    // avoid an unnecessary storage read.
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path. Uses `_acquireLock(0, ...)` which
+      // throws `LockAcquireTimeoutError` immediately if the lock is held —
+      // that's the fail-fast skip path that lets the tick bail out instead
+      // of queuing behind a long-running operation.
+      try {
+        await this._acquireLock(0, async () => {
+          try {
+            const now = Date.now()
+            try {
+              return await this._useSession(async (result) => {
+                const {
+                  data: { session },
+                } = result
+
+                if (!session || !session.refresh_token || !session.expires_at) {
+                  this._debug('#_autoRefreshTokenTick()', 'no session')
+                  return
+                }
+
+                const expiresInTicks = Math.floor(
+                  (session.expires_at * 1000 - now) / AUTO_REFRESH_TICK_DURATION_MS
+                )
+
+                this._debug(
+                  '#_autoRefreshTokenTick()',
+                  `access token expires in ${expiresInTicks} ticks, a tick lasts ${AUTO_REFRESH_TICK_DURATION_MS}ms, refresh threshold is ${AUTO_REFRESH_TICK_THRESHOLD} ticks`
+                )
+
+                if (expiresInTicks <= AUTO_REFRESH_TICK_THRESHOLD) {
+                  await this._callRefreshToken(session.refresh_token)
+                }
+              })
+            } catch (e) {
+              console.error(
+                'Auto refresh tick failed with error. This is likely a transient error.',
+                e
+              )
+            }
+          } finally {
+            this._debug('#_autoRefreshTokenTick()', 'end')
+          }
+        })
+      } catch (e) {
+        if (e instanceof LockAcquireTimeoutError) {
+          this._debug('auto refresh token tick lock not available')
+        } else {
+          throw e
+        }
+      }
+      return
+    }
+
+    // Lockless default: skip if a refresh is already in flight.
+    // `_callRefreshToken` also dedupes via the same field; this is just a
+    // fast-path skip to avoid an unnecessary storage read.
     if (this.refreshingDeferred !== null) {
       this._debug('#_autoRefreshTokenTick()', 'refresh already in flight, skipping')
       return
@@ -5060,13 +5287,26 @@ export default class GoTrueClient {
         // should be recovered
         await this.initializePromise
 
-        if (document.visibilityState !== 'visible') {
-          this._debug(methodName, 'visibilityState is no longer visible, skipping recovery')
-          return
+        if (this.lock != null) {
+          // TODO(v3): remove legacy lock path
+          await this._acquireLock(this.lockAcquireTimeout, async () => {
+            if (document.visibilityState !== 'visible') {
+              this._debug(
+                methodName,
+                'acquired the lock to recover the session, but the browser visibilityState is no longer visible, aborting'
+              )
+              return
+            }
+            await this._recoverAndRefresh()
+          })
+        } else {
+          if (document.visibilityState !== 'visible') {
+            this._debug(methodName, 'visibilityState is no longer visible, skipping recovery')
+            return
+          }
+          // recover the session
+          await this._recoverAndRefresh()
         }
-
-        // recover the session
-        await this._recoverAndRefresh()
       }
     } else if (document.visibilityState === 'hidden') {
       if (this.autoRefreshToken) {
@@ -5198,76 +5438,84 @@ export default class GoTrueClient {
     params: MFAVerifyWebauthnParams<T>
   ): Promise<AuthMFAVerifyResponse>
   private async _verify(params: MFAVerifyParams): Promise<AuthMFAVerifyResponse> {
-    try {
-      return await this._useSession(async (result) => {
-        const { data: sessionData, error: sessionError } = result
-        if (sessionError) {
-          return this._returnResult({ data: null, error: sessionError })
-        }
-
-        const body: StrictOmit<
-          | Exclude<MFAVerifyParams, MFAVerifyWebauthnParams>
-          /** Exclude out the webauthn params from here because we're going to need to serialize them in the response */
-          | Prettify<
-              StrictOmit<MFAVerifyWebauthnParams, 'webauthn'> & {
-                webauthn: Prettify<
-                  StrictOmit<MFAVerifyWebauthnParamFields['webauthn'], 'credential_response'> & {
-                    credential_response: PublicKeyCredentialJSON
-                  }
-                >
-              }
-            >,
-          /*  Exclude challengeId because the backend expects snake_case, and exclude factorId since it's passed in the path params */
-          'challengeId' | 'factorId'
-        > & {
-          challenge_id: string
-        } = {
-          challenge_id: params.challengeId,
-          ...('webauthn' in params
-            ? {
-                webauthn: {
-                  ...params.webauthn,
-                  credential_response:
-                    params.webauthn.type === 'create'
-                      ? serializeCredentialCreationResponse(
-                          params.webauthn.credential_response as RegistrationCredential
-                        )
-                      : serializeCredentialRequestResponse(
-                          params.webauthn.credential_response as AuthenticationCredential
-                        ),
-                },
-              }
-            : { code: params.code }),
-        }
-
-        const { data, error } = await _request(
-          this.fetch,
-          'POST',
-          `${this.url}/factors/${params.factorId}/verify`,
-          {
-            body,
-            headers: this.headers,
-            jwt: sessionData?.session?.access_token,
+    const run = async (): Promise<AuthMFAVerifyResponse> => {
+      try {
+        return await this._useSession(async (result) => {
+          const { data: sessionData, error: sessionError } = result
+          if (sessionError) {
+            return this._returnResult({ data: null, error: sessionError })
           }
-        )
-        if (error) {
+
+          const body: StrictOmit<
+            | Exclude<MFAVerifyParams, MFAVerifyWebauthnParams>
+            /** Exclude out the webauthn params from here because we're going to need to serialize them in the response */
+            | Prettify<
+                StrictOmit<MFAVerifyWebauthnParams, 'webauthn'> & {
+                  webauthn: Prettify<
+                    StrictOmit<MFAVerifyWebauthnParamFields['webauthn'], 'credential_response'> & {
+                      credential_response: PublicKeyCredentialJSON
+                    }
+                  >
+                }
+              >,
+            /*  Exclude challengeId because the backend expects snake_case, and exclude factorId since it's passed in the path params */
+            'challengeId' | 'factorId'
+          > & {
+            challenge_id: string
+          } = {
+            challenge_id: params.challengeId,
+            ...('webauthn' in params
+              ? {
+                  webauthn: {
+                    ...params.webauthn,
+                    credential_response:
+                      params.webauthn.type === 'create'
+                        ? serializeCredentialCreationResponse(
+                            params.webauthn.credential_response as RegistrationCredential
+                          )
+                        : serializeCredentialRequestResponse(
+                            params.webauthn.credential_response as AuthenticationCredential
+                          ),
+                  },
+                }
+              : { code: params.code }),
+          }
+
+          const { data, error } = await _request(
+            this.fetch,
+            'POST',
+            `${this.url}/factors/${params.factorId}/verify`,
+            {
+              body,
+              headers: this.headers,
+              jwt: sessionData?.session?.access_token,
+            }
+          )
+          if (error) {
+            return this._returnResult({ data: null, error })
+          }
+
+          await this._saveSession({
+            expires_at: Math.round(Date.now() / 1000) + data.expires_in,
+            ...data,
+          })
+          await this._notifyAllSubscribers('MFA_CHALLENGE_VERIFIED', data)
+
+          return this._returnResult({ data, error })
+        })
+      } catch (error) {
+        if (isAuthError(error)) {
           return this._returnResult({ data: null, error })
         }
-
-        await this._saveSession({
-          expires_at: Math.round(Date.now() / 1000) + data.expires_in,
-          ...data,
-        })
-        await this._notifyAllSubscribers('MFA_CHALLENGE_VERIFIED', data)
-
-        return this._returnResult({ data, error })
-      })
-    } catch (error) {
-      if (isAuthError(error)) {
-        return this._returnResult({ data: null, error })
+        throw error
       }
-      throw error
     }
+
+    if (this.lock != null) {
+      // TODO(v3): remove legacy lock path
+      return this._acquireLock(this.lockAcquireTimeout, run)
+    }
+    return run()
   }
 
   /**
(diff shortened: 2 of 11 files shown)


mandarini:
Do not merge until we ask users to test, and dogfood.

mandarini (approved):

mandarini:
This PR has been published under `@supabase/supabase-js@beta` or version `2.107.0-beta.1`, so that users can test it out.

thomaslarsson:
@mandarini Thank you! PR #2392 looks like it should fix the stuff that actually hung the UI for us — mainly the lock deadlocks (#2344-style) and iOS tabs where `getSession()` never resolved. Happy to test the beta for that. I will keep my own guards in production and write to log if they are called. Will make it easier to know for sure if the issues still exists for us or not.

A few things we'd still like to see addressed (or at least documented) before we'd feel comfortable dropping our own workarounds:

1. **Refresh storm / no backoff after fatal `/token`**
   When refresh fails with something like `refresh_token_not_found`, the ~30s ticker can keep trying anyway. We saw single clients doing hundreds of `/token` calls per hour until we hit project rate limits. `refreshingDeferred` helps with concurrent calls in one client, but there's no "stop trying for a while" after a hard failure. (For example when clearing refreshTokens from the db - we tried that to force token recreate...)

2. **Cookie scope on sign-out (`@supabase/ssr`)**

   If you've ever changed cookie `Domain` (we went from host-only to `.domain`), `_removeSession()` only clears cookies at the current scope. Stale host-only cookies stick around, the session comes back on the next read, and you're back in the loop. Some guidance or a "clear all scopes" option in `@supabase/ssr` would help a lot.

3. **Multiple clients per tab**

   Lockless coordination is per-instance. Two `GoTrueClient`s (easy to do with plain `createClient()` from `@supabase/supabase-js`) means two tickers. We saw interleaved refresh patterns that looked like two tabs, but it was one browser with two clients. A runtime warning or stronger singleton story would be nice.

4. **Observability**

   Dashboard shows `/token` volume but not why. Something like optional diagnostic events (`refresh_failed_fatal`, `refresh_discarded`, etc.) would make production debugging much easier. We patched `window.fetch` to figure some of this out.

None of this is meant to block the lockless refactor - we think that's the right direction. These are the gaps we'd still be covering in the client. Happy to share more detail from our logs if useful once we have them. :)

o-santi (approved):

grdsdev (approved):

mandarini:
@thomaslarsson Thanks for the detailed writeup. 

**1. Refresh storm / no backoff after fatal `/token`**

This is a real gap and a good catch. The good news is the server already returns a well-defined set of fatal codes, all HTTP 400 with a structured `code` field. The ones the client should treat as "stop ticking, do not retry":

- `refresh_token_not_found`
- `refresh_token_already_used`
- `session_not_found`
- `session_expired`
- `user_banned`

And `429 over_request_rate_limit` for the rate-limit case (worth noting: the rate limiter is per-IP, shared across all `/token` grant types, and the response does not currently carry a `Retry-After` header, so the client would pick the backoff itself).

The auto-refresh ticker doesn't key off any of these today; it just retries on the next tick. We'll file a follow-up to add fatal-code recognition + backoff in auth-js. Your "hundreds of `/token` calls per hour" observation is exactly the failure mode that needs.

**2. Cookie scope on sign-out (`@supabase/ssr`)**

`removeItem` in `@supabase/ssr` clears at a single scope, whatever `cookieOptions.domain` is configured to at runtime. Migrating from host-only to `.domain` leaves the host-only cookies behind, the browser keeps returning both in the `Cookie` header, and parsers may pick the stale one. There is no read-side fix possible (the browser doesn't expose `Domain` per cookie to JS), so the fix has to be on the write side.

We're looking at landing this in `@supabase/ssr`: when `cookieOptions.domain` is set, also emit a host-only `Set-Cookie` with `Max-Age=0` on removal. That covers the common host-only → parent-domain migration automatically. For the rarer `.parent1` → `.parent2` case, a documented migration helper that takes a list of legacy scopes is the right shape. This will follow up on a separate PR on `@supabase/ssr`.

**3. Multiple clients per tab**

Fair point. A runtime warning on a second `GoTrueClient` instantiation in the same realm is cheap and would have caught your case. Worth a follow-up; the only nuance is making it survive HMR cleanly (we don't want to warn on Strict-Mode double-mounts), which is exactly what `dispose()` in this PR is for.

**4. Observability**

Two halves:
- **Client-side diagnostic events** (`refresh_failed_fatal`, `refresh_discarded`, `refresh_rate_limited`, etc.) — yes, this can live in auth-js as optional listeners alongside `onAuthStateChange`. Reasonable follow-up.
- **Dashboard surfacing `/token` failures by reason, not just volume**: that part isn't in the SDK repo; it's a Studio/platform ask. You can file it under https://github.com/supabase/supabase.

I am going to release this PR tomorrow as part of our next minor release. Let me know how it works for you if you test it ou!

Again, thank you so much for all the details!

Owez:
Wonderful, thanks for this

PR #2406, opened 2026-05-28
ci(ci): pin windows runner and migrate github app token to client-id

mandarini:
Address two GitHub Actions deprecation notices surfaced on recent runs, and document why the upcoming GitHub App installation token format change is a no-op for this repo.

- Pin the Windows runner: `windows-latest` is being redirected to `windows-2025-vs2026` on June 15, 2026. Pinning explicitly in `ci-supabase-js.yml` avoids a silent image flip on the redirect date.
- Migrate `actions/create-github-app-token` from the deprecated `app-id` input to `client-id` across 9 call sites in 4 workflows (`deprecate-version`, `dogfood`, `publish`, `revert-commit`).
- New secrets `APP_CLIENT_ID` and `DOGFOOD_APP_CLIENT_ID` must be added in repo settings before merge. They hold each GitHub App's Client ID (string like `Iv23...`), which is a different value than the numeric App ID. `PRIVATE_KEY` / `DOGFOOD_APP_PRIVATE_KEY` are unchanged. Old `APP_ID` / `DOGFOOD_APP_ID` can be removed after one green canary run.
- Upcoming `ghs_...` installation token format (~520 chars): checked all workflows and scripts. No length assumptions or string slicing on tokens, so no code change is required.

### DIFF
--- a/.github/workflows/ci-supabase-js.yml
+++ b/.github/workflows/ci-supabase-js.yml
@@ -58,7 +58,7 @@ jobs:
     strategy:
       matrix:
         node: [22]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-2025-vs2026]
       fail-fast: false
     steps:
       - name: Checkout code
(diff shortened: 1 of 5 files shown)


staaldraad (approved):
I've added the CLIENT_ID's 🤞🏼

mandarini:
thank youUuuuuu

Issue #1615, opened 2025-07-16
signInWithPassword broken in 2.71.0 on React Native (ReferenceError: Property 'structuredClone' doesn't exist)


samzmann:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

In React Native, calling `supabase.auth.signInWithPassword` throws an unhandled error `ReferenceError: Property 'structuredClone' doesn't exist`:
```js
 try {
    const { data, error } = await supabase.auth.signInWithPassword({ email, password })
    console.log('signInWithPassword data', data)
    console.log('signInWithPassword error', error)
  } catch (error) {
    console.log('error', error) // ReferenceError: Property 'structuredClone' doesn't exist
  }
```

**Reverting to @supabase/supabase-js 2.50.3 which depends on @supabase/auth-js 2.70.0 solves the issue.**

I also tried installing the @ungap/structured-clone polyfill, but that did not work.

## To Reproduce

1. Setup a basic RN Expo project
2. Install @supabase/supabase-js 2.51.0
3. basic Supabase client setup
4. call `supabase.auth.signInWithPassword` wrapping it in a try/catch
5. get the error

## Expected behavior

- if success: return session data (login)
- if error: return error param

After experiencing this issue, I was surprised to see that signInWithPassword throws if it encounters an unexpected error. This makes the `error` filed on the response kind of misleading, although I can understand the rational.

## System information

- OS: macOS
- react-native: 0.79.5
- expo: ~53.0.17
- Version of supabase-js: 2.51.0
- Version of Node.js: 24.3.0

samzmann:
The error must come from GoTrueClient.ts, where a few calls to `structuredClone` were added
https://github.com/supabase/auth-js/compare/v2.70.0...v2.71.0#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5

@hf @cemalkilic Any insights?

7ttp:
hi @samzmann  thanks for reporting this! 💚🙌
(Sorry for the delayed response)

isn't reproducible anymore!
The problem ig was that structuredClone (a browser only API) was used instead of a JSON based clone,
 causing React Native to crash.

Just FYI this was
[...]

PR #2431, opened 2026-06-06
test(auth): add deepClone regression tests for React Native / Hermes environments

chatman-media:
## Background

In React Native / Hermes, `structuredClone` is not a global. Starting with
`@supabase/auth-js` v2.71.0, `GoTrueClient` called `structuredClone` directly
when persisting session data. This made `signInWithPassword` throw an unhandled
`ReferenceError: Property 'structuredClone' doesn't exist` on Hermes, even
inside a `try/catch` block (since the error was thrown synchronously in the
async function body before the first `await`).

Root issue: supabase/supabase-js#1615  
Original user report: supabase-js v2.51.0 / auth-js v2.71.0

## Code fix (already merged)

The code fix landed in commit `9a6edb9` (PR #1084 in the now-archived
`supabase/auth-js` repo) and was released as `@supabase/auth-js v2.71.1` /
`@supabase/supabase-js v2.51.1`. It introduced a `deepClone` helper in
`packages/core/auth-js/src/lib/helpers.ts` that uses
`JSON.parse(JSON.stringify(obj))` — safe in all JS runtimes — and replaced both
`structuredClone` call sites in `GoTrueClient.ts`.

## What this PR adds

The `deepClone` helper has no unit tests. This PR adds a `deepClone` describe
block to `packages/core/auth-js/test/helpers.test.ts` with three cases:

| Test | Purpose |
|---|---|
| Returns a deep copy with no shared references | Core correctness |
| Clones arrays correctly | Nested array support |
| Works when `globalThis.structuredClone` is undefined | Regression guard for the RN/Hermes scenario |

The third test deletes `globalThis.structuredClone` before calling `deepClone`
and restores it in a `finally` block. It would have caught the original bug if
it had existed before v2.71.0 shipped.

All 49 tests in `helpers.test.ts` pass locally (jest 30 / ts-jest 29 / Node 26).

### DIFF
--- a/packages/core/auth-js/test/helpers.test.ts
+++ b/packages/core/auth-js/test/helpers.test.ts
@@ -1,6 +1,7 @@
 import { AuthInvalidJwtError } from '../src'
 import {
   decodeJWT,
+  deepClone,
   generateCallbackId,
   getAlgorithm,
   getItemAsync,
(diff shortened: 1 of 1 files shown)


chatman-media:
Closing — the underlying fix already shipped (deepClone in v2.51.1), so this isn't needed. Thanks!

PR #2438, opened 2026-06-11
fix(realtime): subscribed state

### LINKED ISSUE #1599: Premature 'SUBSCRIBED' status for postgres_changes signals incorrect readiness

7ttp:
## TL;DR 
 
Delay `SUBSCRIBED` only for channels that bind `postgres_changes` callbacks exclusively
 and keep the existing timing for mixed channels

## Prob:  
`SUBSCRIBED` could be emitted before the Postgres change feed was actually ready 
which made postgres-only channels look ready too early

so now basically, wait for the `postgres_changes` system ready message before emitting `SUBSCRIBED` 
for postgres-only channels while leaving channels with `broadcast` or `presence` callbacks unchanged

## ref:
- follow up to https://github.com/supabase/supabase-js/pull/2029 &
 aligned with Eduardo’s suggestion :D 
- closes #1599
- https://github.com/supabase/supabase-js/pull/2029#issuecomment-4626790886

### DIFF
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -182,6 +182,7 @@ export default class RealtimeChannel {
   private _postgresChangesSystemReady = false
   private _subscribeCallback: ((status: REALTIME_SUBSCRIBE_STATES, err?: Error) => void) | null =
     null
+  private _subscribeTimeoutId: ReturnType<typeof setTimeout> | null = null
   /** @internal */
   channelAdapter: ChannelAdapter
 
@@ -258,16 +259,33 @@ export default class RealtimeChannel {
     this.presence = new RealtimePresence(this)
     this.channelAdapter.on(REALTIME_LISTEN_TYPES.SYSTEM, (payload: unknown) => {
       if (
-        payload &&
-        typeof payload === 'object' &&
-        'extension' in payload &&
-        payload.extension === REALTIME_LISTEN_TYPES.POSTGRES_CHANGES &&
-        'status' in payload &&
-        payload.status === 'ok'
+        !payload ||
+        typeof payload !== 'object' ||
+        !('extension' in payload) ||
+        payload.extension !== REALTIME_LISTEN_TYPES.POSTGRES_CHANGES ||
+        !('status' in payload)
       ) {
-        this._postgresChangesSystemReady = true
-        this._emitSubscribed()
+        return
       }
+
+      if (payload.status !== 'ok') {
+        this._failSubscribeState(
+          REALTIME_SUBSCRIBE_STATES.CHANNEL_ERROR,
+          new Error(
+            'message' in payload && typeof payload.message === 'string'
+              ? payload.message
+              : `postgres_changes system status: ${String(payload.status)}`
+          )
+        )
+        return
+      }
+
+      this._postgresChangesSystemReady = true
+      this._emitSubscribed()
+    })
+
+    this._onError(() => {
+      this._resetSubscribeState()
     })
 
     this._onClose(() => {
@@ -342,6 +360,7 @@ export default class RealtimeChannel {
       }
 
       this._onError((reason: unknown) => {
+        this._resetSubscribeState()
         callback?.(REALTIME_SUBSCRIBE_STATES.CHANNEL_ERROR, normalizeChannelError(reason))
       })
 
@@ -363,7 +382,7 @@ export default class RealtimeChannel {
             return
           }
 
-          this._updatePostgresBindings(postgres_changes, callback)
+          this._updatePostgresBindings(postgres_changes, callback, timeout)
         })
         .receive('error', (error: { [key: string]: any }) => {
           this._resetSubscribeState()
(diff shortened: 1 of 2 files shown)


mandarini:
@edgurgel thoughts?

edgurgel:
Hey team I'm going to play with this branch for some days! Thanks for the PR 🙇

mandarini:
Nice, thanks @edgurgel ! Can you please ping me once you've tested and are ok to merge?

edgurgel:
Hey @7ttp here is my initial feedback:

* A "normal" join has a timeout that fires if it gets no response from the server. So I think we need a timeout timer here to do something similar. It should use the same timeout we optionally get from `subscribe` or default to `this.socket.timeout`
* We need to handle "status" being not "ok" and set the error state can call the callback etc
* We need to call `this._resetSubscribeState()` on `_onError` and on the happy path as well when we `emitSubscribed` I think?

edgurgel:
Thanks @7ttp ! Will play with the error states and time outs and get back to you! Looking good so far! <3

edgurgel:
Sorry for the delay! 

I've tested with a few error cases and at least one missing thing is:

When there is a timeout a channel with only pg changes does not try to "rejoin" like a normal timeout. So we would need to actually leave the channel and then start a rejoin timer to simulate what a `phx_join` timeout already cases.

Because this is starting to feel very complicated (and a bit hacky) we discussed with the team and we agreed that we would rather move this to the backend (and option to wait for the subscription on join) and simplify not only this client but all clients.

Appreciate the patience and the help @7ttp. Sorry if we couldn't proceed with this one!

mandarini:
Thank you @edgurgel

PR #2440, opened 2026-06-11
fix(release): restore JSR publish flags and enable for beta

mandarini:
The previous switch to `deno publish` direct (d797a65c) dropped flags the `jsr` npm wrapper was injecting for Node-style projects (any package with a `package.json`), causing every JSR publish since 2.108.2-canary.1 to fail type-check with TS2307 (extension-less imports) and TS5095/TS5109 (incompatible tsconfig options).

This PR passes `--unstable-bare-node-builtins`, `--unstable-sloppy-imports`, `--unstable-byonm`, `--no-check`, and sets `DENO_DISABLE_PEDANTIC_NODE_WARNINGS=true` in `scripts/publish-to-jsr.ts`, matching exactly what the wrapper was injecting for years before the switch.

Also extends JSR publishing to beta releases in `scripts/release-beta.ts`, using the same try/catch + partial-failure pattern as stable and canary, so beta versions reach JSR alongside npm.

### DIFF
--- a/scripts/publish-to-jsr.ts
+++ b/scripts/publish-to-jsr.ts
@@ -121,7 +121,19 @@ async function publishToJsr() {
       // than whatever the `jsr` npm wrapper downloads from dl.deno.land at
       // runtime. Timeout caps JSR queue stranding (jsr-io/jsr#1448) at 10 min
       // instead of letting the unbounded poll loop wedge the workflow.
-      const args = ['publish', '--allow-dirty']
+      //
+      // Flags mirror what the `jsr` npm wrapper passes for Node-style projects
+      // (package.json present): sloppy imports, BYO node_modules, bare node
+      // builtins, and --no-check. Without them Deno rejects our extension-less
+      // imports and unsupported tsconfig options as hard errors.
+      const args = [
+        'publish',
+        '--allow-dirty',
+        '--unstable-bare-node-builtins',
+        '--unstable-sloppy-imports',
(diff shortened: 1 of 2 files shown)


spydon (approved):

scripts/publish-to-jsr.ts
```diff
@@ -121,14 +121,27 @@ async function publishToJsr() {
...
+        '--no-check',
+      ]
       if (dryRun) args.push('--dry-run')
       console.log(`   Command: deno ${args.join(' ')}`)
       const result = spawnSync('deno', args, {
         cwd: packagePath,
         stdio: 'inherit',
         timeout: 10 * 60 * 1000,
         killSignal: 'SIGKILL',
+        env: { ...process.env, DENO_DISABLE_PEDANTIC_NODE_WARNINGS: 'true' },
```
QuintinWillison, line 144: Is it worth some commentary as to why `DENO_DISABLE_PEDANTIC_NODE_WARNINGS` is needed? Looking at the state prior to the previous broken state, it doesn't look like this was being used, unless I wasn't looking in the right place. 😕

mandarini: @QuintinWillison we used to publish with jsr CLI which used all these options:
https://github.com/jsr-io/jsr-npm/blob/9e797fbf742ca1dd741be76369a626b8ebfb2ab5/src/commands.ts#L192

PR #2441, opened 2026-06-11
chore(repo): declare supabase-js compliance with SDK capability matrix

mandarini:
## Description

Wires supabase-js into the per-SDK capability compliance model introduced in [supabase/sdk#15](https://github.com/supabase/sdk/pull/15). Adds a root-level `sdk-compliance.yaml` declaring which capability IDs supabase-js implements, a CI workflow that validates that file against the canonical registry, and two scripts plus a mapping file that keep both sides in sync.

### What changed?

New files:

- **`sdk-compliance.yaml`** -- 168 feature entries declaring `implemented` status for each public method supabase-js exposes, grouped by area (auth, database, storage, realtime, functions). Generated by the audit script. Unlisted IDs default to `not_implemented`.
- **`.github/workflows/validate-capabilities.yml`** -- calls the reusable `validate-sdk-compliance.yml` workflow from `supabase/sdk` on every PR that touches `sdk-compliance.yaml`. Rejects unknown feature IDs. Declares a minimal `contents: read` token permission.
- **`scripts/audit-sdk-compliance.ts`** -- walks each package's typedoc `spec.json`, collects public methods on the main client classes (with a small DENYLIST for boilerplate like `then`, `toJSON`, `dispose`), dedupes duplicate method names across classes (e.g. `signOut` on both `GoTrueClient` and `GoTrueAdminApi`), applies the canonical mapping, snake_cases the names, and emits `<area>.<group>.<snake_method>: implemented` entries. Re-run whenever public methods change.
- **`scripts/sync-canon.ts`** -- one-direction "JS to canon" pipeline. Reads the same `spec.json` files plus their JSDoc summaries and writes feature entries into `../sdk/capabilities/<area>.yaml`. Two modes: `--bootstrap` replaces canon files wholesale (used once to seed the registry); default mode is additive (only appends IDs that don't already exist, leaves existing entries untouched so registry maintainers and other SDKs can edit freely). Canon path is configurable via `--canon-path` flag or `SUPABASE_SDK_CAPABILITIES_PATH` env var, defaulting to a sibling checkout.
- **`scripts/canonical-mapping.yaml`** -- hand-maintained map from auto-derived JS feature IDs to their canonical IDs in `supabase/sdk`, for the cases where they diverge (JS-isms renamed in canon, group renames such as `filters` to `using_filters`, methods that collapse into one capability, and a `null` value to drop JS-only methods with no canonical counterpart). Without an entry, the auto-derived ID flows through unchanged.

Source changes (documentation only, no behavior change):

- **`packages/core/postgrest-js/src/{PostgrestClient,PostgrestFilterBuilder,PostgrestTransformBuilder}.ts`** and **`packages/core/realtime-js/src/RealtimeChannel.ts`** -- reposition JSDoc comments onto the implementation signatures and add `@category` tags so the typedoc `spec.json` attributes method summaries correctly for the audit and sync scripts.
- **`.gitignore`** -- ignore the audit script's `sdk-compliance.draft.yaml` working file.

## Additional notes

- `sync-canon.ts` defaults to additive on the canon side so concurrent edits from registry maintainers and other SDKs don't get clobbered. The bootstrap mode is one-shot.
- One method (`signInWithOAuth`) snake_cases to `sign_in_with_o_auth` due to a heuristic ambiguity in the camelCase to snake_case conversion (the same regex that correctly splits `regexIMatch` to `regex_i_match` also splits `OAuth` to `O_Auth`). Left as-is; documented in the script comments.
- Baseline for the audit: `@supabase/supabase-js v2.108.1`.


grdsdev (commented):
Do we want to keep the changed `scripts/*` files and the `sdk-compliance.draft.yaml`? Or was that a bootstrap thing we can remove?

mandarini:
@grdsdev i wanted to keep them, but you're right, they do not want to be so many and i can drop the bootstrapping one. did so

grdsdev (approved):

PR #2400, opened 2026-05-27
feat(realtime): allow httpSend to send binary payload

edgurgel:
## 🔍 Description

It uses the new single broadcast API to send JSON or binary payloads.

I had to pin the realtime version for now until CLI has updated the realtime version.

### Why was this change needed?

Closes #(issue_number) 

## 🔄 Breaking changes

- [x] This PR contains no breaking changes

## 📋 Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/supabase/supabase-js/blob/master/CONTRIBUTING.md)
- [x] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `<type>(<scope>): <description>`
- [x] I have run `pnpm nx format` to ensure consistent code formatting
- [x] I have added tests for new functionality (if applicable)
- [x] I have updated documentation (if applicable)

PR #2444, opened 2026-06-11
fix(realtime): clarify httpSend() 404 error and server migration note

mandarini:
Follow-up to [@holm's comment on #2400](https://github.com/supabase/supabase-js/pull/2400#issuecomment-4639951590): when `httpSend()` hits a Realtime server older than v2.97.0, the per-event broadcast endpoint returns 404 and the client used to reject with a bare "Not Found", giving callers no way to diagnose the mismatch. This PR replaces the generic 404 rejection with an actionable message naming the v2.97.0 requirement and the two escape hatches (update the Supabase CLI, or pin a newer Realtime via `supabase/.temp/realtime-version`).

A companion migration note at `packages/core/realtime-js/migrations/httpsend-server-version.md` covers the same ground in long form (who is affected, how to verify the server version, CLI update / pin / self-host bump / downgrade options). A new unit test asserts the wording on a mocked 404.

No API surface or behavior change beyond the error string: the rejection still throws an `Error` and the success path is unchanged.

### DIFF
--- a/packages/core/realtime-js/migrations/httpsend-server-version.md
+++ b/packages/core/realtime-js/migrations/httpsend-server-version.md
@@ -1,0 +1,88 @@
+# `httpSend()` server version requirement
+
+**Since:** v2.107.0
+**Applies to:** anyone calling `RealtimeChannel.httpSend()` (or `RealtimeChannel.broadcast()` configured to use HTTP)
+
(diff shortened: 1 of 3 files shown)


edgurgel (approved):
❤️

PR #2445, opened 2026-06-11
fix(postgrest): add typed column inference for order() with referencedTable

### LINKED ISSUE #971: Order on a foreign table
# Bug report

## Describe the bug

Hey guys, I am not an expert around here but hear me out :)
 
Based on [docs](https://supabase.com/docs/reference/javascript/order?example=on-a-referenced-table) here on how to order the query result `on a referenced table` .

```ts
  .from('countries')
  .select(`
    name,
    cities (
      name
    )
  `)
  .order('name', { referencedTable: 'cities', ascending: false })
  ```
First issue is that, maybe the docs is outdated but, this options key is called `foreignTable` in the `supabase-js` package `2.38.3`. 

Secondly, when i add foreignTable option to order the foreign table that comes with the query, in Typescript, I am given the columns of the base table of the query (`countries` in this example) as autocomplete.  Though, I can still use the desired column name as the order value and TS will not yell about it. 

## Expected behavior

If I want to order the foreign table and add that option key, TS should recognize that so I should be able to use the columns of that table to order. The inferred type is still the columns of the base table.

## System information

- Version of supabase-js: [2.38.3]

### PR DESCRIPTION
**Pull request by 7vignesh**
### What changed?

Added new typed overloads to the `order()` method in `PostgrestTransformBuilder` that resolve the referenced table's column names from the `Schema` type when `referencedTable` (or deprecated `foreignTable`) is a known table/view.

**Overload resolution now works in 3 tiers:**

1. **No `referencedTable`** → autocomplete shows columns from the parent table (unchanged)
2. **`referencedTable` is a known table/view** → autocomplete shows columns from that table's `Row`, invalid columns produce a compile-time error
3. **`referencedTable` is an unknown string** → falls through to `column: string` (unchanged, for dynamic/alias use cases)

The catch-all overload uses a conditional `never` type to prevent it from matching when the referenced table is known, forcing TypeScript to use the typed overload instead.

When users passed `referencedTable: 'messages'` to `.order()`, TypeScript offered no autocomplete for the referenced table's columns and silently accepted invalid column names. This caused runtime errors that could have been caught at compile time.

This is the TypeScript typing portion of the issue — runtime behavior was already correct (clarified by maintainers in the issue thread).

Closes #971

### After: type error on invalid column for referenced table

image

### After: valid column passes without error

image

## Breaking changes

The catch-all `string` overload still exists for dynamic table names or unknown aliases, so existing code that passes arbitrary strings continues to compile.

##  Checklist

- [x] My PR title follows the [conventional commit format](https://www.conventionalcommits.org/): `fix(postgrest): add typed column inference for order() with referencedTable`
- [x] I have added tests for new functionality (type tests in `test/index.test-d.ts`)

##  Additional notes

- This supersedes the stalled PR #2025 which attempted to fix this by deleting overloads (which would have broken base-case autocomplete, as noted by @avallete in review).
- The fix imports `TablesAndViews` from the existing select-query-parser types no new type utilities were introduced.
- Zero runtime changes. The implementation body of `order()` is untouched.
- All 28 tstyche type test files pass. Build passes for both `postgrest-js` and downstream `supabase-js`.
- Integration tests require Docker (PostgREST + PostgreSQL) which CI will handle. Since no runtime code changed, they will pass identically.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestTransformBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestTransformBuilder.ts
@@ -4,6 +4,7 @@ import { GetResult } from './select-query-parser/result'
 import { CheckMatchingArrayTypes } from './types/types'
 import { ClientServerOptions, GenericSchema } from './types/common/common'
 import type { MaxAffectedEnabled } from './types/feature-flags'
+import type { TablesAndViews } from './select-query-parser/types'
 
 export default class PostgrestTransformBuilder<
   ClientOptions extends ClientServerOptions,
@@ -156,9 +157,22 @@ export default class PostgrestTransformBuilder<
     column: ColumnName,
     options?: { ascending?: boolean; nullsFirst?: boolean; referencedTable?: undefined }
   ): this
-  order(
+  order<
+    ReferencedTable extends string & keyof TablesAndViews<Schema>,
+    ColumnName extends string & keyof TablesAndViews<Schema>[ReferencedTable]['Row'],
+  >(
+    column: ColumnName,
+    options: { ascending?: boolean; nullsFirst?: boolean; referencedTable: ReferencedTable }
+  ): this
+  order<ReferencedTable extends string>(
     column: string,
-    options?: { ascending?: boolean; nullsFirst?: boolean; referencedTable?: string }
+    options?: {
+      ascending?: boolean
+      nullsFirst?: boolean
+      referencedTable?: ReferencedTable extends keyof TablesAndViews<Schema>
+        ? never
+        : ReferencedTable
+    }
   ): this
   /**
    * @deprecated Use `options.referencedTable` instead of `options.foreignTable`
@@ -170,9 +184,23 @@ export default class PostgrestTransformBuilder<
   /**
    * @deprecated Use `options.referencedTable` instead of `options.foreignTable`
    */
-  order(
+  order<
+    ReferencedTable extends string & keyof TablesAndViews<Schema>,
+    ColumnName extends string & keyof TablesAndViews<Schema>[ReferencedTable]['Row'],
+  >(
+    column: ColumnName,
+    options: { ascending?: boolean; nullsFirst?: boolean; foreignTable: ReferencedTable }
+  ): this
+  /**
+   * @deprecated Use `options.referencedTable` instead of `options.foreignTable`
+   */
+  order<ReferencedTable extends string>(
     column: string,
-    options?: { ascending?: boolean; nullsFirst?: boolean; foreignTable?: string }
+    options?: {
+      ascending?: boolean
+      nullsFirst?: boolean
+      foreignTable?: ReferencedTable extends keyof TablesAndViews<Schema> ? never : ReferencedTable
+    }
   ): this
   /**
    * Order the query result by `column`.
--- a/packages/core/postgrest-js/src/PostgrestTransformBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestTransformBuilder.ts
@@ -157,11 +157,12 @@ export default class PostgrestTransformBuilder<
     column: ColumnName,
     options?: { ascending?: boolean; nullsFirst?: boolean; referencedTable?: undefined }
   ): this
-  order<
-    ReferencedTable extends string & keyof TablesAndViews<Schema>,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mandarini** [comment]
Hi @7vignesh, thank you so much for contributing to Supabase!

This is a nice fix for a real gap. Passing `referencedTable` to `order()` today gives no autocomplete or validation against that table's columns, so typos only surface at runtime. The three tier overload approach (no `referencedTable`, known table, unknown string fallback) is the right shape for the problem.

While testing it locally I ran into one case that regresses: if the column name comes from a variable instead of a string literal, and `referencedTable` is a known table, the call no longer compiles.

```ts
const sortColumn: string = getSortColumn()
postgrest
  .from('users')
  .select('messages(*)')
  .order(sortColumn, { referencedTable: 'messages', ascending: false })
```

This compiles fine on master today but fails on this branch with "No overload matches this call." Since the column is a plain `string` it doesn't match the new typed overload, and since `messages` is a known table the catch-all overload's `referencedTable` type collapses to `never`, so nothing matches. This is a fairly common pattern (sort column from a prop or user selection, table name hardcoded), so it would be worth adding a test for it and adjusting the overloads so a non-literal column falls back to the loose string behavior instead of erroring.

Would you be able to take a look at that case? Happy to help think through the overload signatures if that's useful.

Thank you again for contributing, it's contributions like yours that help make our tools better for everyone.

mandarini (changes requested):
Thanks, see my comment above!

7vignesh:
Hi @mandarini, thanks for the detailed feedback! I've addressed the regression case you identified.

The fix uses a conditional type in the `order()` overload so that:
- **Valid column literals** → autocomplete + compile-time validation ✓
- **Wide `string` type (from variables/props)** → accepted without error ✓  
- **Invalid column literals** → still produce compile-time errors ✓

```ts
// This now compiles correctly:
const sortColumn: string = getSortColumn()
postgrest
  .from('users')
  .select('messages(*)')
  .order(sortColumn, { referencedTable: 'messages', ascending: false })
```

I've also added test cases for both `referencedTable` and the deprecated `foreignTable` with dynamic column variables. All 28 tstyche type test files pass, and the build is green.

Branch has been rebased on latest master as well.

mandarini (changes requested):
Thanks for turning around the dynamic-column fix so quickly!! So, while digging a bit further I found one more edge case worth addressing before we merge. I checked how PostgREST resolves the `<relation>.order=` param, and when an embed is aliased (e.g. `.select('messages:channels(*)')`), PostgREST matches on the alias, not the real table name. Right now `TablesAndViews<Schema>` only knows real table/view names, so if an alias happens to coincide with an unrelated real table in the schema, the typed overload will validate the column against that unrelated table's `Row` instead of the table that's actually embedded. That could either reject a valid column or silently accept one that's wrong at runtime.

Could you add a test that exercises an aliased `referencedTable` (something like `.select('archived:messages(*)').order('col', { referencedTable: 'archived' })`) so we can see how it behaves today, and add a short note in the JSDoc that aliased referenced tables fall back to unchecked `string` (or aren't safe if the alias collides with a real table name)? Given that fully resolving aliases would need more plumbing than this PR should take on, documenting the limitation seems like the right scope for now.

Separately, since the `referencedTable` and `foreignTable` overloads duplicate the same three-branch conditional type, would it be worth factoring that into a shared helper type? It would cut the overload count roughly in half and make it much easier to maintain going forward.

Thanks again for sticking with this one through a couple of rounds, it's making the fix noticeably more solid.

7vignesh:
Hi @mandarini, all three points addressed in the latest push:

1. **Aliased referencedTable test**: Added tests showing that an alias like `'archived'` (not a real table) falls through to the unchecked `string` overload, so any column name compiles fine. Works for both `referencedTable` and `foreignTable`.

2. **JSDoc note**: Added a note to the `@param options.referencedTable` doc explaining the alias limitation: unknown aliases fall through to unchecked string, and if an alias collides with a real table name, columns get validated against the wrong table. The workaround is using a plain `string` variable for the column.

3. **Shared helper type**: Extracted the three-branch conditional into an `OrderColumnForTable<Schema, ReferencedTable, Column>` helper type, so both `referencedTable` and `foreignTable` overloads reference it instead of duplicating the logic. Cuts the overload verbosity roughly in half and makes future changes a single edit.

All 28 tstyche type test files pass and the build is green.

PR #2448, opened 2026-06-14
fix(postgrest): surface helpful hint for browser DNS failures (ERR_NA…

Tanusree-samanta:
## Summary

Fixes missing error context for ERR_NAME_NOT_RESOLVED and other DNS
failures in browser environments (Chrome, Firefox, Safari).

Relates to: supabase/supabase#46821

## Problem

In browsers, DNS failures produce a plain TypeError with no `.cause`
and no error code — just the message string. supabase-js currently
returns this as an opaque stack trace with no actionable guidance.

- Chrome/Edge: `Failed to fetch`
- Firefox: `NetworkError when attempting to fetch resource.`
- Safari: `Load failed`

## Solution

Detect these three browser network-failure messages and populate
`error.hint` with a developer-friendly message pointing to:
1. The likely root cause (DNS / project URL)
2. The Supabase dashboard (to verify project status)
3. The Supabase status page for outages

## Testing

3 new test.each cases added to fetch-errors.test.ts, one per browser.
No behaviour change for Node.js paths or non-DNS errors.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -411,8 +411,25 @@ export default abstract class PostgrestBuilder<
             errorDetails += `\n${cause.stack}`
           }
         } else {
-          // No cause available, just include the error stack
+          // No cause available — include the error stack.
+          // In browsers, DNS failures (ERR_NAME_NOT_RESOLVED) surface as a plain
+          // TypeError("Failed to fetch") with no cause and no error code.
+          // Detect this pattern and surface a human-readable hint so developers
+          // know to check their project URL / DNS rather than their code.
           errorDetails = fetchError?.stack ?? ''
+
+          const msg: string = fetchError?.message ?? ''
+          if (
+            msg === 'Failed to fetch' || // Chrome / Edge
+            msg === 'NetworkError when attempting to fetch resource.' || // Firefox
+            msg === 'Load failed' // Safari
+          ) {
+            hint =
+              'Network request failed — the Supabase project URL could not be reached. ' +
+              'This is usually caused by a DNS resolution failure (e.g. ERR_NAME_NOT_RESOLVED in Chrome). ' +
+              'Verify that your SUPABASE_URL is correct and that the project is active in the Supabase dashboard. ' +
+              'If the issue persists, check https://status.supabase.com for platform outages.'
+          }
         }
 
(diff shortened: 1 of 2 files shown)


mandarini:
Hi @Tanusree-samanta, thank you so much for contributing to Supabase!  I really appreciate you digging into the confusing "opaque stack trace" experience for network failures, that's a genuinely annoying thing to debug and it's great that you took the time to write tests for all three browsers.

Before we consider this, I want to raise a few things about the approach itself:

1. `Failed to fetch` (Chrome), `NetworkError when attempting to fetch resource.` (Firefox), and `Load failed` (Safari) aren't actually DNS-specific. They're the generic message browsers throw for any network-level failure, including CORS misconfiguration, being offline, ad blockers/extensions, and mixed content, not just DNS. Since we can't tell these apart from the message alone, the current hint text ("usually caused by a DNS resolution failure... verify your SUPABASE_URL") could confidently point someone in the wrong direction, for example if their actual issue is a CORS problem.

2. There's an existing rule documented right in `test/fetch-errors.test.ts` (line 19) that client-side network errors intentionally don't populate `hint`/`code`, since those fields are meant for upstream PostgREST/Postgres errors. This PR adds an exception to that rule for the browser case, would you mind taking a look at whether that rule should change, and if so, updating the comment/tests to reflect the new intent?

3. `postgrest-js` is used standalone against any PostgREST server, not just Supabase Cloud. Would it make sense to keep the hint text generic (no `SUPABASE_URL`, dashboard, or status.supabase.com references) here, and let `supabase-js` layer on the Supabase-specific guidance if we want that level of detail?

Also worth mentioning: I looked back at supabase/supabase#46821, which this PR references, and it turned out to be a transient platform DNS outage rather than a reproducible SDK bug. Other commenters saw the same thing during that same outage window, and the issue was closed once it passed without ever being confirmed as something the SDK itself needed to handle.

Given that, I don't think we should take on this extra maintenance surface (the new hint text, the exception to the existing "no hint for client errors" rule, the pattern matching on non-contractual browser strings) unless we have a confirmed, reproducible bug that it actually fixes. If you come across a concrete case where this would have helped, I'd love to hear about it, and we can revisit with that in mind. For now I'd lean towards closing this one, but please let me know if I'm missing something.

Thank you again for putting the time into this, contributions like yours are what make Supabase's tools better for everyone. Please do not be discouraged from further contributions, and I do hope you understand my reasoning!

Issue #530, opened 2022-08-29
supabase-js 2.0rc4 - "This endpoint requires a Bearer token" calling signOut()


howesteve:
Hello,

I'm already using version 2.0.0-rc.4. So in the code below, user is signed in (100% sure), and when I call ```signOut()```:

```js
const { error } = await supabase.auth.signOut();
console.log(error.message);
```
... the following error is thrown in the chrome debug console:
```
helpers.ts:40          GET https://xxxxxxxxxxxxxx.supabase.co/auth/v1/user 401
common.ts:91 AuthApiError: This endpoint requires a Bearer token
    at fetch.ts:36:16
```
Anyway, even after the error message, the session is successfully closed. This was working fine on the 1.x branch. What am I doing wrong?

Thanks

alaister:
Hi @howesteve,
Can you confirm this is still an issue on [2.0.0-rc.8](https://www.npmjs.com/package/@supabase/supabase-js/v/2.0.0-rc.8)?

howesteve:
Yes, it's still the same with [2.0.0-rc.8](https://www.npmjs.com/package/@supabase/supabase-js/v/2.0.0-rc.8).

alaister:
Hey @fernandops26,

Are you calling `getUser()` when the user is not logged in? That would be the correct error message in that case.

`getUser()` is not a direct replacement for `user()` in V1, as it makes an HTTP request every time it's called for the user. Here's how I'd get the user in V2:
```typescript
const {
  data: { session },
  error,
} = await supabase.auth.getSession()

if (error) {
  // handle error
}

const user = session?.user
```

@howesteve you wouldn't happen to be calling `getUser()` somewhere in your code as well?

howesteve:
> @howesteve you wouldn't happen to be calling `getUser()` somewhere in your code as well?
Yes, I was actually calling `getUser()` inside a ```onAuthStateChange()``` handler. Now that I switched into a ```getSession()``` -only approach, it looks fixed. However, that was actually quite confusing, since it's supposed to be regular api.

Thanks

mandarini:
I will be closing this issue since it's 3 years old, and also it seems to have a solution. If you think that this issue still persists, let me know, and please open a new issue with minimal reproduction, to help me track it better! Thank you all for contributing!! :D

PR #2453, opened 2026-06-16
fix(realtime): auto-enable presence on track() to repair sender-only channels

### LINKED ISSUE #1536: Presence not working as of supabase-js 2.54
# Bug report

## Describe the bug

In any client pages using version 2.54 or 2.55 of supabase-js, presence simply doesn't work.

## To Reproduce

1. Create two standalone HTML pages page1.html and page2.html.
2. Add a script to both which imports supabase-js@2.54 
3. In page1.html add [tracking state example from the docs](https://supabase.com/docs/guides/realtime/presence?queryGroups=language&language=js#sync-and-track-state).
4. In page2.html add [sending state example from the docs](https://supabase.com/docs/guides/realtime/presence?queryGroups=language&language=js#sending-state)
5. Open both pages. Observe console logs of presence successfully initing **but no logs of the sending state being received by the tracking page**

Repeat the above steps but change import to version 2.53 or earlier and observe presence being sent as expected.

## Expected behavior

Presence of the sending page is received by the tracking page.

## Screenshots

Full code snippet of reproduction pages. Replace xxx and yyy with your own public ID and token.

page1.html
```html
<html>
<head>
</head>
<body>
<script type="module">
    import { createClient } from 'https://cdn.jsdelivr.net/npm/@supabase/supabase-js@2.54/+esm'

    const supabase = createClient('https://xxx.supabase.co', 'yyy', {
    });

    const roomOne = supabase.channel('room_01');

    console.log("room1 open")

    roomOne
        .on('presence', { event: 'sync' }, () => {
            const newState = roomOne.presenceState()
            console.log('sync', newState)
        })
        .on('presence', { event: 'join' }, ({ key, newPresences }) => {
            console.log('join', key, newPresences)
        })
        .on('presence', { event: 'leave' }, ({ key, leftPresences }) => {
            console.log('leave', key, leftPresences)
        })
        .subscribe()
</script>
</body>
</html>
```

page2.html
```html
<html>
<head>
</head>
<body>
    <script type="module">
        import { createClient } from 'https://cdn.jsdelivr.net/npm/@supabase/supabase-js@2.54/+esm'

        const supabase = createClient('https://xxx.supabase.co', 'yyy', {
        });

        const roomOne = supabase.channel('room_01')

        const userStatus = {
            user: 'user-1',
            online_at: new Date().toISOString(),
        }

        roomOne.subscribe(async (status) => {
            if (status !== 'SUBSCRIBED') { return }

            const presenceTrackStatus = await roomOne.track(userStatus)
            console.log(presenceTrackStatus)
        })
    </script>
</body>
</html>
```

Now edit the version string to 2.53 or earlier and try again and observe it working.

## System information

- Browser: Tested in Chrome & Firefox
- Version of supabase-js: 2.54 & 2.55 (latest at time of submission)

mandarini:
## Description

Fixes a regression introduced in `supabase-js` 2.54 where presence stopped working for channels that only call `.track()` without registering a `.on('presence', ...)` listener.

### What changed?

In `RealtimeChannel.track()` (`packages/core/realtime-js/src/RealtimeChannel.ts`):

- If the channel is currently joined and `config.presence.enabled` is not `true`, flip the flag on the channel params and perform one `unsubscribe()` → `subscribe()` cycle before sending the track event. The second join goes out with `presence.enabled: true`, so the server registers the client as a presence participant and starts broadcasting its state.
- Existing channels that already opt in (either via a `.on('presence', ...)` listener, or via explicit `config: { presence: { enabled: true } }`) skip the new branch entirely.

Also:

- Tests: added `track() auto-enables presence when needed` describe block in `RealtimeChannel.presence.test.ts` covering the auto-resubscribe path, the no-op path for channels that opted in via config, and the no-op path for channels with a presence listener.
- TSDoc: documented the `presence.enabled` field on `RealtimeChannelOptions`, expanded the `track()` JSDoc to describe the implicit re-join behavior, and added a `@see {@link track}` cross-link on `untrack()`. These propagate to the generated reference docs.
- README: added a short paragraph in the Presence section explaining the opt-in for sender-only clients.

`realtime-js` 2.12 (shipped with `supabase-js` 2.54) added a `presence.enabled` flag on the channel join payload. The flag defaults to `false` and is auto-flipped to `true` only when the channel has a presence listener registered before `.subscribe()`. A "sender-only" client that only calls `.track()` therefore joined with `enabled: false`, so the server silently dropped its presence from broadcasts to other subscribers, with `track()` still returning `'ok'`.

The escape hatch added in `realtime-js` PR #530 (`config: { presence: { enabled: true } }`) works but is at the channel level. The official guidance to set `realtime: { presence: { enabled: true } }` on `createClient` does not propagate to channels and is a no-op, which is what the user in the issue confirmed. This PR makes the docs example and the most common user pattern work out of the box, while keeping the perf optimization for everyone who opts in explicitly.

Closes #1536

## Screenshots/Examples

Before this fix, the example from the official Presence guide would silently fail for the sender:

```js
const channel = supabase.channel('room_01')

channel.subscribe(async (status) => {
  if (status !== 'SUBSCRIBED') return
  await channel.track({ user: 'user-1', online_at: new Date().toISOString() })
})
```

Other clients listening on the same topic with `.on('presence', ...)` would never see this client. After this fix, the same code works without any changes. To skip the one-time internal resubscribe on the first `track()` call, callers can still opt in explicitly:

```js
const channel = supabase.channel('room_01', {
  config: { presence: { enabled: true } },
})
```

## Breaking changes

`.track()` keeps the same signature and return type. Channels that already opted into presence (listener or `enabled: true`) see no behavior change. Channels currently in the broken case go from silently dropping presence to actually broadcasting it. The only observable side effect is one extra leave/rejoin cycle on the first `track()` for those channels, once per channel lifetime.

## Checklist

## Additional notes

- The docs guide on supabase.com (the "sending state" snippet under https://supabase.com/docs/guides/realtime/presence) still omits the `enabled: true` opt-in. After this PR ships, the docs snippet works as-is, but it is worth updating in the docs repo to recommend the explicit opt-in for sender-only clients to avoid the resubscribe round-trip.
- `untrack()` does not need the same treatment: it is only meaningful after a successful `track()`, by which point presence is already enabled on the channel.

### DIFF
--- a/packages/core/realtime-js/README.md
+++ b/packages/core/realtime-js/README.md
@@ -178,6 +178,14 @@ channel.subscribe(async (status) => {
 })
 ```
 
+Presence is opt-in on the join payload. The client enables it automatically when a `.on('presence', ...)` listener is registered before `.subscribe()`. For "sender-only" clients that call `.track()` without listening for presence events, pass `enabled: true` explicitly to avoid an internal resubscribe round-trip on the first `track()`:
+
+```js
+const channel = client.channel('presence-test', {
+  config: { presence: { enabled: true } },
+})
+```
+
 ## Postgres CDC
 
 Receive database changes on the client.
--- a/packages/core/realtime-js/package.json
+++ b/packages/core/realtime-js/package.json
@@ -42,7 +42,7 @@
     "test:ci": "vitest run --coverage"
   },
   "dependencies": {
-    "@supabase/phoenix": "^0.4.2",
+    "@supabase/phoenix": "^0.4.3",
     "tslib": "2.8.1"
   },
   "devDependencies": {
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -437,14 +437,32 @@ export default class RealtimeChannel {
     payload: { [key: string]: any },
     opts: { [key: string]: any } = {}
   ): Promise<RealtimeChannelSendResponse> {
-    if (this.state === CHANNEL_STATES.joined && this.params.config.presence?.enabled !== true) {
+    // Mirror the auto-detection in subscribe() so we only re-join when the
+    // server actually treats this channel as non-presence. A listener-only
+    // channel already joined with enabled: true and does not need re-joining.
+    const presenceAlreadyEnabled =
+      (!!this.bindings[REALTIME_LISTEN_TYPES.PRESENCE] &&
+        this.bindings[REALTIME_LISTEN_TYPES.PRESENCE].length > 0) ||
+      this.params.config.presence?.enabled === true
+
+    if (this.state === CHANNEL_STATES.joined && !presenceAlreadyEnabled) {
       this.params.config.presence = {
         ...this.params.config.presence,
         enabled: true,
       }
       this.socket.log('channel', `resubscribe to ${this.topic} to enable presence for track()`)
       await this.unsubscribe()
-      await this.subscribe()
+      await new Promise<void>((resolve, reject) => {
+        this.subscribe((status, err) => {
+          if (status === REALTIME_SUBSCRIBE_STATES.SUBSCRIBED) {
+            resolve()
+          } else if (status === REALTIME_SUBSCRIBE_STATES.CHANNEL_ERROR) {
+            reject(err ?? new Error(`channel error during track() re-subscribe`))
+          } else if (status === REALTIME_SUBSCRIBE_STATES.TIMED_OUT) {
+            reject(new Error(`timed out re-subscribing for track()`))
+          }
+        }, opts.timeout || this.timeout)
+      })
     }
     return await this.send(
       {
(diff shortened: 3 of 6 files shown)


mandarini:
Closing this in favor of two follow-ups.

The runtime fix was more invasive than expected — completing the "channel reusable after `unsubscribe`" semantics requires multiple coupled changes in `@supabase/phoenix` (the `joinedOnce` reset shipped in 0.4.3, plus a separate change so the channel isn't removed from `socket.channels` on close). On top of that, bumping to `@supabase/phoenix` 0.4.3 surfaced an unexplained regression in `integration.browser.test.ts` that I haven't been able to reproduce or diagnose remotely.

Splitting into:

1. Investigate the phoenix 0.4.3 regression in `integration.browser.test.ts`. Once we understand the cause, we can either patch phoenix again or work around it in `supabase-js`. This unblocks any future attempt at the runtime fix.

2. ship the documentation half of this PR as a small, non-breaking change. `config: { presence: { enabled: true } }` is the existing escape hatch; users just have no way to discover it today. Updating the TSDoc on `RealtimeChannelOptions.config.presence`, the JSDoc on `.track()` / `.untrack()`, and the README Presence section is a real improvement for anyone hitting #1536, independent of when (or whether) the runtime fix lands.

Phoenix 0.4.3 stays published.

PR #2464, opened 2026-06-20
fix(supabase): avoid dropping realtime auth to manual mode on client init

### LINKED ISSUE #1304: Realtime RLS does not work if I don't explicitly setAuth
# Bug report

## Describe the bug

I have valid session in `supabase.auth`, however if I subscribe to a channel, I'm not getting any data, because of RLS.
If I use `supabase.realtime.setAuth` with the access token, that has already been in the `auth.session`, the data starts coming.

## To Reproduce

1. create table `public.test` with column uid (so the RLS can be set so users can only read their data)
2. allow realtime
3. set auth policy so that users can only read their data
4. sign in user in your codebase, so that you have valid `supabase.auth.getSession`

```javascript
const channel = supabase.channel("premium").on(
        'postgres_changes',
        { event: 'UPDATE', schema: 'public', table: 'test'},
        (payload) => {
            console.log("Premium change payload", payload);
        }
    ).subscribe((status, err) => {
        if (err) {
            console.error("Error subscribing to premium channel", err);
        }
        console.log("Subscription status", status);
    })
```
this code won't reliably receive updates for the signed in user in the `public.test` table. However if we add something like this, the data starts coming.

```javascript
    const { data, error } = await supabase.auth.getSession();
    if (error || !data) {
        console.error("Error getting session", error);
        return;
    }
    supabase.realtime.setAuth(data.session.access_token);
```

## Expected behavior
I was expecting that the session is propagated automatically. I only found this [reference](https://supabase.com/docs/guides/realtime/postgres-changes?queryGroups=language&language=js&queryGroups=database-method&database-method=dashboard#refreshed-tokens) in documentation, but that's referring to custom tokens, but this happened for regular supabase sign in without no custom tokens. So either the auth should be propagated to realtime automatically or add mention do documentation to always use `setAuth` for realtime.

## System information

- OS: macOS
- Browser: electron
- Version of supabase-js: [2.45.2]
- Version of Node.js: [16]

hmnd:
## 🔍 Description

When using a custom callback for `accessToken` on `SupabaseClient`, realtime gets stuck on the access token initially set during init, rather than refreshing from the callback as expected.

### What changed?

I've replaced the previous approach of manually requesting the accesstoken inside `SupabaseClient`'s constructor with an empty call to `realtime.setAuth()`. This way realtime retrieves and sets the access token on its own, without dropping  to manual mode.

I ran into this issue with my custom minted jwts. I found #1304, in which the fix was to manually call `supabase.realtime.setAuth()` with no args to re-enable callback mode for access tokens.

I also couldn't find any documentation advising users of custom tokens that they must always call do this after client init, making for poor UX and confusing bugs.

Closes #1304

## 🔄 Breaking changes

Technically not a *breaking* change but, with this change, we do lose the `Failed to set initial Realtime auth token` error message if the `accessToken` function throws, because realtime's setAuth has its own error handling. 

## 📋 Checklist

## 📝 Additional notes

#1304 really shouldn't have been closed and that is what frustrates me about stale bots. I would kindly suggest reconsidering the repo's usage of stale bot, or altering its config to eg. re-open issues if commented on after auto-close.


mandarini (changes requested):
Hi @hmnd, thank you so much for contributing to Supabase!  This is a great catch, and the root cause diagnosis is spot on. Confirmed that `setAuth(token)` with an explicit value flips realtime into manual mode for good, which is exactly why refresh was getting stuck for custom JWT callbacks. Really appreciate you digging into this instead of just settling for the workaround from #1304.

Two small things before we merge:

1. Could you add a `.catch()` back on the `this.realtime.setAuth()` call in the constructor? Right now a failure there would be an unhandled promise rejection rather than a caught error.
2. A few months ago we specifically wrapped this same call in `Promise.resolve()` to fix a Firefox extension cross-context Promise error (#2033). Since this PR removes that wrap, could you double check (or let us know your reasoning on) whether that issue could resurface here?

Thank you again for taking the time to fix this properly instead of just leaving the workaround in place. Contributions like this are what keep our SDKs solid.

hmnd:
@mandarini thanks for the review!

Regarding your comments:
1. I've re-added the catch, though realistically, it likely won't ever catch anything since everything within `setAuth` is contained in a giant try/catch
2. The `Promise.resolve` wrapping is no longer needed because (as I understand it) the original issue stemmed from `this.accessToken()` being called in a different context from where it was defined. We no longer call `this.accessToken()` from within the constructor - instead only awaiting it within `realtime.setAuth`, where it has never caused such context-crossing issues. Not to mention, we're not doing a `.then` anymore either.

PR #2466, opened 2026-06-23
fix(realtime): forward track() timeout option to send

i-anubhav-anand:
### What changed?

`RealtimeChannel.track()` passed `opts.timeout || this.timeout` (a **number**) as the second argument to `send()`. But `send(args, opts)` treats its second argument as an **options object** and reads `opts.timeout` internally. A number's `.timeout` is `undefined`, so the caller's `track(payload, { timeout })` was silently dropped and the channel's default timeout was always used.

This PR makes `track()` forward the whole `opts` object — matching its sibling `untrack(opts)`, which already does exactly this (`send` then applies `opts.timeout || this.timeout` internally).

```diff
       },
-      opts.timeout || this.timeout
+      opts
     )
```

Custom per-`track` timeouts were silently ignored — `channel.track(payload, { timeout })` had no effect.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

**Fail-before / pass-after** — added `track() forwards a custom timeout option to the underlying push` in `RealtimeChannel.presence.test.ts`, which spies on the underlying `channelAdapter.push` call:

- **Before fix:** ❌ `expected push to be called with timeout 1234, received 1000` (the channel default)
- **After fix:** ✅ passes; full `realtime-js` suite green (355 passed, 1 skipped)

### DIFF
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -428,7 +428,7 @@ export default class RealtimeChannel {
         event: 'track',
         payload,
       },
-      opts.timeout || this.timeout
+      opts
     )
   }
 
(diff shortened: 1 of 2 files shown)


edgurgel:
Fixed by https://github.com/supabase/supabase-js/pull/2490

Sorry I didn't see this PR 🤦

PR #2467, opened 2026-06-23
fix(auth): accept uppercase UUIDs in validateUUID

i-anubhav-anand:
### What changed?

`validateUUID()` checked its input against `UUID_REGEX`, which lacked the case-insensitive (`i`) flag:

```diff
-const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
```

As a result, any UUID containing uppercase hex (e.g. `123E4567-E89B-12D3-A456-426614174000`) was rejected **client-side** with `Expected parameter to be UUID but is not`, before the request was ever sent — even though the value is a perfectly valid UUID and the GoTrue server accepts it.

This affects the public admin methods that validate a UUID argument: `getUserById`, `updateUserById`, `deleteUser`, `_listFactors`, `_deleteFactor`, `_adminListPasskeys`, `_adminDeletePasskey`.

UUIDs are case-insensitive on input. Per **RFC 9562 §4** (and RFC 4122 §3): *"The hexadecimal values 'a' through 'f' … are case insensitive on input."* The GoTrue server (Go `uuid` parser) accepts uppercase UUIDs, so rejecting them in the client is incorrect and blocks valid calls.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

**Fail-before / pass-after** — added uppercase and mixed-case cases to the `validateUUID` test in `helpers.test.ts`:

- **Before fix:** ❌ `should accept uppercase UUID` and `should accept mixed-case UUID` both throw `@supabase/auth-js: Expected parameter to be UUID but is not`
- **After fix:** ✅ all 26 `helpers.test.ts` tests pass

### DIFF
--- a/packages/core/auth-js/src/lib/helpers.ts
+++ b/packages/core/auth-js/src/lib/helpers.ts
@@ -369,7 +369,7 @@ export function getAlgorithm(
   }
 }
 
-const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
 
 export function validateUUID(str: string) {
   if (!UUID_REGEX.test(str)) {
--- a/packages/core/auth-js/test/helpers.test.ts
+++ b/packages/core/auth-js/test/helpers.test.ts
@@ -313,6 +313,16 @@ describe('validateUUID', () => {
       shouldThrow: false,
     },
     {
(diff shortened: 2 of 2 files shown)


mandarini (approved):
@i-anubhav-anand thanks for this fix. Can you please rebase your PR so that it will trigger the suite and we can merge?

i-anubhav-anand:
Thanks! I've pushed a fresh commit to retrigger CI (`23ccf690`) — content is unchanged, just rebased onto latest `master`. Looks like the workflow runs are queued but sitting in `action_required` status, so they'll need an approval click from a maintainer to actually execute: https://github.com/supabase/supabase-js/pull/2467/checks. Let me know if there's anything else needed on my end!

PR #2469, opened 2026-06-23
fix(functions): pass ReadableStream body through without JSON serialization

i-anubhav-anand:
### What changed?

`FunctionInvokeOptions.body` documents `ReadableStream<Uint8Array>` as a supported type, but the body-inference logic in `invoke()` had no branch for it. A `ReadableStream` matched none of the `Blob`/`ArrayBuffer`/`string`/`FormData` cases and fell through to `JSON.stringify(functionArgs)` — which produces `"{}"`, silently dropping the entire stream payload and mislabeling it `application/json`. This happened in **both** body-inference paths (with and without a caller-supplied `Content-Type`).

This PR:
1. Adds a `ReadableStream` branch in both paths that forwards the stream untouched (no forced `application/json`).
2. Sets `duplex: 'half'` on the fetch call when the body is a `ReadableStream`, which native (undici) fetch requires for a streaming request body — mirroring the existing storage-js behavior added in #2287.

A documented, supported body type was silently corrupted into `"{}"`. Per the WHATWG Fetch standard, `ReadableStream` is a valid `BodyInit`, and the library's own `types.ts` lists it as supported — so serializing it as JSON is a contract violation.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

**Fail-before / pass-after** — added `test/spec/readablestream-body.spec.ts` (Docker-free, custom-fetch capture) covering a stream body both with and without a caller `Content-Type`, asserting the stream is forwarded unchanged and `duplex: 'half'` is set:

- **Before fix (on `master`):** ❌ both cases — `expect(received).toBe(stream)` → `Received: "{}"`
- **After fix:** ✅ both pass (stream forwarded, `duplex: 'half'` set)

### DIFF
--- a/packages/core/functions-js/src/FunctionsClient.ts
+++ b/packages/core/functions-js/src/FunctionsClient.ts
@@ -244,6 +244,13 @@ export class FunctionsClient {
           // don't set content-type headers
           // Request will automatically add the right boundary value
           body = functionArgs
+        } else if (
+          typeof ReadableStream !== 'undefined' &&
+          functionArgs instanceof ReadableStream
+        ) {
+          // a stream is a valid fetch body; forward it untouched and let the caller set the
+          // Content-Type (don't serialize it to JSON, which would drop the payload)
+          body = functionArgs
         } else {
           // default, assume this is JSON
           _headers['Content-Type'] = 'application/json'
@@ -255,7 +262,8 @@ export class FunctionsClient {
           typeof functionArgs !== 'string' &&
           !(typeof Blob !== 'undefined' && functionArgs instanceof Blob) &&
           !(functionArgs instanceof ArrayBuffer) &&
-          !(typeof FormData !== 'undefined' && functionArgs instanceof FormData)
+          !(typeof FormData !== 'undefined' && functionArgs instanceof FormData) &&
+          !(typeof ReadableStream !== 'undefined' && functionArgs instanceof ReadableStream)
         ) {
           body = JSON.stringify(functionArgs)
         } else {
@@ -289,6 +297,10 @@ export class FunctionsClient {
         // 3. default Content-Type header
         headers: { ..._headers, ...this.headers, ...headers },
         body,
+        // a web ReadableStream body requires the half-duplex flag for native (undici) fetch
+        ...(typeof ReadableStream !== 'undefined' && body instanceof ReadableStream
+          ? { duplex: 'half' }
+          : {}),
         signal: effectiveSignal,
       }).catch((fetchError) => {
         throw new FunctionsFetchError(fetchError)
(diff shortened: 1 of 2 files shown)


packages/core/functions-js/src/FunctionsClient.ts
```diff
@@ -284,6 +292,10 @@ export class FunctionsClient {
         // 3. default Content-Type header
         headers: { ..._headers, ...this.headers, ...headers },
         body,
+        // a web ReadableStream body requires the half-duplex flag for native (undici) fetch
+        ...(typeof ReadableStream !== 'undefined' && body instanceof ReadableStream
+          ? { duplex: 'half' }
+          : {}),
```
kallebysantos: Instead of ternary matching here with the Readable guard again, could you move this into the Readable branch, defined at `ln 244`.

This can be done by following the same approach that `headers` does

kallebysantos (changes requested):

mandarini:
@i-anubhav-anand are you still working on this? Did you take a look at the feedback @kallebysantos posted?

packages/core/functions-js/src/FunctionsClient.ts
i-anubhav-anand: Done in 068ba054 — I now set a `duplex` flag alongside `body` in the ReadableStream branch (both the no-Content-Type path and the caller-Content-Type path) and just spread `...(duplex ? { duplex } : {})` at the fetch call, so there's no longer a second `instanceof ReadableStream` guard there. Mirrors how the Content-Type header is set per-branch and consumed once. Thanks for the review!

i-anubhav-anand:
@mandarini yes, still on it — sorry for the delay. I've just pushed 068ba054 addressing @kallebysantos's feedback: the `duplex: 'half'` flag is now set inside the ReadableStream branch instead of being re-derived via an `instanceof` check at the fetch call. Build passes and the fail-before/pass-after unit tests are green. Ready for another look.

i-anubhav-anand:
Update: rebased the branch onto the latest `master` (it was ~63 commits behind) and re-applied the fix on top — now `0` behind, no conflicts. The tip is `09519f48`.

Recap of what's in the PR:
- **Feedback from @kallebysantos addressed** — the `duplex: 'half'` flag is now set alongside `body` inside the ReadableStream branch (both the no-Content-Type and caller-Content-Type paths), and consumed once at the fetch call via `...(duplex ? { duplex } : {})`. No more second `instanceof ReadableStream` guard at the fetch site — mirrors how the Content-Type header is set per-branch and read once.
- Rebase preserved the upstream `onAbort` listener-cleanup change; no overlap with my hunks.

Validated locally: `tsc` (main + module) compiles clean, prettier passes, and the fail-before/pass-after unit test (`test/spec/readablestream-body.spec.ts`) is green (2/2). Ready for another look 🙏

PR #2474, opened 2026-06-27
fix(storage): resolve createSignedUrls return type mismatch

gourabsingha1:
This PR fixes a type signature mismatch in `createSignedUrls` (reported in supabase/storage-js#169). The API response actually contains both `signedURL` (raw path) and `signedUrl` (fully qualified URL), but only `signedUrl` was defined in the TypeScript type signature. This PR updates the signature to correctly expose both fields to TypeScript consumers.


mandarini (approved):
Thanks for this fix. In v3 I'll make sure to consolidate the two, but right now it would be a breaking change, so your solution is good.

can you please rebase to trigger CI so we can merge? @gourabsingha1

Issue #1730, opened 2024-03-18
Supabase Realtime React Native silently fails when loading session from storage and using RLS/JWT


kav:
# Bug report
- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Supabase Realtime React Native silently fails when loading session from storage and using RLS that relies on supabase JWT claims.
Subscribing to postgres changes reports ok be receives no updates if the session is loaded from storage. Logging out and back in resolves the issue.

SUBSCRIBED and CLOSED events occur as expected. Logging realtime events at `info` show the same events in either case with the exception of the expected.

Disabling RLS immediately fixes the issue and reenabling immediately re-triggers it. 
RLS policy is checking a property on the JWT so this appears to be the issue. 
Setting the Realtime auth token manually via
```js
supabase.auth.onAuthStateChange((event, session) => {
  if(session) {
    supabase.realtime.setAuth(session?.access_token);
  }
});
```
resolves the issue.

## Expected behavior

Sessions loaded from storage or newly logged in should be able to use properties of the supabase jwt without manually setting.

## Additional context

This is a regression as at some point previously this workaround was not required with the same RLS polices 

Auth options are
```js
{
   autoRefreshToken: true,
    persistSession: true,
    storage: AsyncStorage,
}
```

Versions:
```json
    "@react-native-async-storage/async-storage": "1.22.3",
    "@supabase/supabase-js": "2.39.8",
```

jonxssc:
@rogaha @JonathanLab do you guys have any update to this issue? I am struggeling with this error for a while. Sometimes on launch the Realtime Events jusg dont come in even tho the logs in the Supabase Dashboard say everyrhing is fine and the subscription is successfull. Supabase Support also did not find any solution for the problem

filipecabaco:
Would this issue be related? https://github.com/supabase/supabase-js/issues/1477

jonxssc:
> Would this issue be related? https://github.com/supabase/supabase-js/issues/1477 

Yes, this seems to be the same problem

kav:
This issue remains unresolved.

PR #2475, opened 2026-06-27
fix(supabase): set realtime auth token on INITIAL_SESSION event

gourabsingha1:
This PR fixes a bug (Issue #1730) where the Realtime client fails to set the auth token on initial session load / storage recovery. Currently, the Supabase client handles `TOKEN_REFRESHED` and `SIGNED_IN` events in `_handleTokenChanged`, but completely misses the `INITIAL_SESSION` event which is fired when the session is loaded from storage upon client initialization. This causes Realtime channels with RLS/JWT claims to silently fail. This PR adds `INITIAL_SESSION` to the list of handled events.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -645,7 +645,7 @@ export default class SupabaseClient<
     token?: string
   ) {
     if (
-      (event === 'TOKEN_REFRESHED' || event === 'SIGNED_IN') &&
+      (event === 'TOKEN_REFRESHED' || event === 'SIGNED_IN' || event === 'INITIAL_SESSION') &&
       this.changedAccessToken !== token
     ) {
       this.changedAccessToken = token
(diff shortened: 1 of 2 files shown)


mandarini:
Hi @gourabsingha1, thank you so much for contributing to Supabase! First off, I want to apologize for the timing here. Your diagnosis of this bug is spot on: `_handleTokenChanged` was indeed missing the `INITIAL_SESSION` case, so a session recovered from storage never got pushed to already-connected realtime channels, which is exactly the kind of silent RLS/JWT failure described in #1730. That is a real and easy-to-miss race condition, and you found it and fixed it correctly.

The unlucky part is that an equivalent fix landed on master in #2531 a little while after you opened this PR, adding the same `INITIAL_SESSION` check in the same spot. Since this branch was opened before that merged and hasn't been rebased since, the diff here still shows as needed even though master already has it.

Could you double check against current master and confirm the behavior you were fixing is resolved there? If so, I think we can close this one out with our thanks, since there is nothing left for it to add on top of #2531. And if you spot anything #2531 does not quite cover that your version handles differently, please do flag it, I would rather we catch a gap than assume it is fully covered.

Sorry again for the crossed wires, and thank you so much for contributing to Supabase. Catching subtle races like this is exactly the kind of contribution that makes our tools better.

gourabsingha1:
@mandarini Thanks for checking! I verified against latest master and can confirm that PR #2531 completely resolves the issue (including the unit test). Glad to see it's fixed! Closing this PR now.

mandarini:
Thank you very much @gourabsingha1

Issue #969, opened 2024-02-08
Selecting a schema on createClient does not work


odarriba:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I am moving all contents from a Supabase server to another. I have exported the roles, schema and data, and inserted it again in the new server.

After that, I had to move the storage blobs from the bucket of the previous server to the new one, as indicated on this:

https://supabase.com/docs/guides/platform/migrating-and-upgrading-projects#migrate-storage-objects

But when I do, it fails on the step of obtaining the objects list from the previous DB. Inspecting the error returned I can see it fails because of this:

```
{
  code: '42P01',
  details: null,
  hint: null,
  message: 'relation "public.objects" does not exist'
}
```

and taking into account the script is selecting the schema `storage` for that oldSupabaseRestClient variable initialization, it does not make any sense.

I tried changing the name of the schema in the initialization in case it is just protecting `storage` but the error remains the same: always trying to access `public`.

## To Reproduce

Have two instances and try to run the script on that docs page.

## Expected behavior

Be able to get the list of files and continue with the migration

## Screenshots

-

## System information

- OS: macOS
- Browser (if applies): Does not apply
- Version of supabase-js: 2.39.3
- Version of Node.js: v18.16.1

## Additional context

-

belle-chang:
I have the same issue - please assist!

belle-chang:
Figured out the problem, when accessing the schema, it needs to be:

```
const oldSupabaseRestClient = createClient(OLD_PROJECT_URL, OLD_PROJECT_SERVICE_KEY, {
    schema: 'storage',
})
```

@supabase team, can you update the docs on this?

CooperHash:
same problem

SqlOwl:
same problem with TS project

GorkaMM:
Same issue on my side. Typescript is not recognizing any other schema than `public` for the db.schema option in `createClient`, even when the supabase.types.ts file is correctly including it.

<img width="952" alt="Image" src="https://github.com/user-attachments/assets/869a6f34-26fe-427e-8bd9-ceceaaf949d4" />

It does work when adding it to the query like `supabase.schema('api').from('todos').select('*')`, so it seems to be an issue with the options for the client creation.

[...]

PR #2476, opened 2026-06-27
fix(supabase): relax db schema type constraints in createClient options

gourabsingha1:
This PR fixes a bug (Issue #969) where TypeScript displays a type error when attempting to initialize a client with a custom database schema under `db.schema`, unless the user explicitly passes the schema as the second generic parameter to `createClient`. Because of the lack of partial generic inference in TypeScript, when only the first generic parameter `<Database>` is supplied, the second generic parameter (`SchemaName`) defaults to `'public'`, restricting the type of the `db.schema` option to `'public'`. This PR relaxes the `schema` option type under `db` to `string` so that initialization compiles successfully, while preserving type-safety of database/schema queries.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -394,7 +394,7 @@ export default class SupabaseClient<
 
     this.rest = new PostgrestClient(new URL('rest/v1', baseUrl).href, {
       headers: this.headers,
-      schema: settings.db.schema,
+      schema: settings.db.schema as any,
       fetch: this.fetch,
       timeout: settings.db.timeout,
       urlLengthLimit: settings.db.urlLengthLimit,
--- a/packages/core/supabase-js/src/lib/types.ts
+++ b/packages/core/supabase-js/src/lib/types.ts
@@ -98,7 +98,7 @@ export type SupabaseClientOptions<SchemaName> = {
    * The Postgres schema which your tables belong to. Must be on the list of exposed schemas in Supabase. Defaults to `public`.
    */
   db?: {
-    schema?: SchemaName
+    schema?: string
     /**
      * Optional timeout in milliseconds for PostgREST requests.
      * When set, requests will automatically abort after this duration to prevent indefinite hangs.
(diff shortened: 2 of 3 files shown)


mandarini (changes requested):
Hi @gourabsingha1, thank you so much for contributing to Supabase! Really appreciate you digging into #969, it's a genuinely confusing error to hit.

One thing to flag before this can merge: relaxing `db.schema` to `string` removes the type checking for everyone, including the two-generic form (`createClient<Database, 'personal'>(...)`) where it currently works correctly and catches typos. It also doesn't fully solve the original problem, since with only `createClient<Database>(...)`, the client's `.from()` calls will still be typed against `public` even though `db.schema` is set to `personal` at runtime. So the fix trades a compile error for a silent mismatch between the configured schema and the query types.

The good news is there's already a fully type-safe way to do this: `supabase.schema('personal').from(...)`, which infers its type directly from the call. Would you be up for either narrowing this to only relax the type where inference genuinely can't help, or pivoting the PR toward pointing users at `.schema()` (docs and maybe a better error message) instead? Happy to help think through either direction.

Thank you again for taking the time to contribute, contributions like this are exactly how we find the rough edges worth smoothing out.

gourabsingha1:
@mandarini Thanks for the feedback! I updated the type signature to use `SchemaName | (string & {})`. This preserves strict type checking and IDE autocompletion for `SchemaName` when specified, while relaxing the type to allow custom string schema names when only `<Database>` is supplied. Rebased onto `master`!

mandarini:
Hi @gourabsingha1, sorry for taking SOOOO long to reply... I had to rethink this one. I decided I am going to close this, but not because the problem isn't real. It is, and you digging into #969 is what got it moving.

The reason is that relaxing `db.schema` to accept any string removes a compile error that is actually correct: with only `createClient<Database>(...)`, TypeScript never infers the schema from db.schema, so .from() stays typed against public no matter what you configure. Widening the type hides that rather than fixing it, and the resulting mismatch is silent at runtime. That inference gap is a TypeScript limitation, not something we can close in the library.

The fix that actually helps people is pointing them at the two forms that do work, `createClient<Database, 'myschema'>(...)` and `supabase.schema('myschema').from(...)`. That's now in #2662 (JSDoc) and supabase/supabase#49967 (docs guide).

Thanks again for raising this, it made the gap visible. Hope to see you around the repo again.

PR #2477, opened 2026-06-27
fix(auth): warn at registration when async callback passed to onAuthStateChange

### LINKED ISSUE #1401: Supabase JS Client Hangs During onAuthStateChange Fetch After Refresh (Chrome/Expo Web)
**Description:**

The Supabase JS client appears to become unresponsive to network requests (`supabase.from(...).select()`) under specific conditions when running an Expo Web application in Google Chrome (tested on Ubuntu). The issue manifests after a user logs in (e.g., via email/password), successfully persists the session (`persistSession: true`), and then refreshes the browser page *more than once*.

- The initial page load after login works correctly.
- However, on the *reload* (and subsequent) browser refresh, `onAuthStateChange` fires correctly, indicating the persisted session was read, but when code within the listener attempts an async operation involving a Supabase network request (e.g., `await fetchUserProfile()` which calls `supabase.from('profiles').select()`), the client hangs indefinitely. The network request is never initiated according to browser dev tools, and the `await` never resolves or rejects.
- This behavior **does not** occur in Firefox on the same OS (Ubuntu) – subsequent refreshes work as expected.
- A related issue was observed where calling `supabase.auth.signOut()` (default scope) also caused the client to hang similarly in Chrome/Expo Web. Using `signOut({ scope: 'local' })` resolved the manual sign-out hang but did *not* resolve the refresh hang.
- Disabling session persistence (`persistSession: false`) prevents the refresh hang (as expected, since no session is read), confirming the issue is related to initializing from the persisted state on subsequent loads.

**Environment:**

-   **`@supabase/supabase-js` Version:** 2.49.4
-   **Expo SDK Version:** ~52.0.46
-   **React Version:** ~18.3.1
-   **React Native Version:** ~0.76.9
-   **Platform:** Expo Web (`npx expo start --web`)
-   **Operating System:** Ubuntu
-   **Browser:** Google Chrome - Version 135.0.7049.114 (Official Build) (64-bit)

**Steps to Reproduce:**

1.  Create a basic Expo application (`npx create-expo-app my-app`).
2.  Install `@supabase/supabase-js` and configure the client in `lib/supabaseClient.ts` with `persistSession: true` and default web storage (no custom adapter needed for web reproduction).

    ```typescript
    // Example supabaseClient.ts config
    import { createClient, SupabaseClientOptions } from '@supabase/supabase-js';

    const supabaseUrl = process.env.EXPO_PUBLIC_SUPABASE_URL;
    const supabaseAnonKey = process.env.EXPO_PUBLIC_SUPABASE_ANON_KEY;

    let storageOptions: SupabaseClientOptions<'public'>['auth'] = {
      autoRefreshToken: true,
      persistSession: true, // CRITICAL for bug reproduction
      detectSessionInUrl: false,
      // storage: undefined (defaults to localStorage on web)
    };

    export const supabase = createClient(supabaseUrl!, supabaseAnonKey!, { auth: storageOptions });
    ```

3.  Implement a simple `AuthContext` that listens to `onAuthStateChange`. Inside the listener, upon detecting a session (`INITIAL_SESSION` with session, `SIGNED_IN`, `TOKEN_REFRESHED`), immediately attempt to fetch user data using the client.

    ```typescript
    // Simplified AuthContext demonstrating the pattern
    useEffect(() => {
      const { data: { subscription } } = supabase.auth.onAuthStateChange(async (event, session) => {
        console.log(`Event: ${event}`, session);
        if (session) {
          setIsLoading(true); // Assume setIsLoading state setter
          console.log(`Session detected (${event}), attempting profile fetch for ${session.user.id}`);
          try {
            // PROBLEM AREA: This await hangs on second refresh in Chrome
            const { data: profile, error } = await supabase
              .from('profiles') // Or any other table read
              .select('*')
              .eq('id', session.user.id)
              .single();

            console.log('Profile fetch completed.', { profile, error }); // This log is never reached on second refresh hang
            if (error && error.code !== 'PGRST116') throw error;
            // ... update state with profile ...
          } catch (error) {
            console.error("Profile fetch failed:", error);
            // ... handle error / sign out ...
          } finally {
            setIsLoading(false);
          }
        } else {
          // ... handle sign out / no session ...
          setIsLoading(false);
        }
      });
      return () => subscription.unsubscribe();
    }, [supabase]); // Dependencies simplified for example
    ```

4.  Implement a basic login form using `supabase.auth.signInWithPassword`.
5.  Run the app on **Chrome** (`npx expo start --web`).
6.  Log in successfully. Observe the profile fetch completes.
7.  **Refresh the page.**

**Expected Behavior:**

App re-authenticates, profile fetch completes, UI loads correctly based on the authenticated state.

**Actual Behavior:**

The `onAuthStateChange` listener logs indicate the session was detected, the `console.log` *before* the `await supabase.from('profiles')...` runs, but the `await` never resolves or rejects, and the subsequent `console.log` and state updates do not occur. The app remains stuck in a loading state, and no network request for the profile fetch appears in DevTools.

**Workaround Found:**

The hang can be avoided by decoupling the profile fetch from the `onAuthStateChange` listener. Instead of fetching directly in the listener:
1.  The listener only updates the basic `session` and `supabaseUser` state.
2.  A separate `useEffect` hook depends on the `supabaseUser` state.
3.  When `supabaseUser` changes, *this second effect* initiates the profile fetch (`await supabase.from('profiles')...`).

This suggests a potential timing or internal state issue within the Supabase client when handling session initialization from storage immediately followed by a network request within the same event callback in the Chrome/Expo Web environment.

gourabsingha1:
## Problem

Using `async` callbacks with `onAuthStateChange` is a very common footgun that has burned hundreds of developers (44 comments over 2 years on https://github.com/supabase/auth-js/issues/762 alone).

The deadlock pattern:
1. Auth method (e.g. `signInWithPassword`) starts notifying subscribers via `_notifyAllSubscribers`
2. Subscriber's async callback calls `refreshSession` or anything routing through `_callRefreshToken`
3. The refresh cannot proceed because the notification is still being awaited
4. The operation hangs indefinitely — no error, no timeout

The async overload is already marked `@deprecated` in the JSDoc, but developers rarely read overload-level docs at call sites.

## Solution

Add a `console.warn` the moment an **async** callback is registered with `onAuthStateChange`. The warning:
- fires **once**, at registration time (not on every event)
- points to the tracking issue for more context
- has **zero impact** on sync callbacks or any existing runtime behavior

Detection uses `callback.constructor.name === 'AsyncFunction'` — a standard, portable JavaScript introspection that works in all environments.

## Changes

- `packages/core/auth-js/src/GoTrueClient.ts`: add async-callback detection + `console.warn` in `onAuthStateChange`
- `packages/core/auth-js/test/onAuthStateChange.test.ts` \[NEW\]: dedicated unit tests that run without a live Supabase instance

## Tests

The three new unit tests verify:
1. An async callback triggers exactly one `console.warn`
2. A sync callback triggers no `console.warn`
3. The warning fires only once at registration, not per-event

```
PASS test/onAuthStateChange.test.ts
  onAuthStateChange – async callback deprecation warning
    ✓ emits console.warn when an async function is registered (X ms)
    ✓ does NOT emit console.warn when a sync function is registered (X ms)
    ✓ emits warning only once per async registration, not per event (X ms)
Tests: 3 passed, 3 total
```

## Related

- Closes https://github.com/supabase/supabase-js/issues/1401
- Related https://github.com/supabase/auth-js/issues/762 (2-year-old, 44-comment issue)
- Related https://github.com/supabase/auth-js/pull/1062 (predecessor PR in the old repo, closed when repo was deprecated)
- Related discussion in https://github.com/supabase/supabase-js/pull/2016 — maintainer @mandarini explicitly suggested: *"Detect if callback is async and warn/error if it awaits other auth methods"*

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -4204,6 +4204,18 @@ export default class GoTrueClient {
     data: { subscription: Subscription }
   } {
     const id: string | symbol = generateCallbackId()
+
+    if (callback.constructor.name === 'AsyncFunction') {
+      console.warn(
+        'onAuthStateChange: async callbacks are deprecated and can cause deadlocks. ' +
+          'Calling auth methods that trigger a token refresh (e.g. refreshSession) from inside a ' +
+          'TOKEN_REFRESHED handler will hang indefinitely because the refresh cannot complete ' +
+          'while its notification is still awaited. Move async work outside the callback using ' +
+          'setTimeout(() => { ... }, 0) or a non-async wrapper. ' +
+          'See https://github.com/supabase/supabase-js/issues/1401 for details.'
+      )
+    }
+
     const subscription: Subscription = {
       id,
       callback,
--- a/packages/core/auth-js/test/onAuthStateChange.test.ts
+++ b/packages/core/auth-js/test/onAuthStateChange.test.ts
@@ -1,0 +1,66 @@
+/**
+ * Unit tests for the async-callback deprecation warning in onAuthStateChange.
+ *
+ * These tests do NOT require a running Supabase instance — they only exercise
+ * the synchronous registration path of onAuthStateChange.
+ */
+
+import GoTrueClient from '../src/GoTrueClient'
+import { memoryLocalStorageAdapter } from '../src/lib/local-storage'
+
+function makeClient() {
+  return new GoTrueClient({
+    url: 'http://localhost:9999',
+    autoRefreshToken: false,
+    persistSession: false,
+    storage: memoryLocalStorageAdapter(),
+  })
+}
+
+describe('onAuthStateChange – async callback deprecation warning', () => {
+  test('emits console.warn when an async function is registered', () => {
+    const client = makeClient()
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+
+    const {
+      data: { subscription },
+    } = client.onAuthStateChange(async (_event, _session) => {})
+
+    subscription.unsubscribe()
+
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    expect(warnSpy.mock.calls[0][0]).toMatch(/async callbacks are deprecated/)
+    warnSpy.mockRestore()
+  })
+
+  test('does NOT emit console.warn when a sync function is registered', () => {
+    const client = makeClient()
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+
+    const {
+      data: { subscription },
+    } = client.onAuthStateChange((_event, _session) => {})
+
+    subscription.unsubscribe()
+
+    expect(warnSpy).not.toHaveBeenCalled()
+    warnSpy.mockRestore()
+  })
+
+  test('emits warning only once per async registration, not per event', async () => {
+    const client = makeClient()
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+
+    const {
+      data: { subscription },
+    } = client.onAuthStateChange(async (_event, _session) => {})
+
+    // Give the async INITIAL_SESSION emit a chance to fire
+    await new Promise((resolve) => setTimeout(resolve, 20))
+    subscription.unsubscribe()
+
+    // The warning must fire exactly once — at registration time
+    expect(warnSpy).toHaveBeenCalledTimes(1)
+    warnSpy.mockRestore()
+  })
+})


mandarini:
Hi @gourabsingha1, thank you so much for contributing to Supabase! I appreciate the research here, the async callback and deadlock history is long and you clearly worked through it. Unfortunately the ground shifted under this PR: since #2392 and #2498 landed, the general deadlock is fixed, and async callbacks that call `getUser`, `setSession`, etc. are safe (the `onAuthStateChange` JSDoc now says so explicitly). A registration-time warning on every async callback would mostly fire on code that is perfectly fine.

There is also a hard blocker: `@supabase/ssr` itself registers an async callback in `createServerClient`, so this warning would land in server logs on every single request for every SSR app.

The one hazard that does remain (awaiting `refreshSession` from inside a `TOKEN_REFRESHED` handler) is already documented on the deprecated overload. If we add runtime detection for it, it will need to live at the actual reentry point inside `_callRefreshToken`, where it can fail fast with no false positives, rather than at registration time.

So I am going to close this one. Thank you again for caring about this rough edge, it is contributions like yours that keep these libraries improving.

PR #2478, opened 2026-06-28
fix(postgrest): fall back to RFC 9110 reason phrase when statusText is empty

### LINKED ISSUE #405: statusText returns as empty string

isaka1022:
### What changed?

`PostgrestBuilder` builds its `PostgrestResponse.statusText` directly from `res.statusText`. Over HTTP/2 the
response has no reason phrase at all (HTTP/2 removed the status line — see RFC 7540 §8.1.2.4 / RFC 9113), so
`fetch` exposes an **empty** `statusText`. The same happens with some runtimes/proxies even on HTTP/1.1. As a
result, error objects returned from postgrest-js carry an empty `statusText`, which makes errors harder to read
and breaks code that switches on the reason phrase.

This PR adds a fallback: when `res.statusText` is empty, it derives the canonical RFC 9110 reason phrase from the
numeric status code. The server-provided value is always preferred when present, so HTTP/1.1 behavior is unchanged.

```ts
// before
let statusText = res.statusText
// after
let statusText = res.statusText || HTTP_STATUS_TEXTS[res.status] || ''
```

`HTTP_STATUS_TEXTS` is a small module-level map of the standard status codes to their RFC 9110 reason phrases.

Under HTTP/2, `error.statusText` (and the `statusText` on a successful `PostgrestResponse`) comes back as `""`,
which is confusing when inspecting/logging errors and breaks any consumer that relies on the reason phrase. The
status code is still available, so we can reconstruct the standard phrase deterministically.

Closes supabase/postgrest-js#405

## 📸 Screenshots/Examples

```ts
// HTTP/2 response, status 400, no reason phrase from the server
// before: error.statusText === ""
// after:  error.statusText === "Bad Request"
```

## 🔄 Breaking changes

The server-provided `statusText` is always used when present, so HTTP/1.1 responses are unaffected. The fallback
only fills in a value where there previously was an empty string.

## 📋 Checklist

- [ ] I have updated documentation (if applicable) — n/a

## 📝 Additional notes

Added 4 unit tests in `packages/core/postgrest-js/test/basic.test.ts` covering: `201 → "Created"`,
`206 → "Partial Content"`, `400 → "Bad Request"`, and that a server-provided phrase is never overwritten.
These run as plain unit tests (fetch is mocked with `statusText: ''`) and do not require the Docker
integration stack. Verified locally: the new tests fail against the current code (`Received: ""`) and pass
with the fix.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestBuilder.ts
@@ -16,6 +16,58 @@ import {
 import PostgrestError from './PostgrestError'
 import { ContainsNull } from './select-query-parser/types'
 
+// HTTP/2 omits reason phrases from responses (RFC 7540 §8.1.2.4).
+// This map provides fallback text so callers always receive a non-empty statusText.
+const HTTP_STATUS_TEXTS: Record<number, string> = {
+  100: 'Continue',
+  101: 'Switching Protocols',
+  200: 'OK',
+  201: 'Created',
+  202: 'Accepted',
+  203: 'Non-Authoritative Information',
+  204: 'No Content',
+  205: 'Reset Content',
+  206: 'Partial Content',
+  300: 'Multiple Choices',
+  301: 'Moved Permanently',
+  302: 'Found',
+  303: 'See Other',
+  304: 'Not Modified',
+  307: 'Temporary Redirect',
+  308: 'Permanent Redirect',
+  400: 'Bad Request',
+  401: 'Unauthorized',
+  402: 'Payment Required',
+  403: 'Forbidden',
+  404: 'Not Found',
+  405: 'Method Not Allowed',
+  406: 'Not Acceptable',
+  407: 'Proxy Authentication Required',
+  408: 'Request Timeout',
+  409: 'Conflict',
+  410: 'Gone',
+  411: 'Length Required',
+  412: 'Precondition Failed',
+  413: 'Content Too Large',
+  414: 'URI Too Long',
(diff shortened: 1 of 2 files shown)


mandarini:
Hi @isaka1022, thank you so much for contributing to Supabase! 💚 

This is a genuinely well-executed PR. You diagnosed the HTTP/2 behavior correctly, the server-provided phrase is always preserved when present, and the tests fail before the fix and pass after it. I really appreciate the care that went into this. However, I have decided not to adopt the fallback, and I want to explain the reasoning properly rather than just close.

The status code is the actual contract: it always arrives on every protocol version, and `statusText` carries no information beyond it. Under HTTP/2 there is simply no reason phrase on the wire, so an empty `statusText` is the truthful representation of what the server sent. Reconstructing the phrase client side means shipping and maintaining a static table that invites a long tail of follow-up issues (why is 418 missing, why does the RFC 9110 wording differ from what HTTP/1.1 servers send for codes like 413 and 422), and it nudges users toward branching on `statusText`, which is exactly the pattern that breaks across protocols.

The real bug here is on our side: the JSDoc examples imply `statusText` is always populated, and that mismatch is what misled the reporter of supabase/postgrest-js#405 in the first place. I am going to fix the documentation to make clear that `statusText` may be empty and that `status` is what consumers should rely on, and we will resolve #405 that way.

So I am closing this one, but please do not read that as a knock on the work. It is a quality contribution with solid tests, and we would love to see more from you. Thank you again for taking the time to improve Supabase, contributions like yours are what make these tools better for everyone.

PR #2486, opened 2026-07-02
fix(postgrest): quote array elements with reserved chars in contains/containedBy/overlaps

spokodev:
The array branches of `.contains()`, `.containedBy()` and `.overlaps()` build the
Postgres array literal with a raw `value.join(',')`, so an array element that
contains a comma is split into several elements and the server evaluates a
different filter than the one requested.

For example `.contains('tags', ['Doe, John', 'admin'])` emitted
`tags=cs.{Doe, John,admin}`, which Postgres reads as three elements
(`Doe`, ` John`, `admin`) instead of the intended two. The correct literal is
`tags=cs.{"Doe, John",admin}`. The same defect applied to `containedBy` and
`overlaps`.

A Postgres array literal requires elements that contain the delimiter, braces,
quotes, backslash or surrounding whitespace to be double quoted. The `.in()` and
`.notIn()` filters already quote reserved characters, and the realtime filter
builder does full quoting, so these three were the outliers.

The fix adds a small helper that double quotes and escapes only string elements
with reserved characters or surrounding whitespace. Numbers, booleans, plain
strings, and the range and JSON branches are unchanged. Added a test asserting
the emitted request URL.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -1016,7 +1016,7 @@ export default class PostgrestFilterBuilder<
       this.url.searchParams.append(column, `cs.${value}`)
     } else if (Array.isArray(value)) {
       // array
-      this.url.searchParams.append(column, `cs.{${value.join(',')}}`)
+      this.url.searchParams.append(column, `cs.{${value.map(toPostgrestArrayLiteral).join(',')}}`)
     } else {
       // json
       this.url.searchParams.append(column, `cs.${JSON.stringify(value)}`)
@@ -1165,7 +1165,7 @@ export default class PostgrestFilterBuilder<
       this.url.searchParams.append(column, `cd.${value}`)
     } else if (Array.isArray(value)) {
       // array
-      this.url.searchParams.append(column, `cd.{${value.join(',')}}`)
+      this.url.searchParams.append(column, `cd.{${value.map(toPostgrestArrayLiteral).join(',')}}`)
     } else {
       // json
       this.url.searchParams.append(column, `cd.${JSON.stringify(value)}`)
@@ -1592,7 +1592,7 @@ export default class PostgrestFilterBuilder<
       this.url.searchParams.append(column, `ov.${value}`)
     } else {
       // array
-      this.url.searchParams.append(column, `ov.{${value.join(',')}}`)
+      this.url.searchParams.append(column, `ov.{${value.map(toPostgrestArrayLiteral).join(',')}}`)
     }
     return this
   }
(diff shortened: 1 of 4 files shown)


mandarini (changes requested):
Hi @spokodev, thank you so much for contributing to Supabase!

Good catch, this is a real bug. A couple of changes before merging: the regex still misses a couple of cases Postgres treats specially, an empty string element or the bare word `null` both get emitted unquoted and read back as SQL `NULL` instead of the intended value. Rather than expanding the regex, could you just always double quote string elements? That's what PostgREST's own `pgBuildArrayLiteral` does and it's simpler than what we have now.

Also, `likeAllOf`, `likeAnyOf`, `ilikeAllOf`, and `ilikeAnyOf` build array literals the same unquoted way, so the same fix should apply there too.

Thanks again for digging into this, looking forward to the update!

spokodev:
Thanks! Switched `toPostgrestArrayLiteral` to always double-quote string elements, mirroring `pgBuildArrayLiteral` — an empty string and the literal word `null` now round-trip instead of collapsing to SQL `NULL`. Applied the same helper to `likeAllOf`/`likeAnyOf`/`ilikeAllOf`/`ilikeAnyOf`, and added the empty-string/`null` cases to the tests.

spokodev:
Checking in on this one. The update you asked for went in on 30 July: `toPostgrestArrayLiteral` now always double-quotes string elements the way `pgBuildArrayLiteral` does, so an empty string and the literal word `null` round-trip instead of collapsing to SQL `NULL`, and the same helper is applied to `likeAllOf`, `likeAnyOf`, `ilikeAllOf` and `ilikeAnyOf`. Tests cover both cases.

Anything else you would like changed before this can go in?

mandarini (changes requested):
Hi @spokodev, sorry for taking so long to answer. Thanks for making the update.

One thing before this can merge: the new tests in `array-literal.test.ts` only check the constructed URL string, they don't run against a real PostgREST server. Every neighboring test for these same methods in `test/filters.test.ts` (e.g. `contains with array`, `overlaps with array`) is an integration test that hits the local Dockerized PostgREST instance and snapshots the actual response. Could you add a case there too, with a comma-containing array element, so we have proof the server parses the fixed literal correctly, not just that our own code produces the string we expect?

Also, can you please rebase with `master` just in case?

Thank you so much for sticking with this!

spokodev:
Done — rebased on master and added the integration case.

`test/filters.test.ts` now has `contains with an array element containing a comma`, sitting next to the existing `contains with array` and following the same shape, and `test/supabase/seed.sql` gains one `cornercase` row whose element carries a comma:

```sql
(4, ARRAY['Doe, John', 'admin']);
```

The three existing `cornercase` tests all filter on `['test']`, and the `basic.test.ts` case asserts a duplicate key on `id: 1`, so the new row does not disturb any existing snapshot.

You were right that the URL assertion is only half the chain. I could not bring the Dockerized stack up locally (the daemon on this machine is wedged), so I verified the server half directly against PostgreSQL 14 with the same table and seed. The two literals are read differently:

| literal | parsed as | elements |
|---|---|--:|
| `{Doe, John,admin}` (before the fix) | `{Doe,John,admin}` | 3 |
| `{"Doe, John",admin}` (after) | `{"Doe, John",admin}` | 2 |

and on the actual operator:

```
select id from cornercase where array_column @> '{"Doe, John"}';  -- 4
select id from cornercase where array_column @> '{Doe, John}';    -- 0 rows
```

So the unquoted form did not error, it silently matched a different set — which is what makes the integration case worth having. The client side of the same chain is already pinned in `array-literal.test.ts` (`cs.{"Doe, John","admin"}`), and this new test closes the loop through PostgREST in CI.

I can add the same case for `containedBy` and `overlaps` if you would rather have all three, though they share `toPostgrestArrayLiteral`, so one server-side case covers the literal construction.

PR #2487, opened 2026-07-02
fix(functions): clean up cross-signal abort listener on invoke() return

FrancoKaddour:
When invoke() is called with both timeout and signal, an event listener is registered on the caller's AbortSignal to forward aborts to the internal timeoutController. That listener was never removed, so callers that reuse the same AbortController across multiple invocations accumulate dangling listeners — keeping the timeoutController alive indefinitely.

Root cause: the listener was created as an inline arrow function () => timeoutController!.abort()), so no reference existed to pass to removeEventListener.

Fix: store the listener in onAbort before registering it, and call options.signal?.removeEventListener('abort', onAbort) in the finally block so it's cleaned up on every code path (success, error, and abort).

Two unit tests are included that assert the same function reference is passed to both addEventListener and removeEventListener.

### DIFF
--- a/packages/core/functions-js/src/FunctionsClient.ts
+++ b/packages/core/functions-js/src/FunctionsClient.ts
@@ -205,6 +205,7 @@ export class FunctionsClient {
   ): Promise<FunctionsResponse<T>> {
     let timeoutId: ReturnType<typeof setTimeout> | undefined
     let timeoutController: AbortController | undefined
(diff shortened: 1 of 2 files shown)


kallebysantos (approved):

FrancoKaddour:
Thanks for the review and merge, @kallebysantos! 🙌

PR #2490, opened 2026-07-03
fix(realtime): forward opts to send() in track()

FrancoKaddour:
## What

RealtimeChannel.track() accepts an opts parameter (e.g. { timeout: 5000 }), but was passing opts.timeout || this.timeout — a plain number — as the second argument to send(). Since send() expects opts: { [key: string]: any }, accessing .timeout on a number always returns undefined, so the actual timeout from the caller is silently ignored.

untrack() already passes opts directly and works correctly. This aligns track() with the same behavior.

## Before

```ts
async track(payload, opts = {}) {
  return await this.send({ type: 'presence', event: 'track', payload }, opts.timeout || this.timeout)
  //                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  //                                                          number passed where object is expected
}
```

## After

```ts
async track(payload, opts = {}) {
  return await this.send({ type: 'presence', event: 'track', payload }, opts)
}
```

### DIFF
--- a/packages/core/realtime-js/src/RealtimeChannel.ts
+++ b/packages/core/realtime-js/src/RealtimeChannel.ts
@@ -501,7 +501,7 @@ export default class RealtimeChannel {
         event: 'track',
         payload,
       },
-      opts.timeout || this.timeout
+      opts
     )
   }
 
--- a/packages/core/realtime-js/test/RealtimeChannel.presence.test.ts
+++ b/packages/core/realtime-js/test/RealtimeChannel.presence.test.ts
@@ -320,6 +320,47 @@ describe('Presence helper methods', () => {
 
     expect(resp).toBe('ok')
   })
+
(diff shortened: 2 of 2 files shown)


edgurgel:
@FrancoKaddour would you mind adding some form of test to ensure these are being passed properly? This way we won't break this in the future

FrancoKaddour:
Done — added three tests in RealtimeChannel.presence.test.ts: track() forwarding a
custom opts.timeout down to the underlying push, track() falling back to the channel
timeout when no opts are given, and untrack() forwarding a custom timeout as well.
They follow the same push-spy pattern already used in RealtimeChannel.messaging.test.ts.
Let me know if you'd prefer them structured differently.

edgurgel (approved):
LGTM thanks!

PR #2493, opened 2026-07-06
fix(auth): preserve default lock acquire timeout

### LINKED ISSUE #2492: Unset auth.lockAcquireTimeout is forwarded as undefined, clobbering the 5000ms default and disabling lock-steal recovery (regression from the #2308 fix)
### Describe the bug

When `auth.lockAcquireTimeout` is NOT set (the default for virtually every app), `createClient` forwards it to `GoTrueClient` as an explicit `undefined` property. `Object.assign({}, DEFAULT_OPTIONS, options)` copies own enumerable keys even when their value is `undefined`, so the auth-js default of `5000` ms is clobbered to `undefined`.

Downstream in `navigatorLock`, `undefined > 0` is `false`, so no acquire-timeout timer is ever armed, and since `undefined !== 0` the request waits on the Web Lock with a signal that never aborts. Net effect: **the 5000 ms default + steal-based orphaned-lock recovery (introduced in v2.98.0) is silently disabled for every supabase-js user on default config.** The wait on `lock:sb-<ref>-auth-token` is infinite.

This appears to be a regression introduced by the fix for #2308: before that fix, the unset option was simply absent from the forwarded object, so the `5000` default survived the merge. After it, the destructured-but-unset value is forwarded as `undefined` and overrides the default.

### Real-world impact

We hit this in production (observed on `@supabase/supabase-js` 2.105.4): a browser tab frozen by Chrome Memory Saver held the exclusive Web Lock, and every supabase-js query in every other tab of that browser profile queued behind it forever. Symptom: page stuck on a loading state with zero network requests and zero console errors, in Chrome and Edge simultaneously. `navigator.locks.query()` showed the held + pending exclusive lock. This hits any user who leaves a tab open overnight. The built-in steal recovery would have self-healed this in 5 seconds, but it never runs because the timeout arrives as `undefined`.

### To Reproduce

```ts

const plain = createClient(url, key)
console.log(plain.auth.lockAcquireTimeout)
// expected: 5000 (the documented default)
// actual:   undefined  -> navigatorLock waits forever, steal recovery never fires

const explicit = createClient(url, key, { auth: { lockAcquireTimeout: 5000 } })
console.log(explicit.auth.lockAcquireTimeout)
// actual: 5000 (workaround: set the option explicitly)
```

Full deadlock repro in a browser app using a default-config client:

```js
// hold the auth lock, simulating a frozen/suspended tab
navigator.locks.request('lock:sb-<project-ref>-auth-token', { mode: 'exclusive' }, () => new Promise(() => {}))
// then perform any supabase query / auth.getSession() in the same profile:
// it queues forever (no timeout warning, no steal, no error)
```

With `lockAcquireTimeout: 5000` passed explicitly, the same repro self-heals after 5 s with the expected warning: `Lock "lock:sb-..." was not released within 5000ms. ... Forcefully acquiring the lock to recover.`

### Root cause

`SupabaseClient.ts` `_initSupabaseAuthClient` destructures `lockAcquireTimeout` from the user's `auth` options and passes it as a shorthand property to `new SupabaseAuthClient({ ... lockAcquireTimeout ... })`. When the user never set it, the constructed object contains `lockAcquireTimeout: undefined`, and `GoTrueClient`'s `const settings = Object.assign({}, DEFAULT_OPTIONS, options)` lets that `undefined` override `DEFAULT_OPTIONS.lockAcquireTimeout: 5000`.

### Expected behavior

An unset option should fall back to the library default (5000 ms), preserving the orphaned-lock steal recovery. Possible one-line fixes while the lockless rework lands:

- filter `undefined` values before the merge in `GoTrueClient` (e.g. merge with a helper that skips `undefined`), or
- in `_initSupabaseAuthClient`, only include `lockAcquireTimeout` (and other optional keys) when they are not `undefined`, or
- default it at the forwarding site: `lockAcquireTimeout: lockAcquireTimeout ?? 5000`.

I realise the lockless coordination on master removes `navigator.locks` by default and will make this class of deadlock go away; this report is mainly about the currently released 2.x line, where default-config apps have the documented recovery silently disabled and a single frozen tab can hang a whole browser profile until then. `@supabase/ssr`'s `createBrowserClient` forwards `auth` options into `createClient`, so it is presumably affected the same way.

### System information

- OS: Windows 11 / any
- Browser: Chrome + Edge (any browser with `navigator.locks`)
- Version of supabase-js: 2.105.4
- Version of Node.js: 24.x (bug is browser-side)

pjpjq:
## Description

Fixes supabase/supabase-js#2492.

`SupabaseClient` was always forwarding `lockAcquireTimeout` to `SupabaseAuthClient`, even when the caller did not configure it. Because the forwarded property was explicitly `undefined`, auth-js' default merge treated it as an override and replaced the documented `5000` ms default with `undefined`.

This PR only includes `lockAcquireTimeout` in the auth client options when it is actually defined, so default-config clients keep auth-js' fallback while explicit values still pass through.

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm nx test:unit supabase-js --testPathPatterns=SupabaseAuthClient.test.ts`
- `pnpm prettier --check packages/core/supabase-js/src/SupabaseClient.ts packages/core/supabase-js/test/unit/SupabaseAuthClient.test.ts`
- `pnpm exec eslint packages/core/supabase-js/src/SupabaseClient.ts packages/core/supabase-js/test/unit/SupabaseAuthClient.test.ts` (0 errors; existing warnings only)
- `git diff --check`

Notes:

- `pnpm nx typecheck supabase-js` was blocked before running by an existing workspace sync warning from Nx.
- `pnpm nx lint supabase-js` was blocked by a missing optional Expo integration lint dependency (`eslint-config-expo/flat`), so I ran ESLint on the changed files directly instead.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -615,7 +615,7 @@ export default class SupabaseClient<
       throwOnError,
       experimental,
       fetch,
-      lockAcquireTimeout,
+      ...(lockAcquireTimeout !== undefined ? { lockAcquireTimeout } : {}),
       skipAutoInitialize,
       // auth checks if there is a custom authorizaiton header using this flag
       // so it knows whether to return an error when getUser is called with no session
--- a/packages/core/supabase-js/test/unit/SupabaseAuthClient.test.ts
+++ b/packages/core/supabase-js/test/unit/SupabaseAuthClient.test.ts
@@ -123,6 +123,11 @@ test('createClient with auth.experimental.passkey enables the admin passkey API'
 // surfaces both the production code AND these tests together so the
 // removal is mechanical.
 
+test('createClient should preserve the default auth.lockAcquireTimeout when unset', () => {
+  const supa = new SupabaseClient('https://example.supabase.com', 'supabaseKey')
+  expect((supa.auth as unknown as { lockAcquireTimeout: number }).lockAcquireTimeout).toBe(5000)
+})
+
 test('_initSupabaseAuthClient should pass through lockAcquireTimeout option', () => {
   const client = new SupabaseClient('https://example.supabase.com', 'supabaseKey')
   const authClient = client['_initSupabaseAuthClient'](


mandarini:
Thanks @pjpjq, the fix here is correct, but #2392 (shipped in 2.107.0) removed `navigator.locks` as the default coordination mechanism, so the infinite-wait this was guarding against no longer applies to default-config clients. Closing in favour of that broader fix.

PR #2495, opened 2026-07-06
fix(auth): validate session before user hydration in _recoverAndRefresh

### LINKED ISSUE #2494: auth-js: _recoverAndRefresh crashes with 'Cannot create property user on string' when the stored session fails JSON.parse, and the corrupted value is never cleaned up
# Bug report

## Describe the bug

When the persisted session in `storage` is a string that fails `JSON.parse` (e.g. a corrupted cookie), `getItemAsync` intentionally returns the **raw string** (`try { return JSON.parse(value) } catch { return value }`).

`_recoverAndRefresh()` then hits this branch **before** `_isValidSession` is consulted:

```js
const currentSession = (await getItemAsync(this.storage, this.storageKey));
if (currentSession && this.userStorage) {
  // ...
  currentSession.user = maybeUser?.user ?? userNotAvailableProxy();  // ← TypeError on a string
} else if (currentSession && !currentSession.user) {
  // ...
  currentSession.user = userNotAvailableProxy();                     // ← TypeError on a string
}
```

Assigning `.user` on a string primitive throws `TypeError: Cannot create property 'user' on string '...'` in strict mode. The surrounding `catch` only does `console.error(err); return;` — so:

1. the client silently behaves as "no session", **and**
2. `_removeSession()` is never reached, so **the corrupted value is never cleaned up**, and every subsequent `initialize()` repeats the same failure.

With `@supabase/ssr` cookie storage this is nasty in practice: a user whose chunked auth cookie (`sb-*-auth-token.0/.1`) got corrupted (e.g. concurrent writers across tabs interleaving chunk writes) is permanently treated as signed-out on every request, until the cookie is manually cleared. The library never repairs or removes the broken value.

Note that `__loadSession()` handles the same situation correctly: it checks `_isValidSession(maybeSession)` first and calls `_removeSession()` for invalid values. `_recoverAndRefresh()` performs the user-hydration assignments *before* that check.

## To Reproduce

Unit-level reproduction (no network needed):

```js
import { GoTrueClient } from '@supabase/auth-js'

const storage = {
  isServer: false,
  getItem: async () => '{"access_token":"x","refresh_token":"y"', // unparseable → returned as raw string
  setItem: async () => {},
  removeItem: async () => {},
}

const client = new GoTrueClient({ url: 'http://localhost:9999', storage, autoRefreshToken: false })
await client.initialize()
// console: TypeError: Cannot create property 'user' on string '{"access_token":"x","refresh_token":"y"'
// storage.removeItem is never called → the corrupted value persists forever
```

Observed in production on `2.103.2`; the same unguarded assignments are present in `2.109.0` (`_recoverAndRefresh`, both the `userStorage` branch and the `currentSession && !currentSession.user` branch).

## Expected behavior

An unparseable / non-object stored session should be treated like any other invalid session: skipped **and removed** (`_removeSession()`), exactly as `__loadSession()` already does — instead of throwing inside `_recoverAndRefresh` and leaving the broken value in storage.

Suggested fix: move the `_isValidSession(currentSession)` check (or at least a `typeof currentSession === 'object'` guard) above the user-hydration branches in `_recoverAndRefresh`.

## System information

- OS: any (observed on Vercel edge runtime + Node 20/24)
- Version of supabase-js: @supabase/supabase-js 2.103.2 (@supabase/auth-js 2.103.2, also inspected 2.109.0)
- Version of Node.js: 24.x

AmariahAK:
### What changed?

Reordered `_recoverAndRefresh` so `_isValidSession` validation and `_removeSession` cleanup run **before** user-hydration (`.user` assignment), matching the existing pattern in `__loadSession`.

**Files:**
- `packages/core/auth-js/src/GoTrueClient.ts` — moved the validation + cleanup block ahead of the user-hydration branches
- `packages/core/auth-js/test/GoTrueClient.test.ts` — 3 new unit tests (JSON string primitive, JSON number, null session)
- `packages/core/auth-js/test/GoTrueClient.browser.test.ts` — 1 new browser test

When a stored auth session was a valid JSON value but not a `Session` object (e.g. corrupted to `"hello"`, `123`, `true`), `_recoverAndRefresh` tried `currentSession.user = ...` before checking `_isValidSession`. In strict mode, assigning `.user` to a primitive throws `TypeError: Cannot create property 'user' on string`. The outer `catch` silently logged the error without calling `_removeSession()`, so the corrupted value was never cleaned up and every subsequent `initialize()` repeated the same failure.

`__loadSession` already handled this correctly — the fix mirrors that pattern.

Closes #2494

---

## 🔄 Breaking changes

---

## 📋 Checklist

---

## 📝 Additional notes

The change is a pure reorder — no new validation logic, no new functions. `_isValidSession` is a TypeScript type guard (`maybeSession is Session`), so after the early return on invalid, the compiler narrows `currentSession` to `Session` for the hydration code below.

**Test locally:**
```bash
cd packages/core/auth-js
node test/generate-signing-keys.js
./node_modules/.bin/jest --config jest.config.js \
  --testPathPatterns="GoTrueClient.test" \
  --testNamePattern="corrupted|no session exists" \
  --forceExit
```

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -4715,6 +4715,17 @@ export default class GoTrueClient {
     try {
       const currentSession = (await getItemAsync(this.storage, this.storageKey)) as Session | null
 
+      this._debug(debugName, 'session from storage', currentSession)
+
+      if (!this._isValidSession(currentSession)) {
+        this._debug(debugName, 'session is not valid')
+        if (currentSession !== null) {
+          await this._removeSession()
+        }
+
+        return
+      }
+
       if (currentSession && this.userStorage) {
         let maybeUser: { user: User | null } | null = (await getItemAsync(
           this.userStorage,
@@ -4752,17 +4763,6 @@ export default class GoTrueClient {
             currentSession.user = userNotAvailableProxy()
           }
         }
-      }
-
-      this._debug(debugName, 'session from storage', currentSession)
-
-      if (!this._isValidSession(currentSession)) {
-        this._debug(debugName, 'session is not valid')
-        if (currentSession !== null) {
-          await this._removeSession()
-        }
-
-        return
       }
 
       const expiresWithMargin =
--- a/packages/core/auth-js/test/GoTrueClient.browser.test.ts
+++ b/packages/core/auth-js/test/GoTrueClient.browser.test.ts
@@ -978,6 +978,27 @@ describe('Session Management Tests', () => {
     expect(client).toBeDefined()
   })
 
+  it('should remove corrupted storage value when stored session is a JSON primitive', async () => {
+    // JSON.parse('"corrupted"') returns the string "corrupted", not a Session
+    const mockStorage = {
+      getItem: jest.fn().mockResolvedValue('"corrupted"'),
+      setItem: jest.fn(),
+      removeItem: jest.fn(),
+    }
+
+    const client = new (require('../src/GoTrueClient').default)({
+      url: 'http://localhost:9999',
+      autoRefreshToken: false,
+      persistSession: true,
+      storage: mockStorage,
+    })
+
+    await client.initialize()
+
+    // The corrupted value should be removed
+    expect(mockStorage.removeItem).toHaveBeenCalled()
+  })
+
   it('should handle session with invalid tokens', async () => {
     const mockSession = {
       access_token: '',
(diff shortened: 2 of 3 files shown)


mandarini:
Hi @AmariahAK, thank you so much for the contribution and the thorough analysis! 💚 

After digging into this, the corruption scenario described in the issue (invalid JSON failing `JSON.parse`) was already addressed in PR #2336 (v2.106.0), which changed `getItemAsync` to return `null` on parse failure instead of the raw string. So the problematic code path in `_recoverAndRefresh` is no longer reachable via that trigger.

Since the root cause is already fixed at the `getItemAsync` layer, we would prefer not to introduce new code here. We are going to close the PR for now.

Thank you again for taking the time to trace through this so carefully, the issue report and the fix were both well-reasoned.

PR #2504, opened 2026-07-07
fix(auth): clear local session on signout failures

### LINKED ISSUE #2501: auth-js: signOut() throws before clearing local session when offline (all scopes, including 'local')
## Bug report

**Package**: `@supabase/auth-js` (via `@supabase/supabase-js`)
**Version observed**: 2.110.0
**File**: `packages/core/auth-js/src/GoTrueClient.ts`, method `_signOut`

### What happens

`GoTraClient._signOut({ scope })` always calls the network first:

```js
async _signOut({ scope } = { scope: 'global' }) {
  return await this._useSession(async (result) => {
    ...
    const accessToken = data.session?.access_token;
    if (accessToken) {
      const { error } = await this.admin.signOut(accessToken, scope); // <-- network call
      if (error) {
        if (!(/* 404/401/403/session-missing */)) {
          return this._returnResult({ error }); // returns, but at least reaches here
        }
      }
    }
    if (scope !== 'others') {
      await this._removeSession(); // <-- local cleanup, only reached if the above didn't throw
      await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`);
    }
    return this._returnResult({ error: null });
  });
}
```

If the client is **offline**, `this.admin.signOut(accessToken, scope)` performs a `fetch` that **throws** (network error), rather than resolving with `{ error }`. That exception propagates out of `_signOut()` uncaught, so `_removeSession()` is never reached — **the local session/token is never removed from storage**, regardless of `scope` (including `scope: 'local'`, which a caller would reasonably expect to never touch the network at all).

### Impact

An app calling `supabase.auth.signOut()` (or `{ scope: 'local' }`) while offline gets a rejected promise, but **the user is still authenticated locally** (valid session token still in storage) after the call — even though the app has already navigated away as if logged out. Reloading the page, or any code path that reads the persisted session, will find the user still "logged in" with no indication anything went wrong, unless the caller wraps every `signOut()` in a try/catch AND manually clears the storage key itself as a fallback (which defeats the purpose of having a public `signOut()` API).

### Expected behavior

- `scope: 'local'` should never perform a network call at all (it's documented as "sign out only on this device/browser"), or
- at minimum, `_removeSession()` should run in a `finally` (or before attempting the network call, followed by a best-effort network revoke) so that local session state is guaranteed to be cleared even if the network request fails/throws.

### Repro

```ts
// Simulate offline (e.g. via navigator.onLine / fetch mock)
const { error } = await supabase.auth.signOut(); // rejects (or resolves with error) while offline
// Session token is STILL present in the configured storage (e.g. localStorage[storageKey])
```

### Workaround we're using

We manually remove the storage key(s) ourselves when `signOut()` throws or returns an error, as a forced fallback, before treating the local logout as complete. Happy to open a PR if a fix direction is agreed (e.g. moving `_removeSession()` into a `finally`, or making `scope: 'local'` skip the network call entirely).

pjpjq:
Replaces #2503, which was closed when the fork branch was renamed from `codex/...` to `fix/...`.

## Description

Fixes supabase/supabase-js#2501.

`signOut()` previously returned immediately when the remote logout request produced a retryable/network auth error. For `global` and `local` sign-out scopes that meant the persisted browser/device session remained in storage, so an offline sign-out could reject while leaving the user locally authenticated.

This PR keeps the existing `others` semantics, but ensures `global` and `local` scopes always clear the current local session before returning the remote logout error.

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm --dir packages/core/auth-js exec jest --config jest.config.js test/GoTrueClient.signOut.test.ts --runInBand --silent=false`
- `pnpm nx build auth-js`
- `pnpm prettier --check packages/core/auth-js/src/GoTrueClient.ts packages/core/auth-js/test/GoTrueClient.signOut.test.ts`
- `pnpm exec eslint packages/core/auth-js/test/GoTrueClient.signOut.test.ts`
- `git diff --check`

Note: `pnpm exec eslint packages/core/auth-js/src/GoTrueClient.ts ...` is currently blocked by pre-existing lint errors in `GoTrueClient.ts` (unused variables / no-empty-object-type outside this diff), so I linted the new test file directly and relied on `pnpm nx build auth-js` for the changed source file.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -3963,6 +3963,10 @@ export default class GoTrueClient {
     { scope }: SignOut = { scope: 'global' }
   ): Promise<{ error: AuthError | null }> {
     return await this._useSession(async (result) => {
+      const removeCurrentSession = async () => {
+        await this._removeSession()
+        await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
+      }
       const { data, error: sessionError } = result
       if (sessionError && !isAuthSessionMissingError(sessionError)) {
         return this._returnResult({ error: sessionError })
@@ -3980,13 +3984,15 @@ export default class GoTrueClient {
               isAuthSessionMissingError(error)
             )
           ) {
+            if (scope !== 'others') {
+              await removeCurrentSession()
+            }
             return this._returnResult({ error })
           }
         }
       }
       if (scope !== 'others') {
-        await this._removeSession()
-        await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
+        await removeCurrentSession()
       }
       return this._returnResult({ error: null })
     })
--- a/packages/core/auth-js/test/GoTrueClient.signOut.test.ts
+++ b/packages/core/auth-js/test/GoTrueClient.signOut.test.ts
@@ -1,0 +1,78 @@
+import GoTrueClient from '../src/GoTrueClient'
+import { memoryLocalStorageAdapter } from '../src/lib/local-storage'
+
+const createStoredSession = () => ({
+  access_token: 'test-access-token',
+  refresh_token: 'test-refresh-token',
+  token_type: 'bearer',
+  expires_in: 3600,
+  expires_at: Math.floor(Date.now() / 1000) + 3600,
+  user: {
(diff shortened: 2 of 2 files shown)


mandarini (approved):
Thank you

PR #2508, opened 2026-07-10
chore(ci): consolidate node version references

kanadgupta:
## 🔍 Description

Tiny refactor to make it so all of the GitHub Actions workflow files rely on `.nvmrc` as its single source of truth for the Node.js version to use.

### What changed?

(see above)

n/a

## 🔄 Breaking changes

## 📋 Checklist


mandarini (dismissed):
This is a nice cleanup. But, I found one thing that needs to change before this can merge: `deprecate-version.yml` doesn't have a checkout step in its `deprecate` job (it only calls `npm deprecate` against the registry, so it never needed the repo checked out before). Now that the `setup-node` step there uses `node-version-file: '.nvmrc'`, it needs to read that file from the workspace, and without a checkout step it won't be there. Could you add an `actions/checkout` step before `setup-node` in that job, same as the other workflows in this PR? Since this workflow is `workflow_dispatch` only, none of our CI checks would catch this, so it'd only show up the next time someone tries to deprecate a package.

Also, two workflows still hardcode `node-version: '22'` and got left out of this pass: `docs.yml` and `preview-release.yml`. Would you mind updating those too so we don't end up with a couple of stragglers to fix the next time the Node version changes?

kanadgupta:
@mandarini this is great feedback, thanks for catching these! I addressed your feedback in https://github.com/supabase/supabase-js/pull/2508/commits/9e86564fa21605d908564fe95a465d6bfa4c920d and https://github.com/supabase/supabase-js/pull/2508/commits/1df6e601430a4ad2e16cdf9457074374b3ec8f9d.

mandarini (approved):

PR #2511, opened 2026-07-10
fix(functions): stop sending API key in Authorization header for function calls

mandarini:
## Description

Edge Function calls (`supabase.functions.invoke()`) no longer put the project API key in the `Authorization` header when there is no user session. The key is sent only in the `apikey` header, and `Authorization` is reserved for a real user (or custom) session token, matching the Server SDK pattern.

### What changed?

- `packages/core/supabase-js/src/lib/fetch.ts`: `fetchWithAuth` accepts a new optional `{ isFunctionsClient }` flag. When set, it no longer falls back to sending the API key as `Authorization: Bearer <key>`, but only for new-format keys (`sb_publishable_...` / `sb_secret_...`). A real session/custom token is still sent, and the `apikey` header is unchanged.
- `packages/core/supabase-js/src/SupabaseClient.ts`: added a dedicated `functionsFetch` built with `{ isFunctionsClient: true }`. The `functions` getter uses it; REST (PostgREST), Storage, and Realtime keep the shared `fetch` and are unaffected.
- `packages/core/functions-js/src/FunctionsClient.ts`: updated the `invoke` JSDoc to describe the `apikey` vs `Authorization` header semantics.
- `docs/MIGRATION.md`: added an "Edge Functions auth headers" section documenting the behavior and the one edge case that is affected.
- Tests: added unit coverage in `fetch.test.ts` for the `isFunctionsClient` option and an integration case in `SupabaseClient.test.ts` proving functions omit `Authorization` for a new-format key with no session while other services do not.

New-format API keys (`sb_publishable_...` / `sb_secret_...`) are not JWTs, so sending them as `Authorization: Bearer <key>` caused the gateway to reject the request with "invalid token format." A platform apikey-compatibility patch (now GA) lets `verify_jwt=true` functions be called with only the `apikey` header, so the SDK no longer needs to duplicate the key into `Authorization`. This aligns Functions with the Server SDK convention: `Authorization` is for user/custom tokens only.

Closes SDK-1050.

## Screenshots/Examples

Unauthenticated function call with a new-format key:

Before:
```
apikey: sb_publishable_...
Authorization: Bearer sb_publishable_...   # rejected as "invalid token format"
```

After:
```
apikey: sb_publishable_...
# no Authorization header
```

Authenticated call (unchanged):
```
apikey: sb_publishable_...
Authorization: Bearer <user_jwt>
```

## Breaking changes

The change is scoped to Edge Functions only, and legacy JWT keys keep their existing behavior (still sent in `Authorization` for backward compatibility), so this is not a breaking change in practice. The only impacted case is code that reads the API key out of the `Authorization` header inside an Edge Function; those should read the `apikey` header instead, or migrate to `@supabase/server`.

## Checklist

- [ ] I have run `pnpm nx format` to ensure consistent code formatting

## Additional notes

Scope is intentionally limited to new-format keys so there is zero regression risk for self-hosted or legacy-key users on older gateways. Moving to an unconditional drop for all key types later is a one-line change (removing the key-format guard in `fetch.ts`), best done on v3 or once apikey-compatibility is universal.

### DIFF
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -89,6 +89,7 @@ export default class SupabaseClient<
   protected rest: PostgrestClient<Database, ClientOptions, SchemaName>
   protected storageKey: string
   protected fetch?: Fetch
+  protected functionsFetch?: Fetch
   protected changedAccessToken?: string
   protected accessToken?: () => Promise<string | null>
 
@@ -364,6 +365,17 @@ export default class SupabaseClient<
       settings.global.fetch,
       settings.tracePropagation
     )
+    // Edge Functions use a dedicated fetch that reserves the Authorization header for a
+    // real user/custom session token — it never falls back to a new-format API key
+    // (that key travels only in the `apikey` header). Other services keep `this.fetch`.
+    this.functionsFetch = fetchWithAuth(
+      supabaseKey,
+      supabaseUrl,
+      this._getAccessToken.bind(this),
+      settings.global.fetch,
+      settings.tracePropagation,
+      { isFunctionsClient: true }
+    )
     this.realtime = this._initRealtimeClient({
       headers: this.headers,
       accessToken: this._getAccessToken.bind(this),
@@ -404,7 +416,7 @@ export default class SupabaseClient<
   get functions(): FunctionsClient {
     return new FunctionsClient(this.functionsUrl.href, {
       headers: this.headers,
-      customFetch: this.fetch,
+      customFetch: this.functionsFetch,
     })
   }
 
--- a/packages/core/supabase-js/src/SupabaseClient.ts
+++ b/packages/core/supabase-js/src/SupabaseClient.ts
@@ -20,7 +20,7 @@ import {
   DEFAULT_REALTIME_OPTIONS,
   DEFAULT_TRACE_PROPAGATION_OPTIONS,
 } from './lib/constants'
-import { fetchWithAuth } from './lib/fetch'
+import { assertSupportedApiKey, fetchWithAuth } from './lib/fetch'
 import {
   applySettingDefaults,
   validateSupabaseUrl,
@@ -315,6 +315,7 @@ export default class SupabaseClient<
   ) {
     const baseUrl = validateSupabaseUrl(supabaseUrl)
     if (!supabaseKey) throw new Error('supabaseKey is required.')
+    assertSupportedApiKey(supabaseKey)
 
     this.realtimeUrl = new URL('realtime/v1', baseUrl)
     this.realtimeUrl.protocol = this.realtimeUrl.protocol.replace('http', 'ws')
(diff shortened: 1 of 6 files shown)


kallebysantos (dismissed):

Tr00d (dismissed):

grdsdev (dismissed):

packages/core/supabase-js/src/lib/fetch.ts
```diff
@@ -21,12 +21,21 @@ export const resolveHeadersConstructor = () => {
...
+  key.startsWith('sb_publishable_') || key.startsWith('sb_secret_')
+
 export const fetchWithAuth = (
   supabaseKey: string,
   supabaseUrl: string,
   getAccessToken: () => Promise<string | null>,
   customFetch?: Fetch,
-  tracePropagationOptions?: TracePropagationOptions
+  tracePropagationOptions?: TracePropagationOptions,
+  options?: { isFunctionsClient?: boolean }
```
spydon: This couples a generic fetch util to a specific consumer (fetch.ts shouldn't know a "functions client" exists). Maybe name the option by the behavior it controls instead, e.g. { omitApiKeyAsBearer?: boolean }?

packages/core/supabase-js/src/lib/fetch.ts
```diff
@@ -48,7 +57,15 @@ export const fetchWithAuth = (
     }
 
     if (!headers.has('Authorization')) {
-      headers.set('Authorization', `Bearer ${accessToken}`)
+      // Edge Functions: a new-format API key must never be sent as a Bearer token —
+      // it belongs only in the `apikey` header. Reserve Authorization for a real user
+      // or custom session token. `accessToken === supabaseKey` means there is no such
+      // token (getAccessToken falls back to the key). Legacy JWT keys are unchanged.
+      const omitKeyInAuth =
+        options?.isFunctionsClient && accessToken === supabaseKey && isNewApiKey(supabaseKey)
```
spydon: LLM comment:

The accessToken === supabaseKey check is reverse-engineering something we already know. _getAccessToken() coalesces session ?? supabaseKey before this wrapper sees it, so we compare back to the key to guess "was there a real session?". If we don't pre-coalesce (line 52) and instead keep the raw getAccessToken() result, the distinction is direct: const bearer = realToken ?? (allowKeyAsBearer ? supabaseKey : null); if (bearer) headers.set('Authorization', ...). No equality proxy, no coupling to _getAccessToken's fallback.

spydon: `isNewApiKey(supabaseKey)` is static per instance, so maybe move it out of the per-request path like traceEnabled above (e.g. const allowKeyAsBearer = !options?.omitApiKeyAsBearer || !isNewApiKey(supabaseKey)), computed once at construction.

packages/core/supabase-js/src/lib/fetch.ts
```diff
@@ -21,12 +21,21 @@ export const resolveHeadersConstructor = () => {
   return Headers
 }
 
+/**
+ * New-format Supabase API keys (`sb_publishable_…` / `sb_secret_…`) are not JWTs and
+ * must never be sent as a Bearer token — they belong only in the `apikey` header.
+ * Legacy (JWT) keys predate this format and keep their existing behavior.
+ */
+const isNewApiKey = (key: string): boolean =>
+  key.startsWith('sb_publishable_') || key.startsWith('sb_secret_')
```
spydon, line 32: The rationale ("new-format keys aren't JWTs, don't send as Bearer") is documented three times: here, in the inline comment at line 61, and in SupabaseClient.ts. Maybe collapse to one?

packages/core/supabase-js/src/SupabaseClient.ts
```diff
@@ -364,6 +365,17 @@ export default class SupabaseClient<
...
+    // Edge Functions use a dedicated fetch that reserves the Authorization header for a
+    // real user/custom session token — it never falls back to a new-format API key
+    // (that key travels only in the `apikey` header). Other services keep `this.fetch`.
+    this.functionsFetch = fetchWithAuth(
+      supabaseKey,
+      supabaseUrl,
+      this._getAccessToken.bind(this),
+      settings.global.fetch,
+      settings.tracePropagation,
+      { isFunctionsClient: true }
```
spydon: If you agree with the change on fetch.ts:38, a change is needed here too

kallebysantos (approved):
LGTM 💚

spydon (approved):

PR #2514, opened 2026-07-12
refactor(auth): build PKCE verifier from a Uint8Array and drop deprecated substr

PedroHenrique0713:
### What changed?

`generatePKCEVerifier()` built the PKCE code verifier from a `Uint32Array(56)` but kept only the low byte of each element (via `dec2hex`). This switches it to a `Uint8Array(56)` and replaces the deprecated `String.prototype.substr` with `slice`. The typed-array allocation was also moved inside the `crypto` branch so the no-WebCrypto fallback no longer allocates an unused buffer.

- **Over-allocation of CSPRNG output:** `new Uint32Array(56)` draws 224 bytes from `crypto.getRandomValues`, but `dec2hex` (`('0' + dec.toString(16)).substr(-2)`) keeps only the last hex pair — the low byte — discarding 3 of every 4 bytes. A `Uint8Array(56)` requests exactly the 56 bytes actually consumed.
- **Deprecated API:** `String.prototype.substr()` is a [legacy/deprecated feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr); `slice(-2)` is the standard equivalent.

**No behavior change.** The output is byte-for-byte equivalent: 56 bytes → two hex chars each → the same 112-character, hex-only verifier (well within RFC 7636's 43–128 range). The low byte of a uniform `Uint32` is itself uniform, so the entropy distribution is identical — this only stops requesting (and discarding) 4× the randomness.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

Added a `generatePKCEVerifier` test block (length within the RFC 7636 range, unreserved charset, uniqueness). Verified locally: `jest test/helpers.test.ts` → 27 passed, `tsc -p tsconfig.json` → clean, `prettier --check` → clean.

### DIFF
--- a/packages/core/auth-js/src/lib/helpers.ts
+++ b/packages/core/auth-js/src/lib/helpers.ts
@@ -259,13 +259,12 @@ export function retryable<T>(
 }
 
 function dec2hex(dec: number) {
-  return ('0' + dec.toString(16)).substr(-2)
+  return ('0' + dec.toString(16)).slice(-2)
 }
 
 // Functions below taken from: https://stackoverflow.com/questions/63309409/creating-a-code-verifier-and-challenge-for-pkce-auth-on-spotify-api-in-reactjs
 export function generatePKCEVerifier() {
   const verifierLength = 56
-  const array = new Uint32Array(verifierLength)
   if (typeof crypto === 'undefined') {
     const charSet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~'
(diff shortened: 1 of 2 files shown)


PedroHenrique0713:
Closing this one. It is a refactor with no behaviour change — the verifier is byte-for-byte identical either way — so it costs review time without fixing anything a user can hit, and the queue is better spent on the PRs that do. Happy to reopen if the reduced CSPRNG draw or the deprecated `substr` is worth picking up on its own.

PR #2515, opened 2026-07-12
fix(functions): match response Content-Type case-insensitively

PedroHenrique0713:
### What changed?

`FunctionsClient.invoke()` matched the response `Content-Type` case-sensitively when deciding how to parse the body (`responseType === 'application/json'`, etc.). This normalizes the media type to lower case before matching.

HTTP media types are case-insensitive ([RFC 9110 §8.3.1](https://www.rfc-editor.org/rfc/rfc9110#name-media-type)). If an Edge Function responds with `Content-Type: Application/JSON` (or `APPLICATION/JSON`), the client currently falls through to the `text` branch and returns the raw, unparsed string instead of the parsed JSON object.

It's also inconsistent with the **request** side, which already detects the caller's `Content-Type` case-insensitively (`key.toLowerCase() === 'content-type'`); only the response side wasn't.

## 🔄 Breaking changes

The compared media types are all lower case already, so lower-casing the incoming value only makes matching more tolerant.

## 📋 Checklist

## 📝 Additional notes

Added `test/spec/response-content-type.spec.ts` — a unit test (no relay) using a mocked `customFetch` asserting that lowercase, mixed-case, and uppercase `application/json` responses all parse as JSON. Verified locally: the mixed/upper cases fail without the change and pass with it; `jest` (new spec) → 3 passed, `tsc --noEmit` → clean, `prettier --check` → clean.

### DIFF
--- a/packages/core/functions-js/src/FunctionsClient.ts
+++ b/packages/core/functions-js/src/FunctionsClient.ts
@@ -298,7 +298,12 @@ export class FunctionsClient {
         throw new FunctionsHttpError(response)
       }
 
-      let responseType = (response.headers.get('Content-Type') ?? 'text/plain').split(';')[0].trim()
+      // HTTP media types are case-insensitive (RFC 9110), so normalize before
+      // matching — otherwise an "Application/JSON" response falls through to text.
+      let responseType = (response.headers.get('Content-Type') ?? 'text/plain')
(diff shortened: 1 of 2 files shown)


kallebysantos (approved):

kallebysantos:
Vlw ai manim 💚

PR #2516, opened 2026-07-12
fix(realtime): encode broadcast header fields as UTF-8

PedroHenrique0713:
### What changed?

In the binary broadcast serializer, the header fields (`join_ref`, `ref`, `topic`, `event`, `metadata`) are now length-prefixed and written as **UTF-8 bytes**, instead of using `String.length` for the size prefix and `charCodeAt(0)` for the bytes.

The decode side reads these fields with `TextDecoder` (UTF-8), but the encode side measured them with `String.length` (UTF-16 code units) and wrote them with `char.charCodeAt(0)` (one truncated byte per code unit). For any non-ASCII value — e.g. a broadcast `event` name like `café` or an emoji — this:

- writes the wrong length prefix (`café` → `4` instead of `5` UTF-8 bytes),
- corrupts multi-byte characters (`café` decodes back as `caf�`),
- desynchronizes the whole buffer for characters outside the BMP (`String.length` counts 2 code units for `🎉`, but `Array.from(...)` iterated 1 code point, so the declared size and the bytes written disagreed).

Using `TextEncoder` + `Uint8Array.set()` makes the encode side byte-accurate and consistent with the UTF-8 decode. ASCII values are unaffected (UTF-8 bytes equal the char codes and byte length equals string length), so the existing tests are unchanged.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

Added a `binary` test that encodes a non-ASCII event (`café-🎉`) and asserts the length prefix equals the UTF-8 byte length and that the written bytes round-trip through `TextDecoder`. The test fails on `master` (`expected 7 to be 10`) and passes with this change. Verified locally: `vitest run test/serializer.test.ts` → 24 passed, `tsc --noEmit` → clean, `prettier --check` → clean.

### DIFF
--- a/packages/core/realtime-js/src/lib/serializer.ts
+++ b/packages/core/realtime-js/src/lib/serializer.ts
@@ -62,19 +62,24 @@ export default class Serializer {
     encodingType: number,
     encodedPayload: ArrayBuffer
   ) {
-    const topic = message.topic
-    const ref = message.ref ?? ''
-    const joinRef = message.join_ref ?? ''
-    const userEvent = message.payload.event
+    // Encode each header field as UTF-8. The length prefixes are byte counts and
+    // the decode side uses TextDecoder (UTF-8), so measuring with String.length
+    // and writing with charCodeAt would corrupt any multi-byte character (e.g.
+    // accents or emoji) and desynchronize the buffer.
+    const encoder = new TextEncoder()
+    const topic = encoder.encode(message.topic)
+    const ref = encoder.encode(message.ref ?? '')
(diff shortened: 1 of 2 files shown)


PedroHenrique0713:
Thanks for the re-run, @edgurgel. This one looks like the integration job hitting a Docker Hub pull rate limit (`toomanyrequests: Rate exceeded`) while pulling the service images, rather than a test regression — the realtime broadcast tests passed in both attempts. Happy to rebase or push whenever it's useful to trigger CI again.

edgurgel (approved):
LGTM! Thanks!

PR #2520, opened 2026-07-13
fix(postgrest): type hinted self-referencing embeds as arrays

### LINKED ISSUE #2517: Self-referencing FK hint: _fkey suffix truncated at runtime, wrong types with column hint
### Describe the bug

Self-referencing table embeds behave inconsistently depending on whether the relationship hint is a column name or a foreign key constraint name.

Table:

```sql
CREATE TABLE public.lab (
  id     uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  name   text,
  type   text,
  parent uuid REFERENCES lab(id) ON DELETE CASCADE,  -- constraint: lab_parent_fkey
  main   uuid REFERENCES lab(id) ON DELETE CASCADE   -- constraint: lab_main_fkey
);
```

**Case 1 — column name hint (`!parent`)**

```ts
const query = supabase.from("lab").select("*,lab!parent(name,id)").eq("type", "board");
```

- Runtime response: correct, `lab` is returned as an array of children.
- Inferred type: incorrect — `lab` is typed as `{ name: string | null; id: string } | null` instead of an array.

**Case 2 — foreign key constraint name hint (`!lab_parent_fkey`)**

```ts
const query = supabase.from("lab").select("*,lab!lab_parent_fkey(name,id)").eq("type", "board");
```

- Inferred type: correct — `lab: { name: string | null; id: string }[]`.
- Runtime response: fails with PGRST200:

```json
{
    "code": "PGRST200",
    "details": "Searched for a foreign key relationship between 'lab' and 'lab' using the hint 'lab_parent' in the schema 'public', but no matches were found.",
    "hint": null,
    "message": "Could not find a relationship between 'lab' and 'lab' in the schema cache"
}
```

Note that the hint received by PostgREST is `lab_parent`, **not** `lab_parent_fkey` as written in the query. Confirmed via the browser Network tab that the actual outgoing request also has the truncated hint — this is not a typo in our code. The `_fkey` suffix appears to be stripped somewhere in the select-string building for self-referencing embeds.

### Expected behavior

- `lab!lab_parent_fkey(name,id)` should send the full constraint name `lab_parent_fkey` to PostgREST and resolve as one-to-many (array).
- `lab!parent(name,id)` should infer the correct array type at compile time, matching runtime behavior.

### Library affected

supabase-js

### Reproduction

_No response_

### Steps to reproduce

1. Create a table with two self-referencing foreign keys as above.
2. Run both queries shown.
3. Compare inferred TS types vs actual runtime response, and inspect the outgoing request URL for Case 2.

### System Info

```shell
supabase-js version: 2.110.2
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

mandarini:
Self-referencing embeds resolved via an explicit hint (e.g. `collections!parent_id(*)` or `collections!collections_parent_id_fkey(*)`) were incorrectly typed as a single object when the hint matched a column name, while returning an array at runtime.

PostgREST only ever resolves the one-to-many direction for a hinted self-reference, so the result is always an array regardless of what the hint matched. Check for an explicit hint before the `match: 'col'` single-object branch, which now applies only to the no-hint case (`parent_id(*)`). Add regression tests for both hint forms.

Fixes #2517

### DIFF
--- a/packages/core/postgrest-js/src/select-query-parser/result.ts
+++ b/packages/core/postgrest-js/src/select-query-parser/result.ts
@@ -459,18 +459,17 @@ type ProcessEmbeddedResourceResult<
                   : ProcessedChildren | null
                 : ProcessedChildren | null
               : ProcessedChildren[]
-          : // If the relation is a self-reference it'll always be considered as reverse relationship
+          : // If the relation is a self-reference it'll always be considered as a reverse relationship.
+            // PostgREST returns arrays for a self-reference when using the table name
+            // (collections(*)) or any hint (collections!parent_id(*)), and a single object
+            // only when using the column name directly (parent_id(*)).
+            // See https://github.com/PostgREST/postgrest/blob/8776ece7d5fd612fad44151f9be4dffd855b28f0/src/PostgREST/Plan.hs#L633-L644
             Resolved['relation']['referencedRelation'] extends CurrentTableOrView
-            ? // When an explicit hint is used to embed a self-reference (eg: collections!parent_id(*)
-              // or collections!collections_parent_id_fkey(*)), PostgREST only ever resolves the
-              // one-to-many direction, so the result is always an array. See PostgREST's `findRel`
-              // `relIsSelf`/`Just hint` branch, which has no many-to-one/one-to-one match for
-              // self-relationships. This holds regardless of what the hint matched (column name,
-              // fk constraint name, or referenced relation), so it must be checked before `match`.
+            ? // An explicit hint must be checked before `match: 'col'`, since a hinted
+              // self-reference is always an array (eg: collections!parent_id(*)).
               Resolved['relation'] extends { hint: string }
               ? ProcessedChildren[]
-              : // Without a hint it can either be a reverse reference via a column inclusion
-                // (eg: parent_id(*)) in such case the result will be a single object
+              : // Without a hint, a column inclusion (eg: parent_id(*)) is a single object.
                 Resolved['relation']['match'] extends 'col'
                 ? IsRelationNullable<
                     TablesAndViews<Schema>[CurrentTableOrView],
@@ -478,8 +477,7 @@ type ProcessEmbeddedResourceResult<
                   > extends true
                   ? ProcessedChildren | null
                   : ProcessedChildren
-                : // Or it can be a reference via the reference relation (eg: collections(*))
-                  // in such case, the result will be an array of all the values (all collection with parent_id being the current id)
+                : // The table name (eg: collections(*)) is an array of children.
                   ProcessedChildren[]
             : // Otherwise if it's a non self-reference reverse relationship it's a single object
               IsRelationNullable<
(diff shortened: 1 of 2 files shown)


packages/core/postgrest-js/src/select-query-parser/result.ts
```diff
@@ -461,18 +461,26 @@ type ProcessEmbeddedResourceResult<
...
-                : ProcessedChildren
-              : // Or it can be a reference via the reference relation (eg: collections(*))
-                // in such case, the result will be an array of all the values (all collection with parent_id being the current id)
-                ProcessedChildren[]
+            ? // When an explicit hint is used to embed a self-reference (eg: collections!parent_id(*)
+              // or collections!collections_parent_id_fkey(*)), PostgREST only ever resolves the
+              // one-to-many direction, so the result is always an array. See PostgREST's `findRel`
+              // `relIsSelf`/`Just hint` branch, which has no many-to-one/one-to-one match for
+              // self-relationships. This holds regardless of what the hint matched (column name,
+              // fk constraint name, or referenced relation), so it must be checked before `match`.
```
laurenceisla: I understand the matching needs to be done this way in TS. However it's a bit confusing to read the explanation like this. Would it be better to have a general explanation on how the Array or Single Object returns are generated before this? Something like:

```
PostgREST returns arrays in these cases:
* When using the table name in the self reference, e.g. collections(*)
* When using any type of hint in the self reference , e.g. collections!parent_id(*)
And it only returns a single object:
* When using the column name directly, e.g. parent_id(*)
See: https://github.com/PostgREST/postgrest/blob/8776ece7d5fd612fad44151f9be4dffd855b28f0/src/PostgREST/Plan.hs#L633-L644
```

This would simplify the comments. For example, this one could say:

```
We need to verify if an explicit hint is used first, before checking if the reference is done via a column inclusion. For example, `collection!parent_id(*)` returns an array.
```

The above is just an idea, can be better phrased according to the code/other comments.

packages/core/postgrest-js/test/relationships.test.ts
```diff
@@ -541,6 +541,103 @@ test('self reference relation via column', async () => {
...
+  ExpectedSchema.parse(res.data)
+})
+
+// Documents a known PostgREST limitation: a self-referencing embed cannot be
+// disambiguated by its foreign key constraint name. PostgREST's `findRel`
+// `relIsSelf`/`Just hint` branch only matches the hint against a column name
+// (`matchFKRefSingleCol`), never a constraint name (`matchConstraint`), so this
+// query fails at runtime with PGRST200. Use the column name hint instead (see the
+// test above). The inferred type is still (correctly) an array. See
+// https://github.com/supabase/supabase-js/issues/2517
```
laurenceisla, line 599: Same here, we also could remove the code references in this comment:

```suggestion
// Documents a known PostgREST limitation: a self-referencing embed cannot be
// disambiguated by its foreign key constraint name, so this
// query fails at runtime with PGRST200. Use the column name hint instead (see the
// test above). The inferred type is still (correctly) an array. See
// https://github.com/supabase/supabase-js/issues/2517
```

mandarini: addressed in lataest commit

laurenceisla (approved):

PR #2525, opened 2026-07-15
fix(postgrest): correct self-reference inference

### LINKED ISSUE #2517: Self-referencing FK hint: _fkey suffix truncated at runtime, wrong types with column hint

7ttp:
## TL;DR

Extends #2520 to cover the remaining direct column self reference case in `2.110.6-canary.0`

`parents:parent(*)` works at runtime but was still typed as `SelectQueryError` 
when multiple self referencing FKs exist

Direct column matches now resolve correctly while real table name ambiguity errors remain covered

## ref

- closes #2517 
 basically [the remaining case](https://github.com/supabase/supabase-js/issues/2517#issuecomment-4978100601)
- followup to: https://github.com/supabase/supabase-js/pull/2520


mandarini (changes requested):
Thank youuuu!!! :D 

One thing I'd love to see before I merge: the new `MultipleSelfReferencesDatabase` type and `lab` table are hand-written and don't correspond to anything in our actual test database. Every other test in `relationships.test.ts` runs against the real schema in `test/supabase/migrations/00000000000000_schema.sql`, where `collections` only has a single self-referencing FK. The tests you added are compile-time only (the queries are built but never awaited), so nothing actually exercises this against a real PostgREST instance with two self-referencing FKs on the same table.

Would you be up for adding a second self-referencing FK to an existing table (or a small new one) in the migration, and turning at least the `parents:parent(*)` case into a runtime-executed test like the other self-reference tests in that file? That would give end-to-end confidence that this resolves the way we expect, not just that the types compile.

7ttp:
> Thank youuuu!!! :D
> 
> One thing I'd love to see before I merge: the new `MultipleSelfReferencesDatabase` type and `lab` table are hand-written and don't correspond to anything in our actual test database. Every other test in `relationships.test.ts` runs against the real schema in `test/supabase/migrations/00000000000000_schema.sql`, where `collections` only has a single self-referencing FK. The tests you added are compile-time only (the queries are built but never awaited), so nothing actually exercises this against a real PostgREST instance with two self-referencing FKs on the same table.
> 
> Would you be up for adding a second self-referencing FK to an existing table (or a small new one) in the migration, and turning at least the `parents:parent(*)` case into a runtime-executed test like the other self-reference tests in that file? That would give end-to-end confidence that this resolves the way we expect, not just that the types compile.

absolutely! 💚🙂
added this in the latest commit: https://github.com/supabase/supabase-js/pull/2525/commits/246c5650e8d587e972e0bed6fcda0c96d7d375ba

mandarini (approved):

PR #2537, opened 2026-07-19
fix(storage): expose service error code on StorageApiError

### LINKED ISSUE #2536: StorageApiError does not expose documented Storage error codes (error)
### Describe the bug

The Storage error handling documentation at [https://supabase.com/docs/guides/storage/debugging/error-codes](url) describes using Storage error codes such as AccessDenied, NoSuchKey, and ResourceAlreadyExists to identify and handle errors. However, the StorageApiError object returned by @supabase/supabase-js does not expose these codes, it only exposes [ 'stack', 'message', '__isStorageError', 'namespace', 'name', 'status', 'statusCode' ].

As a result, it is not possible to implement the error handling pattern described in the documentation.

If this is intentional, the Storage error documentation appears to describe information that is not exposed by the JavaScript SDK. If it is unintentional, exposing the Storage error code would allow applications to implement the documented error handling strategy without relying on HTTP status codes or parsing error messages.

### Library affected

storage-js

### Reproduction

_No response_

### Steps to reproduce

const { error } = await supabase.storage
  .from("bucket")
  .download("missing-file");

if (error) {
  console.log(error.error); // "NoSuchKey"
}

### System Info

```shell
System:
    OS: Linux 6.18 Arch Linux
    CPU: (6) x64 AMD Ryzen 5 4500U with Radeon Graphics
    Memory: 1.17 GB / 7.15 GB
    Container: Yes
    Shell: 5.9.2 - /usr/bin/zsh
Binaries:
    Node: 26.4.0 - /usr/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 12.0.1 - /usr/bin/npm
    Deno: 2.9.3 - /usr/bin/deno
Browsers:
    Chromium: 150.0.7871.128
    Firefox: 152.0.6
    Firefox Developer Edition: 152.0.6
npmPackages:
    @supabase/ssr: ^0.12.3 => 0.12.3 
    @supabase/supabase-js: ^2.110.7 => 2.110.7
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

thribhuvan003:
### What changed?

`StorageApiError` now carries the service error code (`NoSuchKey`, `AccessDenied`, `ResourceAlreadyExists`, etc.) on a dedicated `code` property, and includes it in `toJSON()`.

The response body already contains this in `err.code` — `handleError` in `storage-js/src/lib/common/fetch.ts` was only using it as a fallback for `statusCode` and otherwise dropping it, so it never reached the error object. Now it is passed through to the new field. `statusCode` is unchanged.

The [Storage error-codes docs](https://supabase.com/docs/guides/storage/debugging/error-codes) tell you to branch on codes like `AccessDenied` / `NoSuchKey`, but the thrown `StorageApiError` never exposed them — it only had `status` and `statusCode` — so that pattern was impossible to implement without parsing the message string. This closes that gap.

Closes #2536

## 📸 Screenshots/Examples

```ts
const { error } = await supabase.storage.from("bucket").download("missing.png")
if (error && error.name === "StorageApiError" && error.code === "NoSuchKey") {
  // handle the specific case
}
```

## 🔄 Breaking changes

Additive only — a new optional constructor arg and a new field. `code` is `undefined` when the response has no code, and existing `status` / `statusCode` behaviour is untouched.

## 📋 Checklist

## 📝 Additional notes

Added unit tests in both `errors.test.ts` (the `code` field + `toJSON`) and `fetch.test.ts` (that a `code` in the response body reaches the error) — they fail without the change and pass with it.

### DIFF
--- a/packages/core/storage-js/src/lib/common/errors.ts
+++ b/packages/core/storage-js/src/lib/common/errors.ts
@@ -58,17 +58,26 @@ export function isStorageError(error: unknown): error is StorageError {
 export class StorageApiError extends StorageError {
   override status: number
   override statusCode: string
+  /**
+   * Service-specific error code from the Storage API response body, such as
+   * `NoSuchKey`, `AccessDenied` or `ResourceAlreadyExists`. Use this to branch
+   * on the specific error rather than parsing the message.
+   * @see https://supabase.com/docs/guides/storage/debugging/error-codes
+   */
+  code: string | undefined
 
   constructor(
     message: string,
     status: number,
     statusCode: string,
-    namespace: ErrorNamespace = 'storage'
+    namespace: ErrorNamespace = 'storage',
+    code?: string
   ) {
     super(message, namespace, status, statusCode)
     this.name = namespace === 'vectors' ? 'StorageVectorsApiError' : 'StorageApiError'
     this.status = status
     this.statusCode = statusCode
+    this.code = code
   }
 
   toJSON(): {
@@ -76,9 +85,11 @@ export class StorageApiError extends StorageError {
     message: string
     status: number | undefined
     statusCode: string | undefined
+    code: string | undefined
   } {
     return {
       ...super.toJSON(),
+      code: this.code,
     }
   }
 }
--- a/packages/core/storage-js/src/lib/common/fetch.ts
+++ b/packages/core/storage-js/src/lib/common/fetch.ts
@@ -76,7 +76,9 @@ const handleError = async (
       .then(
         (err: { statusCode?: string; code?: string; error?: string; message?: string } | null) => {
           const statusCode = err?.statusCode || err?.code || status + ''
-          reject(new StorageApiError(_getErrorMessage(err), status, statusCode, namespace))
+          reject(
+            new StorageApiError(_getErrorMessage(err), status, statusCode, namespace, err?.code)
+          )
         }
       )
       .catch(() => {
--- a/packages/core/storage-js/test/common/errors.test.ts
+++ b/packages/core/storage-js/test/common/errors.test.ts
@@ -79,6 +79,28 @@ describe('Common Errors', () => {
         statusCode: 'InternalError',
       })
     })
+
+    it('should expose the service error code when provided', () => {
+      const error = new StorageApiError('Object not found', 404, '404', 'storage', 'NoSuchKey')
+      expect(error.code).toBe('NoSuchKey')
+      expect(error.statusCode).toBe('404')
+    })
+
+    it('should leave code undefined when not provided', () => {
+      const error = new StorageApiError('API error', 404, 'NotFound')
+      expect(error.code).toBeUndefined()
+    })
+
+    it('should include code in JSON when present', () => {
+      const error = new StorageApiError('Access denied', 403, '403', 'storage', 'AccessDenied')
+      expect(error.toJSON()).toEqual({
+        name: 'StorageApiError',
+        message: 'Access denied',
+        status: 403,
+        statusCode: '403',
+        code: 'AccessDenied',
+      })
+    })
   })
 
   describe('StorageUnknownError', () => {
--- a/packages/core/storage-js/test/common/fetch.test.ts
+++ b/packages/core/storage-js/test/common/fetch.test.ts
@@ -99,7 +99,8 @@ describe('Common Fetch', () => {
       })
 
       it('should expose the service error code from the response body', async () => {
-        const errorResponse = { code: 'NoSuchKey', message: 'Object not found' }
+        // Mirror a real Storage payload, which sends both `statusCode` and `code`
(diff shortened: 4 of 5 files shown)


mandarini (changes requested):
Hi @thribhuvan003, thank you so much for contributing to Supabase! 💚 

This is exactly the right fix. You landed on the correct field: the Storage service always sends the documented code (`NoSuchKey`, `AccessDenied`, and so on) as `code` in the response body, and the SDK was quietly dropping it because `statusCode` took precedence. Passing `err.code` through onto a dedicated `code` property, and including it in `toJSON()`, is precisely what we want.

Two small things before we merge, neither blocking:

1. In the new fetch test the mock body is `{ code: 'NoSuchKey', message: 'Object not found' }` with no `statusCode`. That happens to exercise the legacy fallback where `statusCode` also becomes `'NoSuchKey'`, which is not what a real server sends. The real payload always includes both, like `{ statusCode: '404', code: 'NoSuchKey', message: '...' }`. Could you tweak the fixture to include `statusCode` so it mirrors reality? The assertion on `code` stays the same.

2. Optional, but nice to have: a small integration test that downloads a missing object and asserts `error.code`, so we lock in that the server really does send what we are reading. Happy to merge without it if you would rather keep the PR tight.

Thank you again for the clean, focused patch and for the tests that actually fail without the change. Contributions like this are what keep the SDK healthy.

thribhuvan003:
thanks @mandarini, that means a lot 💚

done on (1) — the fetch fixture now sends `{ statusCode: '404', code: 'NoSuchKey', message: '...' }` like the real server, and asserts both `statusCode: '404'` and `code: 'NoSuchKey'` so it no longer leans on the legacy fallback. all unit tests still green.

on (2): i'd like to add it, but i can't run the storage docker stack in my dev env to verify the exact `code` a real missing-object download returns, and i didn't want to push an integration assertion i haven't actually seen pass. happy to add one if you're fine trusting CI to validate it, or if you can confirm the code a 404 download surfaces — otherwise glad to keep it tight per your offer. either way, thanks for the super clear review.

mandarini:
Really appreciate the follow-up, and you asked exactly the right question in (2): *what `code` does a real 404 download surface?* I went into the Storage server to confirm, and the answer turned out to be surprising: **none does.** That also means the fixture in (1) doesn't quite mirror the real server. Τhe actual 4xx body is `{ statusCode: '404', error: 'not_found', message: 'Object not found' }`, with no `code` field.

Here's why: the server generates `NoSuchKey` internally ([codes.ts](https://github.com/supabase/storage/blob/308ac4820d6f8c48d6b77b0eece5bc3f7951c63b/src/internal/errors/codes.ts#L129-L137)) and even sets it in `render()` ([storage-error.ts](https://github.com/supabase/storage/blob/308ac4820d6f8c48d6b77b0eece5bc3f7951c63b/src/internal/errors/storage-error.ts#L90-L97)) — but every object route serializes its `4xx` body through a Fastify response schema, [`errorSchema`](https://github.com/supabase/storage/blob/308ac4820d6f8c48d6b77b0eece5bc3f7951c63b/src/http/schemas/error.ts#L1-L10), that whitelists only `statusCode`/`error`/`message`. Fastify drops anything else, so `code` never reaches the client. The server's own tests confirm it: a 404 `NoSuchKey` asserts an exact body with no `code` ([object.test.ts#L246](https://github.com/supabase/storage/blob/308ac4820d6f8c48d6b77b0eece5bc3f7951c63b/src/test/object.test.ts#L246-L262)); a 500 keeps `code` only because 5xx has no schema to strip it.

Your instinct not to push an integration assertion you hadn't seen pass was correct, it would have failed, and that's what surfaced this. The client patch is the right shape, but it can't deliver until the server includes `code` in `errorSchema`. I'll follow up on the server side to get `code` emitted; once it is, your change is exactly what we pair with it. Genuinely great catch! It spans both repos.

mandarini:
@thribhuvan003 good news btw, storage team added it on their timeline! so once they make a PR i will link it here, and once PR is merged and released, we can move on with your PR!

thribhuvan003:
amazing, thank you 🙌 no rush at all on my end — i'll watch for the storage team's PR and align mine once it lands.

ferhatelmas:
https://github.com/supabase/storage/pull/1256

mandarini:
@thribhuvan003 this is now shipped on the storage server!

thribhuvan003:
Thanks @mandarini — great news, and thank you for driving the server-side change!

I've rebased onto the latest master, and the client already reads the newly shipped `code` field, so both sides now line up.

One thing on the SDK Compliance check: it flags `StorageApiError.code` and the inherited `StorageVectorsApiError.code` as new public symbols, but I couldn't find an existing Storage error-code capability to register them under. Could you let me know where you'd prefer them mapped? Happy to update the compliance file accordingly.

mandarini (approved):

grdsdev (approved):

PR #2538, opened 2026-07-19
fix(storage): expose service error codes

### LINKED ISSUE #2536: StorageApiError does not expose documented Storage error codes (error)

7ttp:
## TL;DR

Expose Storage service error codes on `StorageError`

pass through the canonical `code` field while supporting the legacy `error` response field
 used by older and self-hosted versions (BC)

## Ref

- Closes #2536
- https://supabase.com/docs/guides/storage/debugging/error-codes

### DIFF
--- a/packages/core/storage-js/src/lib/common/errors.ts
+++ b/packages/core/storage-js/src/lib/common/errors.ts
@@ -13,6 +13,8 @@ export class StorageError extends Error {
   protected namespace: ErrorNamespace
   status: number | undefined
   statusCode: string | undefined
+  code?: string
+  error?: string
 
   constructor(
     message: string,
@@ -32,12 +34,16 @@ export class StorageError extends Error {
     message: string
     status: number | undefined
     statusCode: string | undefined
+    code?: string
+    error?: string
   } {
     return {
       name: this.name,
       message: this.message,
       status: this.status,
       statusCode: this.statusCode,
+      ...(this.code !== undefined ? { code: this.code } : {}),
+      ...(this.error !== undefined ? { error: this.error } : {}),
     }
   }
 }
@@ -63,12 +69,16 @@ export class StorageApiError extends StorageError {
     message: string,
     status: number,
     statusCode: string,
-    namespace: ErrorNamespace = 'storage'
+    namespace: ErrorNamespace = 'storage',
+    code?: string,
+    error?: string
   ) {
     super(message, namespace, status, statusCode)
     this.name = namespace === 'vectors' ? 'StorageVectorsApiError' : 'StorageApiError'
     this.status = status
     this.statusCode = statusCode
+    if (code !== undefined) this.code = code
+    if (error !== undefined) this.error = error
   }
 
   toJSON(): {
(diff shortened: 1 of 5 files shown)


mandarini:
Helloooo @7ttp! :D You diagnosed this perfectly: the service error code is present in the response body and the SDK was throwing it away, and I really like that you were the one to add an integration test that downloads a missing object!

I think I'm going to go with the other PR for this issue (#2537) and close this one, and I want to explain why so it is useful feedback rather than just a decision.

The main reason is the `error` field. This PR exposes both `code` and `error`, but on the server those two are not the same thing. For a missing object the body is actually:

```json
{ "statusCode": "404", "code": "NoSuchKey", "error": "not_found", "message": "Object not found" }
```

The documented codes live on `code` (`NoSuchKey`). The `error` field is an older, inconsistent string (`not_found` here, `invalid_mime_type` or `Internal` in other cases). Exposing `error` as a public field would hand people a second value that looks authoritative but disagrees with the docs, which is the very confusion the original issue ran into. Related to that, the `code = err.code || err.error` fallback can end up putting a non-code like `not_found` into the field we tell people holds canonical codes, so we would rather not have that fallback either.

None of this is a knock on your work. If you strip it down to exposing just `code`, your PR essentially becomes #2537 with the added integration test, which is a great combination.

Thank you again for taking the time on this!! :D

PR #2553, opened 2026-07-23
fix: PKCE code challenge double-encoding and storage path URI encoding

PedroHenrique0713:
## Summary

Two encoding fixes in supabase-js.

### 1. PKCE code challenge double-encoding (auth-js)
`URLSearchParams` already applies `encodeURIComponent` to its values internally. The explicit `encodeURIComponent()` call on the code challenge values caused **double-encoding** — `%2B` became `%252B`. This broke PKCE verification on the auth server, silently failing OAuth login flows for all apps using PKCE (the default flow type since v2).

### 2. Storage path segments not URI-encoded (storage-js)
`_getFinalPath` interpolated raw path segments into URLs without `encodeURIComponent`. File names containing `#`, `?`, `&`, or spaces produced broken URLs where the fragment/query boundary truncated the URL before reaching the storage server. Affected every storage operation (`upload`, `download`, `getPublicUrl`, `move`, `copy`, `remove`).

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -5595,8 +5595,8 @@ export default class GoTrueClient {
       )
 
       const flowParams = new URLSearchParams({
-        code_challenge: `${encodeURIComponent(codeChallenge)}`,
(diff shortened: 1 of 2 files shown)


PedroHenrique0713:

PR #2554, opened 2026-07-23
fix(storage): wrap non-StorageError exceptions in StorageUnknownError

PedroHenrique0713:
## Summary

`handleOperation`'s documented contract is to always return `{ data, error }` tuples. However, when the operation threw a non-StorageError exception (JSON parse failure, TypeError from missing fetch, etc.), the error was re-thrown instead of being wrapped.

**Fix:** Wrap non-StorageError exceptions in `StorageUnknownError` and return `{ data: null, error }`.


PedroHenrique0713:

PR #2555, opened 2026-07-23
fix(auth): getUserIdentities null guard and validateUUID error envelope

PedroHenrique0713:
Fix 2 auth-js bugs: getUserIdentities crashes when no session (null user → TypeError), getUserById validateUUID throws outside try/catch bypassing { data, error } envelope.


PedroHenrique0713:

PR #2559, opened 2026-07-24
fix(auth): downgrade stale refresh token console noise

### LINKED ISSUE #2558: Accept external log handler (as an alternative to direct console.error calls)
### Description

Currently Supabase calls `console.error` directly (for instance [here](https://github.com/supabase/supabase-js/blob/485695ff777be75e1713586c40492d663cd9e3d2/packages/core/auth-js/src/GoTrueClient.ts#L4299)). In my system we are monitoring stdout and stderr and treat every error as a bug and alert for it. We often see errors such as "AuthApiError: Invalid Refresh Token: Refresh Token Not Found" which are, if I understand correctly, expected or are at least warnings, not errors. Accepting a logger (in whatever form you choose) as an optional alternative to `console` calls would help users decide what should be logged to the console (and potentially trigger their alerting services) and what's not.

Happy to hear if I got this wrong and there is another way to approach this.
Thanks

### Suggested solution

 Accepting a logger (in whatever form you choose) as an optional alternative to `console` calls (`console.error`, etc.).

### Alternative

AuthSessionMissingError is already downgraded to console.warn (#2214). The similar case in _recoverAndRefresh was fixed in #2428 / #2416, but the `@supabase/ssr` INITIAL_SESSION path still logs AuthApiError at error level. A quick fix would be to narrow here too: also warn/_debug for refresh_token_not_found, refresh_token_already_used, session_expired, session_not_found in _emitInitialSession.

### Additional context

Error surfaced when using SSR (Next.js middleware + @supabase/ssr createServerClient).

### Validations

7ttp:
## TL;DR 

`_emitInitialSession` already warns instead of erroring for a missing session or a `retryable `fetch failure. 
a dead refresh token
`(refresh_token_not_found, refresh_token_already_used, session_expired)` still hit raw `console.error` even 
though it's the same kind of expected outcome on page load....

this just widens that same check to cover those three codes too. 
## ref: 
- closes #2558
- somewhat the same idea: 
#2214 
 #2428

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -4290,10 +4290,18 @@ export default class GoTrueClient {
       } catch (err) {
         await this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
         this._debug('INITIAL_SESSION', 'callback id', id, 'error', err)
-        if (isAuthSessionMissingError(err) || isAuthRetryableFetchError(err)) {
-          // A missing session or a transient/aborted network failure (e.g. a
-          // superseded page navigation cancelling the in-flight request) is
-          // not an application error — warn rather than surface it raw.
+        if (
+          isAuthSessionMissingError(err) ||
+          isAuthRetryableFetchError(err) ||
+          (isAuthApiError(err) &&
+            (err.code === 'refresh_token_not_found' ||
+              err.code === 'refresh_token_already_used' ||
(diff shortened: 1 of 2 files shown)


mandarini:
@7ttp thanks for this PR,let's wait for [my other PR](https://github.com/supabase/supabase-js/pull/2561) to be merged and you rebase

mandarini (approved):

mandarini:
let's wait for this now https://github.com/supabase/supabase-js/pull/2564

PR #2566, opened 2026-07-27
fix(realtime): preserve presence refs

### LINKED ISSUE #2565: Presence: payload building mutates shared metas (deletes phx_ref), leaves never match, departed members stay online forever
# Bug report

## Describe the bug

Since the Phoenix adapter refactor, presence "leave" diffs can fail to remove a departed member from `presenceState()`: the member stays online forever for other clients, and no later `sync` ever corrects it.

Root cause: `transformState()` in `packages/core/realtime-js/src/phoenix/presenceAdapter.ts` (still present on `main`) **mutates the meta objects in place** while building the `join`/`leave` event payloads:

```ts
function transformState(presences: PresenceState) {
  return presences.metas.map((presence) => {
    presence['presence_ref'] = presence['phx_ref']

    delete presence['phx_ref']
    delete presence['phx_ref_prev']

    return presence
  }) as RealtimePresenceType[]
}
```

Those meta objects are **shared with the live Phoenix presence state**, in two ways:

1. In `Presence.syncDiff` (joins), the surviving old metas are `unshift`ed into `state[key].metas` **by reference** (`curMetas` are the same objects as `currentPresence.metas`). The adapter's `onJoin` callback then calls `parseCurrentPresences(currentPresence)` → `transformState`, which deletes `phx_ref` on those live objects.
2. In the leaves branch, `onLeave(key, currentPresence, leftPresence)` receives `currentPresence` which *is* `state[key]`; `parseCurrentPresences` strips `phx_ref` from the remaining metas.

After either path runs, the corrupted meta has `phx_ref: undefined`. The next leave diff filters with `refsToRemove.indexOf(p.phx_ref) < 0`, never matches, `metas` never reaches length 0, and the key is never deleted: a ghost presence.

## To Reproduce

Two users, presence key = user id, one of them tracked from two clients (two "devices", same key):

1. A, B1 and B2 each `channel('presence:x', { config: { presence: { key: <user id> } } })`, subscribe, `track({})`.
2. B1 and B2 leave (`untrack()` or `removeChannel()`).
3. On A, `presenceState()` intermittently still contains B's key, forever. Whether it reproduces depends on diff delivery order (whether A's cached metas got mutated by a join/leave payload first), so run it a few times; we saw roughly 1 failure in 4.

Instrumented state on A at failure time (B's key is `6794`; note the `presence_ref: null` meta that no leave can ever match):

```
[A] join  key=6794 refs=GMYj8_YMd76EpljQ        state {"6794":["GMYj8_YMd76EpljQ"]}
[A] join  key=6794 refs=GMYj9IQosf88gG3I        state {"6794":[null,"GMYj9IQosf88gG3I"]}   <- old meta lost its ref
[A] leave key=6794 refs=GMYj9IQosf88gG3I        state {"6794":[null]}
[A] leave key=6794 refs=GMYj8_YMd76EpljQ        state {"6794":[null]}                      <- ghost, permanent
```

## Expected behavior

Building event payloads must not mutate the shared presence state; after all metas of a key leave, the key disappears from `presenceState()`.

## Fix

Make `transformState` non-mutating (this is what we ship via patch-package, it fully fixes the repro, 15/15 runs green vs ~25% failures before):

```js
function transformState(presences) {
    return presences.metas.map((presence) => {
        const copy = Object.assign({}, presence);
        copy['presence_ref'] = copy['phx_ref'];
        delete copy['phx_ref'];
        delete copy['phx_ref_prev'];
        return copy;
    });
}
```

## System information

- OS: macOS 15 / ubuntu-latest (GitHub Actions), Node 26
- @supabase/supabase-js 2.110.1 (realtime-js 2.110.1); the mutating `transformState` is still present in 2.110.8

7ttp:
## TL;DR

Build presence event payloads from copies so live presence refs remain intact and
 departed members are removed correctly.

## ref:

- closes #2565

### DIFF
--- a/packages/core/realtime-js/src/phoenix/presenceAdapter.ts
+++ b/packages/core/realtime-js/src/phoenix/presenceAdapter.ts
@@ -94,12 +94,13 @@ export default class PresenceAdapter {
 
 function transformState(presences: PresenceState) {
   return presences.metas.map((presence) => {
-    presence['presence_ref'] = presence['phx_ref']
+    const transformedPresence = { ...presence }
+    transformedPresence['presence_ref'] = transformedPresence['phx_ref']
 
-    delete presence['phx_ref']
-    delete presence['phx_ref_prev']
+    delete transformedPresence['phx_ref']
+    delete transformedPresence['phx_ref_prev']
 
-    return presence
+    return transformedPresence
   }) as RealtimePresenceType[]
 }
 
--- a/packages/core/realtime-js/src/phoenix/presenceAdapter.ts
+++ b/packages/core/realtime-js/src/phoenix/presenceAdapter.ts
@@ -94,7 +94,9 @@ export default class PresenceAdapter {
 
 function transformState(presences: PresenceState) {
   return presences.metas.map((presence) => {
-    const transformedPresence = { ...presence }
+    // Object spread compiles to Object.assign for ES2017, which treats __proto__ as a setter.
+    const descriptors = Object.getOwnPropertyDescriptors(presence)
+    const transformedPresence = Object.defineProperties({}, descriptors) as typeof presence
     transformedPresence['presence_ref'] = transformedPresence['phx_ref']
 
     delete transformedPresence['phx_ref']
--- a/packages/core/realtime-js/test/phoenix/presenceAdapter.test.ts
+++ b/packages/core/realtime-js/test/phoenix/presenceAdapter.test.ts
@@ -111,6 +111,47 @@ describe('transformState', () => {
   })
 })
 
+describe('presence diff payloads', () => {
+  test('removes a presence key after all of its metas leave', () => {
+    const key = 'user-123'
+    // Each server message is decoded into fresh meta objects.
+    const meta = (phx_ref: string) => ({ phx_ref, user_id: key })
+    let state: PresenceStates = {
+      [key]: { metas: [meta('device-1')] },
+    }
+
+    state = Presence.syncDiff(
+      state,
+      {
+        joins: { [key]: { metas: [meta('device-2')] } },
+        leaves: {},
+      },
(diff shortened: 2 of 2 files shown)


mandarini (changes requested):
One small thing before I merge. Can we please add a test for the presence update path? Your test covers two devices under one key. The more common trigger is a single client calling `track()` again with a changed payload: the server responds with `Phoenix.Presence.update` (`presence_handler.ex` L162-163), which sends a join and a leave for the same key in a single diff. That shape duplicates the member immediately with the stale payload still visible, and then both metas ghost on disconnect. It is a different symptom from the one you assert on, so it needs its own case:

```ts
test('collapses a presence update instead of duplicating the member', () => {
  const key = 'user-123'
  let state: PresenceStates = { [key]: { metas: [{ phx_ref: 'r1', status: 'online' }] } }

  // track() with a changed payload: the server sends Presence.update as join + leave
  state = Presence.syncDiff(
    state,
    {
      joins: { [key]: { metas: [{ phx_ref: 'r2', phx_ref_prev: 'r1', status: 'away' }] } },
      leaves: { [key]: { metas: [{ phx_ref: 'r1', status: 'online' }] } },
    },
    PresenceAdapter.onJoinPayload,
    PresenceAdapter.onLeavePayload
  )
  expect(state[key].metas).toHaveLength(1)

  state = Presence.syncDiff(
    state,
    { joins: {}, leaves: { [key]: { metas: [{ phx_ref: 'r2', status: 'away' }] } } },
    PresenceAdapter.onJoinPayload,
    PresenceAdapter.onLeavePayload
  )
  expect(state).not.toHaveProperty(key)
})
```

`toHaveLength(1)` is 2 without your change and 1 with it, so it earns its place.

mandarini (approved):

PR #2568, opened 2026-07-27
fix(auth): fail token refresh fast when offline and resume on reconnect

kklem0:
### What changed?

Token refreshing is now aware of the browser's affirmative offline signal:

1. `_refreshAccessToken` always sends the first attempt through the configured `fetch`, but stops the retry/backoff loop after an actual failure while `navigator.onLine === false`, and only where the signal is binding: the default transport (no custom `fetch` was supplied; supabase-js forwards `global.fetch` verbatim, undefined when not configured) toward a non-loopback auth URL (`localhost`, `*.localhost`, their fully qualified trailing-dot forms, `127.0.0.0/8` and `[::1]` are exempt, mirroring the loopback rules of the potentially trustworthy origin definition; the check parses the configured auth endpoint, not the page origin, so WebView apps served from a synthetic localhost origin keep suppression for remote Supabase URLs). Custom transports (request interception, service-worker-backed or native-bridged fetch) are not bound by `onLine`, and loopback endpoints stay reachable without it, so both keep the full retry loop in every case; what no longer happens on the default transport toward remote hosts is the exponential backoff loop of doomed retries (about 25 seconds of fetches and sleeps per refresh in browsers, longer on flaky mobile networks where each fetch can hang).
2. A new helper `isProvablyOffline()` encapsulates the check. Per the HTML spec, `onLine === false` means the user agent will not contact the network for remote requests, so retrying through the user agent's own transport cannot succeed until connectivity returns. The `true` value can be wrong in both directions and is never trusted or used for decisions, and the signal is only used to stop retrying default-transport requests after a real failure, never to skip a request and never for custom transports.
3. The client registers an `online` event listener on the global returned by `getOnlineEventTarget()`: `window` in documents, the worker global scope in Web Workers (both fire `online` and expose the boolean `navigator.onLine` the retry gate reads), nothing in environments without that signal. On reconnect it clears the refresh-failure cooldown and runs an auto-refresh tick, but only while the ticker is active, keeping the foreground-tab-only refresh discipline. The listener is removed in `dispose()`.
4. Disposal races are handled: `_initialize` re-checks a `_disposed` flag around its lifecycle setup (dispose can interleave with the awaited visibility handler), and `_startAutoRefresh` creates no timers when dispose ran during its internal await, so React Strict Mode / HMR teardown cannot leak listeners or restart the ticker.

Everything else engages unchanged: proactive-preserve keeps the session in storage and in `getSession()` results, the failure cooldown caches the failed outcome, and callers see the transport's real error.

Apps used on devices that go offline (mobile web and hybrid apps in areas with poor coverage) currently hit two problems:

- With the stored access token inside the expiry margin, every `getSession()` call triggers a real refresh attempt. While offline this blocks callers for up to the full tick duration in browsers and effectively unbounded time on flaky networks, so auth-dependent UI stalls or appears logged out at startup.
- After connectivity returns, nothing refreshes until the failure cooldown expires and the next auto-refresh tick fires, which can add up to another 90 seconds.

With this change, offline behavior becomes deterministic and immediate:

- `getSession()` resolves quickly offline: a single transport probe with no backoff sleeps, returning the preserved session while the access token is inside its real expiry, or `{ session: null, error: AuthRetryableFetchError }` once it has truly expired, letting apps distinguish "cannot verify right now" from "signed out" (the session stays in storage).
- Refreshing resumes the moment connectivity returns, in documents and in Web Workers.

## 📸 Screenshots/Examples

N/A

## 🔄 Breaking changes

React Native, Node.js and Deno have no boolean `navigator.onLine`, so `isProvablyOffline()` returns false there and no listener is registered; behavior in those environments is identical. Transports that succeed while `onLine === false` keep working because the first attempt is always sent. In browsers and workers the only observable differences are a faster failure while provably offline (single probe, same error surface) and an immediate refresh on reconnect.

## 📋 Checklist

## 📝 Additional notes

Test coverage added:

- `GoTrueClient.test.ts`, "offline fail-fast (provably offline)" suite: single default-transport probe with no backoff while offline against a remote auth URL (observed via a global-fetch override so no custom `fetch` is configured; cooldown entered, storage preserved), preserved session via `getSession()` inside the margin, null plus retryable error once fully expired, a regression test proving a navigator without a boolean `onLine` (React Native) keeps the full retry loop, regression tests proving a custom transport is never suppressed (succeeds on the first attempt, and fails once then succeeds on the retried second attempt, both while `onLine === false`), a regression test proving a loopback auth URL keeps the retry loop on the default transport (fails once, succeeds on the second attempt), and worker-style global registration with cooldown clearing and removal on dispose.
- `GoTrueClient.browser.test.ts` (jsdom): `online` clears the cooldown, the tick is gated on an active ticker, `dispose()` removes the listener, dispose during pending initialization registers no lifecycle listeners, dispose interleaving with visibility setup registers no online listener and restarts no ticker, and `_startAutoRefresh` resumed across dispose creates no timers (the interleaving tests fail without the fixes).
- `helpers.test.ts`: `isProvablyOffline`, `getOnlineEventTarget` and `isLoopbackHost` truth tables.

All existing refresh lifecycle tests pass unchanged. Documentation is TSDoc on the new helpers and methods; no separate docs apply.

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -300,6 +300,14 @@ export default class GoTrueClient {
    * restart background work on a client that was already torn down.
    */
   protected _disposed = false
+  /**
+   * True when no custom `fetch` was supplied and requests go through the
+   * environment's own transport. Offline retry suppression in
+   * `_refreshAccessToken` applies only in that case: `navigator.onLine`
+   * describes the user agent's network stack, not custom transports, which
+   * may fail transiently and succeed on a later attempt regardless of it.
+   */
+  protected readonly usesDefaultFetch: boolean
   protected refreshingDeferred: Deferred<CallRefreshTokenResult> | null = null
   /**
    * Cache of the most recent refresh failure, keyed by the refresh token
@@ -445,6 +453,7 @@ export default class GoTrueClient {
     this.url = settings.url
     this.headers = settings.headers
     this.fetch = resolveFetch(settings.fetch)
+    this.usesDefaultFetch = settings.fetch == null
     this.detectSessionInUrl = settings.detectSessionInUrl
     this.flowType = settings.flowType
     this.hasCustomAuthorizationHeader = settings.hasCustomAuthorizationHeader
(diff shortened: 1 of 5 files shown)


kklem0:
@coderabbitai resume

kklem0:
@mandarini thoughts? :)

PR #2573, opened 2026-07-28
feat(auth): add maxAutoRefreshFailures option

### LINKED ISSUE #1680: Always retry fetching
# Bug report

## Describe the bug

It keeps trying to refresh the token when the server is unavailable. I can't get it to stop and show an error.

https://github.com/supabase/auth-js/blob/daa266949b336d2e78f2a7b9c9837b70abeab7a6/src/GoTrueClient.ts#L1836-L1859

## To Reproduce

1. Login to a working project
2. Change supabase url to `https://foo.example.com`. simulate a server error

## Expected behavior

Stop and throw an error after a certain number of attempts.

## System information

- OS: Windows
- Browser (if applies): chrome
- Version of supabase-js: 2.49.1
- Version of Node.js: Bun 1.2.5

7vignesh:
### What changed?

Added an opt-in `maxAutoRefreshFailures` option to the auth client that stops the auto-refresh ticker after N consecutive failed ticks and emits a new `TOKEN_REFRESH_FAILED` event via `onAuthStateChange`.

When the auth server is permanently unreachable (wrong URL, server down, DNS failure), the auto-refresh ticker retries every 30 seconds forever with no way for the app to know it's failing. Users reported pages hanging for 14-21 seconds on load due to retry backoff, with no signal to react to.

This was raised in #1680 and acknowledged by @mandarini as a valid feature request for a configurable "stop after N failed ticks and emit an error event" option that is opt-in.

### How it works

- New option `maxAutoRefreshFailures` (defaults to `0` which means retry forever, preserving existing behavior)
- A counter tracks consecutive failed auto-refresh ticks
- Counter resets to 0 on any successful token refresh
- When the counter reaches the configured limit, the client stops the auto-refresh ticker and emits `TOKEN_REFRESH_FAILED` through `onAuthStateChange`

### Usage

```ts
const supabase = createClient(URL, KEY, {
  auth: { maxAutoRefreshFailures: 5 }
})

supabase.auth.onAuthStateChange((event) => {
  if (event === 'TOKEN_REFRESH_FAILED') {
    // Server unreachable after 5 consecutive failures (~2.5 min)
    // Redirect to login or show a banner
  }
})
```

Closes #1680

## Breaking changes

The option defaults to 0 (no limit), so existing behavior is completely unchanged. The new `TOKEN_REFRESH_FAILED` event is additive and only fires when the option is explicitly configured.

## Checklist

## Testing

- 5 new unit tests covering: indefinite retry (default), stop after max failures, counter reset on success, no false positive events, thrown error handling
- All 170 existing auth-js tests pass
- Build passes for both auth-js and supabase-js

## Additional notes

- Both the legacy lock path and the default lockless path are covered
- The failure counter is also reset inside `_callRefreshToken` on success, so any refresh path (manual or automatic) resets it
- Zero runtime impact when the option is not set (the counter increments but the check short-circuits on `maxAutoRefreshFailures === 0`)

### DIFF
--- a/packages/core/auth-js/src/GoTrueClient.ts
+++ b/packages/core/auth-js/src/GoTrueClient.ts
@@ -207,6 +207,7 @@ const DEFAULT_OPTIONS: Omit<
   throwOnError: false,
   lockAcquireTimeout: 5000, // 5 seconds. Only used when a custom `lock` is supplied. TODO(v3): remove.
   skipAutoInitialize: false,
+  maxAutoRefreshFailures: 0,
   experimental: {},
 }
 
@@ -372,6 +373,16 @@ export default class GoTrueClient {
    */
   protected lockAcquireTimeout: number
   /**
+   * Maximum consecutive auto-refresh tick failures before the client stops
+   * retrying and emits `TOKEN_REFRESH_FAILED`. 0 means no limit (default).
+   */
+  protected maxAutoRefreshFailures: number
+  /**
+   * Tracks consecutive auto-refresh tick failures. Reset to 0 on any
+   * successful token refresh.
+   */
+  protected autoRefreshFailureCount: number = 0
+  /**
    * Opt-in flags for experimental features. Defaults to an empty object.
    * See `GoTrueClientOptions.experimental`.
    */
@@ -449,6 +460,7 @@ export default class GoTrueClient {
     // (including supabase-js tests) read it off the client to verify option
     // flow-through.
     this.lockAcquireTimeout = settings.lockAcquireTimeout
+    this.maxAutoRefreshFailures = settings.maxAutoRefreshFailures
 
     // TODO(v3): remove. Legacy opt-in path preserved for backwards
     // compatibility with callers passing a custom `lock` (typically React
@@ -5188,6 +5200,16 @@ export default class GoTrueClient {
     // _saveSession is always called whenever a new session has been acquired
     // so we can safely suppress the warning returned by future getSession calls
     this.suppressGetSessionWarning = true
+
+    // A new session means auth is working — reset the failure counter.
+    this.autoRefreshFailureCount = 0
+
+    // In non-browser environments (React Native, Node) nothing restarts the
+    // ticker after the limit was hit. Re-enable it here so that a successful
+    // sign-in or setSession recovers auto-refresh.
+    if (this.autoRefreshToken && this.autoRefreshTicker === null && !isBrowser()) {
+      void this._startAutoRefresh()
+    }
     // Create a shallow copy to work with, to avoid mutating the original session object if it's used elsewhere
     const sessionToProcess = { ...session }
 
@@ -5233,6 +5255,7 @@ export default class GoTrueClient {
     // The session is gone — no point holding on to a cached refresh failure
     // for a token that no longer exists. Synchronous, before any `await`.
     this.lastRefreshFailure = null
+    this.autoRefreshFailureCount = 0
 
     this.suppressGetSessionWarning = false
 
@@ -5477,10 +5500,19 @@ export default class GoTrueClient {
               return await this._useSession(async (result) => {
                 const {
                   data: { session },
+                  error,
                 } = result
 
                 if (!session || !session.refresh_token || !session.expires_at) {
                   this._debug('#_autoRefreshTokenTick()', 'no session')
+                  if (error && isAuthRetryableFetchError(error)) {
+                    this.autoRefreshFailureCount++
+                    this._debug(
+                      '#_autoRefreshTokenTick()',
+                      `session load failed, consecutive failures: ${this.autoRefreshFailureCount}`
+                    )
+                    await this._handleAutoRefreshFailure(session ?? null)
+                  }
                   return
                 }
 
@@ -5494,14 +5526,30 @@ export default class GoTrueClient {
                 )
 
                 if (expiresInTicks <= AUTO_REFRESH_TICK_THRESHOLD) {
-                  await this._callRefreshToken(session.refresh_token)
+                  const refreshResult = await this._callRefreshToken(session.refresh_token)
+                  if (refreshResult.error) {
+                    if (isAuthRetryableFetchError(refreshResult.error)) {
+                      this.autoRefreshFailureCount++
+                      this._debug(
+                        '#_autoRefreshTokenTick()',
+                        `refresh failed, consecutive failures: ${this.autoRefreshFailureCount}`
+                      )
+                      await this._handleAutoRefreshFailure(session ?? null)
+                    }
+                  }
                 }
               })
             } catch (e) {
+              this.autoRefreshFailureCount++
+              this._debug(
+                '#_autoRefreshTokenTick()',
+                `tick threw, consecutive failures: ${this.autoRefreshFailureCount}`
+              )
               console.error(
                 'Auto refresh tick failed with error. This is likely a transient error.',
                 e
               )
+              await this._handleAutoRefreshFailure()
             }
           } finally {
             this._debug('#_autoRefreshTokenTick()', 'end')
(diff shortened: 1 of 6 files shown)


7vignesh:
@mandarini pls review

mandarini (changes requested):
Hi @7vignesh, thank you so much for contributing to Supabase! This is a well-shaped implementation of exactly the opt-in behavior we discussed in #1680! There are a few things that need to change before we can merge it.

1. **The option never reaches auth-js when set through `createClient`.** `SupabaseClientOptions['auth']` is an explicit field list (`packages/core/supabase-js/src/lib/types.ts`) and `_initSupabaseAuthClient` destructures an allow-list (`packages/core/supabase-js/src/SupabaseClient.ts`), so the usage example in the PR description is a type error for TypeScript users and a silent no-op for JavaScript users. Please add `maxAutoRefreshFailures` to both, plus a small supabase-js test that reads the option off the client to verify flow-through (we do the same for `lockAcquireTimeout`).

2. **The counter goes stale because it only resets on a successful refresh.** If a user hits the limit and later signs back in, the count is still at max, so one transient blip will stop auto-refresh and emit `TOKEN_REFRESH_FAILED` after a single failure instead of N. Related: `_onVisibilityChanged` restarts the ticker on every tab focus, so during an outage the event re-fires on each refocus. Resetting the counter in `_startAutoRefresh()` fixes both. It would also be good to reflect the "resumes on tab focus" reality in the JSDoc, since the ticker stop is not permanent in browsers.

3. **The error path can re-enter itself.** `_notifyAllSubscribers` re-throws subscriber errors, so if an app callback throws while handling `TOKEN_REFRESH_FAILED`, the error lands in the tick's catch block, which increments the counter and calls `_handleAutoRefreshFailure` again: the event fires twice, and a second subscriber throw becomes an unhandled rejection out of the `setInterval` tick. Please make sure the failure handler cannot run twice for the same tick.

4. **Every error currently counts toward the limit**, including non-retryable 4xx responses (revoked token, reuse detection, rate limits) that already have their own handling in `_callRefreshToken` (sign-out or session preserve). Since the feature targets an unreachable server, please gate the counting on `isAuthRetryableFetchError(error)` so terminal auth states and rate limits do not shut down auto-refresh. Also worth deciding: the event fires with a `null` session even when a still-valid session was preserved, and apps commonly treat a null session as signed out, so consider passing the current session or calling this out in the JSDoc.

On tests: the suite mocks both `_useSession` and `_callRefreshToken`, so it mostly exercises the plumbing between mocks. Could you add at least one test that goes through the real session load path (a fetch-level mock is fine)? None of the tests pass a custom `lock` either, so the legacy path is not actually covered yet despite the PR notes.

One thing you do not need to solve: adding `TOKEN_REFRESH_FAILED` to `AuthChangeEvent` has cross-SDK naming implications that we will coordinate on our side.

Really appreciate the care that went into this, the write-up alone made it a pleasure to review. Thank you for helping make Supabase better, contributions like this are what keep these libraries moving.

7vignesh:
Hi @mandarini, thank you for the thorough review! All 4 points addressed in the latest push:

**1. Option flow-through via createClient**
Added `maxAutoRefreshFailures` to `SupabaseClientOptions['auth']` in `packages/core/supabase-js/src/lib/types.ts`, destructured and forwarded it in `_initSupabaseAuthClient`, and added two supabase-js unit tests verifying the value reaches the auth client (same pattern as `lockAcquireTimeout`).

**2. Counter goes stale**
Added `this.autoRefreshFailureCount = 0` at the top of `_startAutoRefresh()`. This resets the budget on sign-in, tab refocus (`_onVisibilityChanged` calls `_startAutoRefresh`), or any explicit `startAutoRefresh()` call. Also updated the JSDoc to note that in browsers the stop is not permanent since visibility changes restart the ticker.

**3. Re-entry guard**
Added `this.autoRefreshTicker !== null` check to `_handleAutoRefreshFailure()` so it short-circuits if already stopped. Wrapped `_notifyAllSubscribers` in a try/catch so subscriber errors are swallowed with a console.error instead of propagating back into the tick's catch block. Added a test that verifies TOKEN_REFRESH_FAILED fires exactly once even when a subscriber throws.

**4. Only count retryable errors**
Gated all failure-count increments behind `isAuthRetryableFetchError(error)`. Non-retryable 4xx responses (revoked token, rate limits) no longer count toward the limit. Added a dedicated test confirming non-retryable errors don't increment the counter.

**Tests:** 11 tests now covering default behavior, max limit, counter reset, non-retryable errors, thrown errors, retryable session-load errors, no-session without error, startAutoRefresh reset, legacy lock path, and subscriber re-entry. Plus 2 supabase-js flow-through tests. All pass.

mandarini (changes requested):
Hi @7vignesh! Sorry for the back and forth: my point 2 last round asked you to reset the counter in `_startAutoRefresh()`, and that was wrong advice. `_onVisibilityChanged` calls it on every tab refocus, so the budget refills constantly and the event would almost never fire in a browser.

Changes needed:

1. Move the reset out of `_startAutoRefresh()` into `_saveSession()` and `_removeSession()` (next to the existing `this.lastRefreshFailure = null`). The reset in `_callRefreshToken` then becomes redundant.

2. Make it recoverable off-browser. In React Native and Node nothing restarts the ticker after the limit is hit, not even signing back in. In `_saveSession`, after the reset:
   ```ts
   if (this.autoRefreshToken && this.autoRefreshTicker === null && !isBrowser()) {
     void this._startAutoRefresh()
   }
   ```

3. Pass `false` as the third arg to `_notifyAllSubscribers`. It currently broadcasts, so other tabs emit `TOKEN_REFRESH_FAILED` while their own ticker is still running.

4. Pass the session instead of `null`. `_callRefreshToken` preserves sessions whose access token is still valid, and `(event, session) => setUser(session?.user ?? null)` would log those users out.

5. Move the caveats onto the public JSDoc for `maxAutoRefreshFailures` in both `auth-js/src/lib/types.ts` and `supabase-js/src/lib/types.ts`: counts ticks not network attempts (N x 30s), retryable errors only, and the tab refocus / off-browser behavior.

6. Add one test that goes through the real path (seed storage, make the injected `fetch` reject) instead of mocking `_useSession`, and make `nonRetryableError()` a real error: `new AuthApiError('Token revoked', 401, 'refresh_token_not_found')`. A plain object has no `__isAuthError` brand, so that test currently passes for the wrong reason.

7. `pnpm nx format`, line 41 of the new test file is over our 100 char width.

Your call on one thing: the two `catch (e)` blocks count without the retryable gate, so a storage adapter throw counts too. Gate them or document the counter as "failed ticks".

Thank you again for sticking with this, and sorry for the wrong turn on the reset location.

7vignesh:
Hi @mandarini, all 7 points addressed. Sorry for the churn on the earlier rounds, this push should be solid.

**1. Reset location:** Moved from `_startAutoRefresh()` to `_saveSession()` and `_removeSession()`. Removed the redundant reset in `_callRefreshToken` and the tick paths (since `_saveSession` is always called on successful refresh).

**2. Recoverable off-browser:** Added restart logic in `_saveSession()`: when the ticker is null and `autoRefreshToken` is enabled in a non-browser environment, `_startAutoRefresh()` is called so signing back in or `setSession` recovers auto-refresh.

**3. No broadcast:** Changed to `_notifyAllSubscribers('TOKEN_REFRESH_FAILED', session, false)` so other tabs don't emit the event while their own ticker is still running.

**4. Pass session:** `_handleAutoRefreshFailure` now accepts and passes the current session (from the tick callback) instead of null. Catch blocks outside the session scope pass null (default param).

**5. JSDoc caveats:** Updated both `auth-js/src/lib/types.ts` and `supabase-js/src/lib/types.ts` to document: counts ticks not network attempts (~30s each), retryable errors only, and tab refocus / off-browser recovery behavior.

**6. Real-path test + real errors:** Added a test that seeds storage directly and goes through the real `__loadSession` path (no `_useSession` mock). Changed `nonRetryableError()` to use a real `AuthApiError('Token revoked', 401, 'refresh_token_not_found')` with the proper `__isAuthError` brand.

**7. Format:** Ran `nx format:write`, all clean.

**On the catch blocks:** Left them counting all thrown errors (including storage adapter throws) since they represent genuinely failed ticks from the app's perspective. Documented the counter as "failed ticks" in the JSDoc.

13 tests passing, build green. Thanks for the patience and the detailed guidance!

zsspan:
@mandarini

PR #2576, opened 2026-07-29
fix(storage): url-encode object keys in file API request paths

Sy-D:
## Description

Object keys may contain characters the Storage server accepts but that change the meaning of a URL. `_getFinalPath` interpolated them into the request path raw, so a valid key was sent as key-plus-querystring and the server resolved **a different object**, with no error:

```
exists("folder/a?b.txt")
  → GET /object/bucket/folder/a?b.txt     // server sees key "folder/a"
```

This is the same failure #2545 fixed for the CDN purge methods — the helper and rationale added there (`encodeStoragePath`) were never applied to the rest of the file API.

Affected: `upload`/`update`, `createSignedUploadUrl`, `createSignedUrl`, `createSignedUrls`, `info`, `exists`, `getPublicUrl`.

## Which characters this is about

I probed a local Storage server rather than guessing. It **accepts** these in an object key and stores them verbatim:

`?` &nbsp; `space` &nbsp; `&` &nbsp; `=` &nbsp; `+` &nbsp; `,` &nbsp; `(` &nbsp; `)` &nbsp; `@` &nbsp; `:` &nbsp; `;` &nbsp; `$` &nbsp; `*` &nbsp; `!` &nbsp; and the apostrophe

It **rejects** `#`, `%`, `[`, `]`, `~`, backtick, `^`, `|` and all non-ASCII with `400 InvalidKey`. So `#` was never the interesting case — `?`, a space, `&`, `=` and `+` are.

## What changed

Encoding moved into `_getFinalPath`, so every call site is covered and a new one cannot silently reintroduce the bug. All 8 uses of `_getFinalPath` feed a URL; none goes into a request body.

Two follow-on removals, both required to avoid double-encoding:

- `purgeCache` no longer wraps with `encodeStoragePath` (now redundant).
- `getPublicUrl` no longer wraps with `encodeURI`. That wrapper was insufficient anyway — `encodeURI` leaves `?`, `&`, `=` and `+` untouched — and it would double-encode the new output.

The `encodeURI` in `createSignedUrl` **stays**. It wraps `data.signedURL` from the server response, and the server returns that path with `%20` already decoded to a literal space, so it is still required there.

## Testing

`packages/core/storage-js/test/encode-object-key.test.ts` — 4 unit tests (captured `fetch`, URL shape) plus 3 integration tests against the Docker-backed Storage server that upload `folder/a?b&c=d e+f.txt` and read the same object back.

On `master`, 6 of the 7 fail. The integration failure is the telling one: the upload lands as an object literally named `a`, and `list("folder")` returns `["a"]`.

Full `nx test:storage storage-js`:

| | Suites | Tests | Snapshots |
|---|---|---|---|
| `master` | 18 passed / 18 | 411 passed, 2 skipped | 6 passed |
| this branch | 19 passed / 19 | 418 passed, 2 skipped | 6 passed |

Both fully green; the delta is exactly this suite and its 7 tests.

`nx lint storage-js` reports 15 errors both here and on `master` — all pre-existing. `nx format:check` and `nx build storage-js` pass.

## A server-side bug this does *not* fix

With this patch, `upload`, `exists`, `info`, `download` and `getPublicUrl` work for every accepted character. `createSignedUrl` does not: it returns 400 for `?` `&` `=` `+` `,` `;` `@` `:` `$`, and works only for keys whose sole escape is `%20` — plus the characters that need no escaping at all.

The cause is server-side: the token is signed over the percent-encoded key, but the request is validated against the decoded path, so the signature never matches. **The URL the server itself returns fails identically**, so no client-side encoding can work around it. Two tests pin this behaviour explicitly, including the 400, so the limitation is documented rather than silently shipped. Happy to open a separate `storage-api` issue if that would help.

## Also worth knowing

Since `%` is an invalid key character, callers who percent-encode their keys before passing them in were already broken and remain so — they now fail at the client-encoding step instead of reaching the server. I do not believe that is a regression, but it is a behaviour change worth a maintainer eye.

## Type of Change

- [x] Bug fix (fix)

## Checklist

- [x] Unit tests added and passing
- [x] Integration tests added and passing (`nx test:storage storage-js`, fully green)
- [x] Builds passing (`nx build storage-js`)
- [x] Used conventional commits

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -1101,8 +1101,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
     return {
       data: {
         publicUrl:
-          encodeURI(`${this.url}/${renderPath}/public/${_path}`) +
-          (queryString ? `?${queryString}` : ''),
+          `${this.url}/${renderPath}/public/${_path}` + (queryString ? `?${queryString}` : ''),
       },
     }
   }
@@ -1216,7 +1215,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       }
   > {
     return this.handleOperation(async () => {
-      const _path = encodeStoragePath(this._getFinalPath(path))
+      const _path = this._getFinalPath(path)
       const query = new URLSearchParams()
       if (options?.transformations) {
         query.set('transformations', 'true')
@@ -1493,7 +1492,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
   }
 
   private _getFinalPath(path: string) {
-    return `${this.bucketId}/${path.replace(/^\/+/, '')}`
+    return encodeStoragePath(`${this.bucketId}/${path.replace(/^\/+/, '')}`)
   }
 
   private _removeEmptyFolders(path: string) {
--- a/packages/core/storage-js/test/encode-object-key.test.ts
+++ b/packages/core/storage-js/test/encode-object-key.test.ts
@@ -1,0 +1,120 @@
+import { StorageClient } from '../src/index'
+
+// `?` ` ` `&` `=` `+` are all accepted by the Storage server as part of an object
+// key, and all of them change the meaning of a URL if interpolated raw.
+const KEY = 'folder/a?b&c=d e+f.txt'
+const ENCODED = 'folder/a%3Fb%26c%3Dd%20e%2Bf.txt'
+
+// Unit test (custom fetch capture) — proves such a key stays inside the path
+// instead of starting a querystring.
+describe('object keys containing URL delimiters', () => {
+  const URL = 'http://localhost/storage/v1'
+
+  const makeClient = () => {
+    const urls: string[] = []
+    const fetch = (async (url: string) => {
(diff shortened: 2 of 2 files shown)


Sy-D:
Thanks — I looked at this one carefully, but I do not think it applies here.

The token is not a credential. Its payload is:

```json
{ "iss": "supabase-demo", "role": "service_role", "exp": 1983812996 }
```

That is the public demo key the Supabase CLI ships for local development, not a secret belonging to any project. The exact same string is already hardcoded in three other test files in this package:

- `packages/core/storage-js/test/storageFileApi.test.ts`
- `packages/core/storage-js/test/storageFileApiNode.test.ts`
- `packages/core/storage-js/test/storageApi.test.ts`

each with the same `// secret key - bypasses RLS for testing` / `// Supabase CLI local development defaults` comment. I copied that convention deliberately so the new test reads like its neighbours.

Loading it from an env var or CI secret instead would make this one file behave differently from the other three, and add a failure mode (missing secret) for a value that is a public constant. There is also nothing to rotate.

Happy to switch if maintainers would rather move all four files to a shared constant or an env lookup — but that seems like a separate change, not something this PR should do on its own.

PR #2582, opened 2026-07-30
docs(repo): fix broken RELEASE.md and example migration links

AlvaroBalbin:
While reading the docs I noticed two relative markdown links that point at paths which do not exist. CONTRIBUTING.md linked to ./RELEASE.md, but the file lives at docs/RELEASE.md, and the realtime-js example README linked to the SQL migration using a repository-root path instead of one relative to that README. I corrected both targets so they resolve on GitHub and in local editors.

I left the ./docs/SECURITY.md link in README.md alone: that file is not present anywhere in the repository, and I did not want to guess whether it should point at an org-level policy or a file still to be added. Worth noting AGENTS.md has the same broken link, so a maintainer may want to decide on both together.

I pushed this to my fork first and the project's own CI workflows pass on the commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -336,7 +336,7 @@ This allows you to test package changes without waiting for an official release,
 
 ### Official Releases
 
-Official releases are handled by maintainers using Nx Release. You can read more in the [RELEASE.md](./RELEASE.md)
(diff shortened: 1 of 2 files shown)


mandarini (approved):

mandarini:
Thanks

PR #2597, opened 2026-08-06
fix(realtime): clear stale join payload on sign-out

### LINKED ISSUE #2596: realtime-js: setAuth(null) leaves the previous JWT in channel join payloads, causing endless rejoins with a dead token after sign-out
## Bug description

In `realtime-js`, `_performAuth` only updates each channel's join payload when the new token is truthy:

https://github.com/supabase/supabase-js/blob/master/packages/core/realtime-js/src/RealtimeClient.ts#L640-L655

```ts
if (this.accessTokenValue != tokenToSend) {
  this.accessTokenValue = tokenToSend
  this.channels.forEach((channel) => {
    const payload = { access_token: tokenToSend, version: DEFAULT_VERSION }
    tokenToSend && channel.updateJoinPayload(payload)   // <-- skipped when tokenToSend is null
    ...
```

When a user signs out (`supabase.auth.signOut()` → the client ends up calling `setAuth()` with no session, so `tokenToSend` resolves to `null`):

- joined channels are sent `access_token: null` (the `push` below the guard still runs), but
- the channel's **join payload keeps the previous user JWT**.

From then on, any rejoin — socket blip, server restart, or the server erroring the channel — replays the stale JWT. Once that JWT expires, the join is rejected (`JwtSignatureError` / token expired), the channel goes to `errored`, and the rejoin timer retries forever (there is no retry cap), producing `CHANNEL_ERROR` roughly every 5–6 seconds indefinitely plus continuous websocket traffic.

This bites any app that keeps a `postgres_changes` subscription mounted across an auth sign-out — e.g. a second browser tab receiving the cross-tab `SIGNED_OUT` broadcast while its UI (and channel) stays alive.

## Reproduction

Observed on `@supabase/supabase-js` / `@supabase/realtime-js` 2.105.4; the guard is unchanged on master.

1. Table with an RLS SELECT policy `TO authenticated`, realtime enabled.
2. Sign in, subscribe:
   ```js
   const channel = supabase
     .channel('repro')
     .on('postgres_changes', { event: 'UPDATE', schema: 'public', table: 't' }, () => {})
   await supabase.realtime.setAuth(session.access_token)
   channel.subscribe((status, err) => console.log(status, err ?? ''))
   ```
3. `await supabase.auth.signOut()` (or directly `await supabase.realtime.setAuth(null)`) while the channel stays subscribed.
4. Force a rejoin (drop the connection, or wait for the server to error the channel after the null-token push).

**Actual:** the rejoin's `phx_join` carries the old JWT; after expiry every retry fails and the `CHANNEL_ERROR` → rejoin loop runs unbounded.

**Expected:** after `setAuth(null)` the join payload should fall back to the apikey (matching a channel created fresh in the signed-out state, which joins as `anon` and works), or the channel should stop rejoining once its token is known-dead.

A separate but related hardening: the rejoin timer retries forever on `CHANNEL_ERROR` with no cap or terminal state for permanently rejected joins (e.g. invalid JWT) — a cap or an "auth rejected" terminal status would keep this failure mode observable instead of an infinite loop.

## Suggested fix

Always call `channel.updateJoinPayload({ access_token: tokenToSend, version: DEFAULT_VERSION })`, letting the downstream join fall back to the apikey when the token is null — i.e. drop the `tokenToSend &&` guard (or explicitly substitute the apikey when the token is null).

7ttp:
## TL;DR

Clear existing channel join payloads when the callback-based access token becomes `null` after sign-out
Subsequent rejoins then use the signed-out API-key path instead of replaying the previous JWT

## Problem

When the access-token callback returned `null`, `RealtimeClient` cleared its cached token but
 skipped updating the channels' stored join payloads. A later `phx_join` therefore replayed the previous JWT

## sol

Always update each channel's join payload when the resolved token changes, including `null`.

Track `_authGeneration` so only the latest `setAuth()` call can apply its result, 
preventing an older refresh from restoring a stale JWT after sign-out.

The join-payload change only affects future joins; manual-token semantics, callback refresh behavior, and the existing in-place `access_token: null` message are unchanged...

## ref:

- Closes supabase/supabase-js#2596

### DIFF
--- a/packages/core/realtime-js/src/RealtimeClient.ts
+++ b/packages/core/realtime-js/src/RealtimeClient.ts
@@ -645,7 +645,7 @@ export default class RealtimeClient {
           version: DEFAULT_VERSION,
         }
 
-        tokenToSend && channel.updateJoinPayload(payload)
+        channel.updateJoinPayload(payload)
 
         if (channel.joinedOnce && channel.channelAdapter.isJoined()) {
           channel.channelAdapter.push(CHANNEL_EVENTS.access_token, {
--- a/packages/core/realtime-js/src/RealtimeClient.ts
+++ b/packages/core/realtime-js/src/RealtimeClient.ts
@@ -226,6 +226,7 @@ export default class RealtimeClient {
 
   private _manuallySetToken: boolean = false
   private _authPromise: Promise<void> | null = null
+  private _authGeneration: number = 0
   private _workerHeartbeatTimer: HeartbeatTimer = undefined
   private _pendingWorkerHeartbeatRef: string | null = null
   private _pendingDisconnectTimer: ReturnType<typeof setTimeout> | null = null
@@ -502,11 +503,17 @@ export default class RealtimeClient {
    * @category Realtime
    */
   async setAuth(token: string | null = null): Promise<void> {
-    this._authPromise = this._performAuth(token)
+    const authGeneration = ++this._authGeneration
+    const authPromise = this._performAuth(token, authGeneration)
+    if (authGeneration === this._authGeneration) {
+      this._authPromise = authPromise
+    }
     try {
-      await this._authPromise
+      await authPromise
     } finally {
-      this._authPromise = null
+      if (this._authPromise === authPromise) {
+        this._authPromise = null
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


mandarini (approved):

mandarini:
@7ttp, thank you so much!! :D One small ask before merging: could you update the PR description to also mention the `_authGeneration` addition? It fixes a real race (a stale token refresh resolving after sign-out could otherwise res
urrect the old JWT), but right now the description only covers the join payload guard, so it is easy for a reviewer to miss what that part is actually for.

PR #2615, opened 2026-08-15
fix(auth): convert stolen lock AbortError when acquireTimeout is 0

### LINKED ISSUE #2620: auth-js: stolen-lock AbortError leaks unconverted when acquireTimeout is 0 (_autoRefreshTokenTick)
### Describe the bug

`navigatorLock` converts a stolen-lock `AbortError` into a typed `NavigatorLockAcquireTimeoutError` — explicitly so it does not leak to error trackers — but the conversion is gated behind `acquireTimeout > 0`. `_autoRefreshTokenTick()` is the only caller that passes `acquireTimeout = 0`, so it is the one caller the conversion never protects. The raw `AbortError` escapes and surfaces as an unhandled rejection.

I hit this in production as:

```
AbortError: Lock broken by another request with the 'steal' option.
```

reported with `mechanism: auto.browser.global_handlers.onunhandledrejection`.

### To Reproduce

Conditions: a browser client created with `autoRefreshToken: true` (so `_autoRefreshTokenTick()` polls in the background), and a second context contending for the same auth lock — another tab, or another auth call using the default `lockAcquireTimeout` of 5000 ms.

1. The background tick calls `_acquireLock(0, …)`, which requests the lock with `{ mode: 'exclusive', ifAvailable: true }` and **no abort signal**.
2. The other context times out waiting, concludes the lock is orphaned, and re-requests with `{ steal: true }` — the recovery path added for the orphaned-lock hang.
3. The steal breaks the tick's pending request. The browser rejects it with `AbortError: Lock broken by another request with the 'steal' option.`

### The gap

In `src/lib/locks.ts`, `navigatorLock`'s catch:

```js
if (e.name === 'AbortError' && acquireTimeout > 0) {
    if (abortController.signal.aborted) {
        // OUR timeout fired — steal it
    } else {
        // We HELD the lock but another request stole it from us.
        // Convert to a typed error so callers (e.g. _autoRefreshTokenTick)
        // can handle/filter it without it leaking to Sentry as a raw AbortError.
        throw new NavigatorLockAcquireTimeoutError(...)
    }
}
throw e   // ← acquireTimeout === 0 lands here, raw
```

The comment names `_autoRefreshTokenTick` as the caller the conversion is meant to serve, but that caller passes `acquireTimeout = 0`, so the guard is false and neither branch runs.

Then in `GoTrueClient._autoRefreshTokenTick()`:

```js
catch (e) {
    if (e.isAcquireTimeout || e instanceof LockAcquireTimeoutError) {
        this._debug('auto refresh token tick lock not available')
    } else {
        throw e   // ← raw AbortError is rethrown
    }
}
```

The two halves are designed to fit together, but the guard excludes exactly the caller that needs it.

Note that `ifAvailable: true` requests carry no abort signal yet are still breakable by a steal, so `acquireTimeout > 0` is not a sound proxy for "this request could have been aborted by our own timeout".

### Expected behavior

A stolen lock should produce the typed `NavigatorLockAcquireTimeoutError` regardless of `acquireTimeout`, so `_autoRefreshTokenTick()`'s existing catch swallows it as intended and it never reaches an error tracker.

A possible shape: drop `acquireTimeout > 0` from the outer guard and branch on `abortController.signal.aborted` alone — with no timer armed, `signal.aborted` is false for the `acquireTimeout <= 0` case, which routes it to the "stolen from us" branch. The steal-back protection in that branch already prevents a cascade.

### Version

Observed in `@supabase/auth-js` 2.99.3. I checked the published 2.112.3 and the guard is unchanged there — the only difference in that catch is a more robust structural check for `DOMException`, so upgrading does not resolve it.

### Additional context

The `steal: true` recovery this comes from was introduced to fix orphaned locks hanging auth methods. This is not a request to revert that — the recovery works. It is only that the victim of a steal reports a raw `AbortError` for one specific caller, which turns a handled, benign situation into unhandled-rejection noise that users of the library then have to filter out by message.

abhaysahu1008:
### What changed?
Removed the `acquireTimeout > 0` condition inside the `AbortError` catch/check block within `navigatorLock` (`packages/core/auth-js/src/lib/locks.ts`).

When `acquireTimeout` is set to `0` (such as in `_autoRefreshTokenTick`), stolen Web Locks produce an `AbortError`. Previously, because `acquireTimeout > 0` was required to process the `AbortError`, stolen locks bypassed error handling when `acquireTimeout` was `0`. This caused unhandled raw `AbortError` exceptions to leak to client application error trackers (e.g., Sentry) instead of properly resolving or throwing a structured `NavigatorLockAcquireTimeoutError`.

### How was it tested?
- Tested locally using unit tests for lock mechanisms: `npx nx test auth-js -- test/lib/locks.test.ts`
- Verified all lock tests pass (`locks.test.ts`).
- Ran `npx nx format:write` to ensure code style compliance.

Closes supabase/supabase-js#2620

### DIFF
--- a/packages/core/auth-js/src/lib/locks.ts
+++ b/packages/core/auth-js/src/lib/locks.ts
@@ -200,8 +200,7 @@ export async function navigatorLock<R>(
       e !== null &&
       typeof e === 'object' &&
       'name' in e &&
-      e.name === 'AbortError' &&
-      acquireTimeout > 0
+      e.name === 'AbortError' // <--- Fixed!
     ) {
       if (abortController.signal.aborted) {
         // OUR timeout fired — the lock is genuinely orphaned. Steal it.


mandarini:
Hi @abhaysahu1008, thank you so much for contributing to Supabase! 💚 

You correctly identified the root cause here, and you proposed this fix first. I ended up merging #2616, which contains the same guard change along with a regression test covering the stolen-lock case with `acquireTimeout = 0`, so I'm closing this one as a duplicate.

Your diagnosis was spot on and directly matched what we confirmed on our side, so please don't let the close discourage you. If you'd like to pick up something else, we're happy to help you find a good issue.

Thank you again for taking the time to dig into this, contributions like yours are what keep these libraries healthy.

PR #2623, opened 2026-08-20
fix(functions): honour an already-aborted signal when timeout is set

### LINKED ISSUE #2622: functions-js: invoke() ignores an already-aborted signal when `timeout` is set, so the function still runs
### Describe the bug

`FunctionsClient.invoke()` accepts both a caller `signal` and a `timeout`. When both are given, it creates an internal `AbortController` for the timeout and bridges the caller's abort onto it by subscribing to the caller signal:

```ts
// packages/core/functions-js/src/FunctionsClient.ts (v2.112.3, L272-281)
if (signal) {
  effectiveSignal = timeoutController.signal
  // If the user's signal is aborted, abort our timeout controller too.
  onAbort = () => timeoutController!.abort()
  signal.addEventListener('abort', onAbort)
}
```

An `AbortSignal` that is **already aborted** never dispatches another `abort` event, so that listener never fires. The signal handed to `fetch` is the fresh, un-aborted `timeoutController.signal`, so the request goes out: the Edge Function actually executes and `invoke()` resolves with its response as if nothing had been cancelled.

The same call **without** `timeout` behaves correctly — `effectiveSignal = signal`, and `fetch` rejects immediately without issuing a request. So the two paths disagree, and adding a `timeout` silently turns off pre-flight cancellation.

A second, smaller consequence of the same block: the caller's abort **reason** is dropped. `timeoutController.abort()` is called with no argument, so a caller who aborts with `controller.abort(myReason)` can never see `myReason` — it is replaced by a generic `AbortError`.

### Library affected

functions-js

### Reproduction

Self-contained, no Supabase project needed — a local HTTP server counts how many requests actually arrive.

```js
// repro.cjs — node repro.cjs
const http = require('node:http')
const { FunctionsClient } = require('@supabase/functions-js')

let hits = 0
const server = http.createServer((req, res) => {
  hits++
  res.writeHead(200, { 'Content-Type': 'application/json' })
  res.end(JSON.stringify({ ok: true }))
})

async function run(label, opts, url) {
  controller.abort() // aborted BEFORE invoke
  const before = hits
  const client = new FunctionsClient(url)
  const { data, error } = await client.invoke('my-function', { ...opts, signal: controller.signal })
  console.log(`${label}\n  server hits: ${hits - before}\n  error: ${error ? error.name : 'null'}\n  data: ${JSON.stringify(data)}\n`)
}

server.listen(0, async () => {
  const url = `http://127.0.0.1:${server.address().port}`
  await run('no timeout   (expect 0 hits, aborted)', {}, url)
  await run('with timeout (expect 0 hits, aborted)', { timeout: 5000 }, url)
  server.close()
})
```

### Steps to reproduce

1. `npm i @supabase/functions-js@2.112.3`
2. Save the snippet above as `repro.cjs`
3. `node repro.cjs`

### Actual behaviour

```
no timeout   (expect 0 hits, aborted)
  server hits: 0
  error: FunctionsFetchError
  data: null

with timeout (expect 0 hits, aborted)
  server hits: 1          <-- the function ran
  error: null             <-- and invoke() reports success
  data: {"ok":true}
```

### Expected behaviour

`server hits: 0` in both cases: an already-aborted signal should prevent the request, exactly as it does when `timeout` is not passed.

### Why it matters

This is the ordinary "cancel in-flight work" pattern — a React `useEffect` cleanup controller, `AbortSignal.timeout(...)`, or one request-scoped controller shared across a batch of calls. Any time the controller is aborted before control reaches `invoke()` (the component unmounted while an earlier `await` was pending, an earlier item in the batch already failed), the function is still invoked. That burns an invocation and, more importantly, runs whatever side effects the function has — writes, emails, payment calls — after the caller has cancelled. The caller sees a success result for a request they cancelled.

### Suggested fix

Mirror the already-aborted state onto the timeout controller up front, and forward the reason in both branches:

```ts
if (signal.aborted) {
  timeoutController.abort(signal.reason)
} else {
  onAbort = () => timeoutController!.abort(signal.reason)
  signal.addEventListener('abort', onAbort)
}
```

`AbortSignal.any([signal, timeoutController.signal])` would also solve it, but it needs Chrome 116+ / Safari 17.4+ / Firefox 124+; the explicit guard above leaves the browser support matrix untouched.

I have a PR ready with this fix plus regression tests in `test/FunctionsClient.test.ts`, and will link it here.

### System Info

- `@supabase/functions-js` 2.112.3 (`latest`); code path unchanged on `master` @ `a249594`
- Node.js v25.9.0
- OS: Windows 11

serhiileniv:
### What changed?

`FunctionsClient.invoke()` bridges a caller-supplied `signal` onto its internal timeout `AbortController` with an `abort` listener:

```ts
onAbort = () => timeoutController!.abort()
signal.addEventListener('abort', onAbort)
```

An `AbortSignal` that is **already aborted** never dispatches another `abort` event, so that listener never fires. `fetch` then receives the fresh, un-aborted `timeoutController.signal` and the request goes out — the Edge Function runs, and `invoke()` resolves with its response as if nothing had been cancelled.

This PR mirrors the aborted state onto the timeout controller up front, and forwards `signal.reason` in both branches:

```ts
if (signal.aborted) {
  timeoutController.abort(signal.reason)
} else {
  onAbort = () => timeoutController!.abort(signal.reason)
  signal.addEventListener('abort', onAbort)
}
```

Passing `timeout` silently disabled pre-flight cancellation. The **same call without `timeout` already behaves correctly** (`effectiveSignal = signal`, and `fetch` rejects without issuing a request), so the two paths disagreed — this restores consistency rather than introducing new behaviour.

This is the ordinary "cancel in-flight work" pattern: a React `useEffect` cleanup controller, `AbortSignal.timeout(...)`, or one request-scoped controller shared across a batch. Whenever the controller is aborted before control reaches `invoke()` — the component unmounted while an earlier `await` was pending, an earlier item in the batch failed — the function was still invoked. That burns an invocation and runs whatever side effects the function has (writes, emails, payment calls) after the caller cancelled, while reporting success back to the caller.

The dropped `signal.reason` is the second half of the same block: `timeoutController.abort()` took no argument, so a caller aborting with `controller.abort(myReason)` could never observe `myReason`.

Closes #2622

## 📸 Screenshots/Examples

Self-contained repro (no Supabase project needed) — a local HTTP server counts requests that actually arrive:

```js
// repro.cjs
const http = require('node:http')
const { FunctionsClient } = require('@supabase/functions-js')

let hits = 0
const server = http.createServer((req, res) => {
  hits++
  res.writeHead(200, { 'Content-Type': 'application/json' })
  res.end(JSON.stringify({ ok: true }))
})

async function run(label, opts, url) {
  controller.abort() // aborted BEFORE invoke
  const before = hits
  const client = new FunctionsClient(url)
  const { data, error } = await client.invoke('my-function', { ...opts, signal: controller.signal })
  console.log(`${label}\n  server hits: ${hits - before}\n  error: ${error ? error.name : 'null'}\n  data: ${JSON.stringify(data)}\n`)
}

server.listen(0, async () => {
  const url = `http://127.0.0.1:${server.address().port}`
  await run('no timeout   (expect 0 hits, aborted)', {}, url)
  await run('with timeout (expect 0 hits, aborted)', { timeout: 5000 }, url)
  server.close()
})
```

Before:

```
no timeout   (expect 0 hits, aborted)
  server hits: 0
  error: FunctionsFetchError
  data: null

with timeout (expect 0 hits, aborted)
  server hits: 1          <-- the function ran
  error: null             <-- and invoke() reported success
  data: {"ok":true}
```

After — `server hits: 0`, `error: FunctionsFetchError`, `data: null` in both cases.

## 🔄 Breaking changes

Two behaviour notes, both aligning `timeout + signal` with the existing `signal`-only path:

- An already-aborted signal now prevents the request instead of letting it through. Code that (knowingly or not) relied on the request still going out would change — but that path already reported a success the caller had cancelled.
- The abort reason handed to `fetch` is now the caller's `signal.reason` rather than a generic `AbortError`. `invoke()` still returns a `FunctionsFetchError`, so the shape of the result is unchanged; `test/spec/timeout.spec.ts` (which asserts `error.name === 'FunctionsFetchError'` on a mid-flight abort) still holds.

`AbortSignal.any([signal, timeoutController.signal])` would also fix this, but it needs Chrome 116+ / Safari 17.4+ / Firefox 124+; the explicit guard leaves the browser support matrix untouched.

## 📋 Checklist

- [ ] I have updated documentation (if applicable) — no public API or documented behaviour changed

## 📝 Additional notes

Two regression tests were added to `packages/core/functions-js/test/FunctionsClient.test.ts`, next to the existing listener-cleanup tests for this same block. They use a mock fetch that mirrors the platform contract (reject with `signal.reason` instead of issuing a request when the signal is already aborted), and cover both the request being prevented and the reason being forwarded. Both fail on `master` and pass with this change.

Verified locally: `jest test/FunctionsClient.test.ts` (4/4 passing), `nx build functions-js`, `nx format:check`, and `nx lint functions-js` (no new findings — the two pre-existing `prefer-const` errors in this file are unchanged).

I used Claude Code while investigating and preparing this change; the diagnosis, repro, and tests above are ones I ran and verified myself.

### DIFF
--- a/packages/core/functions-js/src/FunctionsClient.ts
+++ b/packages/core/functions-js/src/FunctionsClient.ts
@@ -272,10 +272,17 @@ export class FunctionsClient {
         // If user provided their own signal, we need to respect both
         if (signal) {
           effectiveSignal = timeoutController.signal
-          // If the user's signal is aborted, abort our timeout controller too.
-          // Store the listener so we can clean it up in finally.
-          onAbort = () => timeoutController!.abort()
-          signal.addEventListener('abort', onAbort)
+          if (signal.aborted) {
+            // A signal that was already aborted never fires another `abort`
+            // event, so mirror its state onto the timeout controller up front —
+            // otherwise the request would go out despite being cancelled.
+            timeoutController.abort(signal.reason)
+          } else {
+            // If the user's signal is aborted, abort our timeout controller too.
+            // Store the listener so we can clean it up in finally.
+            onAbort = () => timeoutController!.abort(signal.reason)
+            signal.addEventListener('abort', onAbort)
+          }
         } else {
           effectiveSignal = timeoutController.signal
         }
--- a/packages/core/functions-js/test/FunctionsClient.test.ts
+++ b/packages/core/functions-js/test/FunctionsClient.test.ts
@@ -1,4 +1,4 @@
-import { FunctionsClient } from '../src/index'
+import { FunctionsClient, FunctionsFetchError } from '../src/index'
 
 describe('FunctionsClient', () => {
   describe('invoke – abort listener cleanup when timeout + signal are both set', () => {
@@ -45,6 +45,50 @@ describe('FunctionsClient', () => {
 
       expect(addedFn).toBeDefined()
       expect(addedFn).toBe(removedFn)
+    })
+  })
+
+  describe('invoke – caller signal already aborted when timeout is also set', () => {
+    // Mirrors the platform contract: fetch rejects with the signal's reason
+    // instead of issuing a request when the signal is already aborted.
+    const abortAwareFetch = () =>
+      jest.fn().mockImplementation((_url: string, init?: RequestInit) => {
+        if (init?.signal?.aborted) {
+          return Promise.reject(init.signal.reason)
+        }
+        return Promise.resolve({
+          ok: true,
+          headers: { get: () => 'application/json' },
+          json: () => Promise.resolve({ ok: true }),
+        })
+      })
+
+    it('does not let the request reach the function', async () => {
+      const mockFetch = abortAwareFetch()
+      const client = new FunctionsClient('http://localhost', { customFetch: mockFetch })
+      const controller = new AbortController()
+      controller.abort()
+
+      const { data, error } = await client.invoke('test-fn', {
+        timeout: 5000,
+        signal: controller.signal,
+      })
+
+      expect(data).toBeNull()
+      expect(error).toBeInstanceOf(FunctionsFetchError)
+      expect(mockFetch.mock.calls[0][1]?.signal?.aborted).toBe(true)
+    })
+
+    it('forwards the caller abort reason', async () => {
+      const mockFetch = abortAwareFetch()
+      const client = new FunctionsClient('http://localhost', { customFetch: mockFetch })
+      const controller = new AbortController()
+      const reason = new Error('caller cancelled')
+      controller.abort(reason)
+
+      await client.invoke('test-fn', { timeout: 5000, signal: controller.signal })
+
+      expect(mockFetch.mock.calls[0][1]?.signal?.reason).toBe(reason)
     })
   })
 })
--- a/packages/core/functions-js/test/FunctionsClient.test.ts
+++ b/packages/core/functions-js/test/FunctionsClient.test.ts
@@ -50,21 +50,27 @@ describe('FunctionsClient', () => {
 
   describe('invoke – caller signal already aborted when timeout is also set', () => {
     // Mirrors the platform contract: fetch rejects with the signal's reason
-    // instead of issuing a request when the signal is already aborted.
-    const abortAwareFetch = () =>
-      jest.fn().mockImplementation((_url: string, init?: RequestInit) => {
+    // instead of issuing a request when the signal is already aborted. `reached`
+    // stands in for the hit counter on the repro's local server: it only counts
+    // requests that would actually have run the Edge Function.
+    const abortAwareFetch = () => {
+      const server = { reached: 0 }
+      const fetch = jest.fn().mockImplementation((_url: string, init?: RequestInit) => {
         if (init?.signal?.aborted) {
           return Promise.reject(init.signal.reason)
         }
+        server.reached++
         return Promise.resolve({
           ok: true,
           headers: { get: () => 'application/json' },
           json: () => Promise.resolve({ ok: true }),
         })
       })
+      return { fetch, server }
+    }
 
     it('does not let the request reach the function', async () => {
-      const mockFetch = abortAwareFetch()
+      const { fetch: mockFetch, server } = abortAwareFetch()
       const client = new FunctionsClient('http://localhost', { customFetch: mockFetch })
       const controller = new AbortController()
       controller.abort()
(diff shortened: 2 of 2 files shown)


cekuu35:
## PR review: fix confirmed correct

The guard is the right approach. The root cause is specific: `AbortSignal.addEventListener("abort", ...)` is a one-shot subscription that only fires on **future** abort events, so a signal that already dispatched `abort` is silently ignored. The `signal.aborted` check up front handles the pre-aborted case, and the `else` branch handles the normal race. Both branches now forward `signal.reason`, which the original code dropped.

One thing to verify in the regression tests: the test with `timeout` should confirm the server-side handler never executes (the local HTTP server in the repro counts `hits`). Worth adding a test for the race where the signal aborts **during** the fetch (not before) — that path already works because the `addEventListener` fires, but it would document the intended behavior and catch any future regression.

The `AbortSignal.any()` alternative would be more idiomatic but has limited runtime support (no Safari as of 2026-08). The explicit guard is the right choice for broad browser compat.

Clean diff, 2 files, focused fix. LGTM.

serhiileniv:
@cekuu35 Thanks — both test suggestions were worth doing, and they're in `b65c017b`.

**"Confirm the function never executes."** The unit test previously only asserted that `fetch` was handed an already-aborted signal, which is a proxy for the thing that actually matters. The mock now carries the repro server's hit counter: it rejects with `signal.reason` when `init.signal.aborted` (the platform contract), and increments `server.reached` only past that gate. The pre-aborted test asserts `server.reached === 0`, so the property under test is now "no request ever reached the function" rather than "the signal looked right on the way in".

**Abort *during* the fetch.** Added as its own case. It's deterministic rather than tick-counted — the mock resolves a `requestStarted` promise when it's called and returns a pending promise that rejects on `abort`, so the test aborts only once the request is genuinely in flight, then asserts the in-flight signal is aborted and carries the caller's `reason`, and that `invoke()` surfaces a `FunctionsFetchError`.

You're right that this path already worked; I checked the test is load-bearing anyway by deleting `signal.addEventListener('abort', onAbort)` from the source — the new case fails without it. So it pins the listener half of the fix, not just documents it.

**`AbortSignal.any()`** — agreed on both counts. Same reasoning kept it out: the package targets browsers broadly and Safari still doesn't ship it, and a polyfill here would be strictly more code than the two-branch guard.

5/5 green in `packages/core/functions-js` (`npx jest test/FunctionsClient.test.ts`), Prettier clean.

PR #2626, opened 2026-08-23
fix(realtime): clear pendingHeartbeatRef on socket open and close

### LINKED ISSUE #2625: realtime: resetHeartbeat's stale-ref clear is gated on autoSendHeartbeat, so worker-mode clients still false-close on reconnect
## Describe the bug

A `pendingHeartbeatRef` armed on one socket can survive a disconnect onto the next, healthy
connection, whose first heartbeat then takes the timeout branch and force-closes it with
`"heartbeat timeout"` without ever sending a frame. Since the phoenix integration (#2119), the
library handles this correctly for the default configuration — phoenix's `resetHeartbeat()`, called
from `onConnOpen()`, clears the stale ref on every open:

```js
// phoenix.cjs.js — onConnOpen(), lines 1577-1587
onConnOpen() {
  ...
  this.establishedConnections++;      // :1581
  this.flushSendBuffer();             // :1582
  if (this.autoSendHeartbeat) {       // :1584 — THE GATE
    this.resetHeartbeat();            // :1585 — only place pendingHeartbeatRef is cleared on open
  }
  this.triggerStateCallbacks("open"); // :1587
}

// resetHeartbeat() — lines 1609-1614
resetHeartbeat() {
  if (this.conn && this.conn.skipHeartbeat) return;
  this.pendingHeartbeatRef = null;    // the clear
  this.clearHeartbeats();
  this.heartbeatTimer = setTimeout(() => this.sendHeartbeat(), this.heartbeatIntervalMs);
}
```

The clear only runs when `this.autoSendHeartbeat` is true. `RealtimeClient` sets
`autoSendHeartbeat = !this.worker` (`RealtimeClient.js:684`, current through 2.112.3 at line 686) —
i.e. it is **false** for any client configured with `worker: true`. Worker mode exists specifically so
the heartbeat timer isn't throttled by background-tab timer clamping, so it's a normal production
configuration, not an edge case. For those clients, `onConnOpen()` never calls `resetHeartbeat()`, so
a `pendingHeartbeatRef` stranded by a drop-with-beat-in-flight survives onto the next, healthy
connection exactly as before the fix — that connection's first heartbeat sees a non-null
`pendingHeartbeatRef`, takes the timeout branch, and force-closes itself with `"heartbeat timeout"`
having sent zero frames.

So the fix conflates two responsibilities that `resetHeartbeat()` performs together: clearing stale
heartbeat state (correct to do unconditionally on every open) and arming the auto-send timer (correct
to gate on `autoSendHeartbeat`, since worker-mode clients drive heartbeats from the worker instead).
Gating both together on `autoSendHeartbeat` means worker-mode clients get neither.

## To Reproduce

Standalone — no app code, only `@supabase/realtime-js` (which brings in `@supabase/phoenix`). Fakes a
WebSocket transport and a Worker so the sequence runs deterministically without real timers or a
network.

```ts
// repro.ts — run with `npx tsx repro.ts` (or transpile/run any way you like)
import assert from 'node:assert/strict'
import { RealtimeClient } from '@supabase/realtime-js'

// ---- Fake Worker: records the latest instance; ticks only on command (no real setInterval) ----
class FakeWorker {
  onmessage: ((e: { data: { event: string } }) => void) | null = null
  onerror: ((e: { message?: string }) => void) | null = null
  terminated = false
  constructor(_url: string) {}
  postMessage(_m: unknown) {}
  terminate() { this.terminated = true }
  tick() { if (this.onmessage && !this.terminated) this.onmessage({ data: { event: 'keepAlive' } }) }
}
let lastWorker: FakeWorker
;(globalThis as any).Worker = class extends FakeWorker {
  constructor(url: string) { super(url); lastWorker = this } // library terminates+recreates per connection
}
if (typeof URL.createObjectURL !== 'function') (URL as any).createObjectURL = () => 'blob:fake'

// ---- Fake WebSocket transport ----
class FakeWS {
  readyState = 0
  sent: string[] = []
  closedWith: { code: number; reason: string } | null = null
  binaryType = 'arraybuffer'
  onopen: ((e: object) => void) | null = null
  onclose: ((e: object) => void) | null = null
  constructor(public url: string) { instances.push(this) }
  send(d: unknown) { this.sent.push(String(d)) }
  close(code = 1000, reason = '') {
    this.closedWith = { code, reason }
    this.readyState = 3
    queueMicrotask(() => this.onclose?.({ code, reason, wasClean: true }))
  }
}
const instances: FakeWS[] = []
const wait = (ms: number) => new Promise((r) => setTimeout(r, ms))

async function main() {
  const statuses: string[] = []
  const client = new RealtimeClient('wss://repro.invalid/realtime/v1', {
    params: { apikey: 'repro-key' },
    transport: FakeWS as any,
    worker: true,                  // production configuration — the gate this bug is about
    heartbeatIntervalMs: 60_000,   // no auto-fire; beats are driven by hand below
    reconnectAfterMs: () => 10,
    heartbeatCallback: (s) => statuses.push(s),
  })
  const beat = () => lastWorker.tick() // worker mode: a beat is driven by the worker's keepAlive tick

  client.connect()
  await wait(30)
  const A = instances[0]
  A.readyState = 1
  A.onopen?.({})
  await wait(50)

  beat() // arm a heartbeat on A
  await wait(30)
  assert.ok(A.sent.some((f) => f.includes('heartbeat')), 'expected a heartbeat frame on A')

  // Drop A with the beat in flight — a Wi-Fi roam, laptop sleep, or server-side close all do this.
  A.readyState = 3
  A.onclose?.({ code: 1006, reason: 'network drop', wasClean: false })

  let B: FakeWS | undefined
  for (let i = 0; i < 200 && !B; i++) { await wait(10); B = instances[1] }
  if (!B) throw new Error('client never reconnected')
  B.readyState = 1
  B.onopen?.({}) // <-- onConnOpen() runs here; autoSendHeartbeat is false, so no clear happens

  await wait(50)
  beat() // first beat on the NEW, healthy socket
  await wait(30)

  console.log('statuses:', statuses)
  console.log('frames on B:', B.sent.filter((f) => f.includes('heartbeat')).length)
  console.log('B closed with:', B.closedWith)
}

main()
```

## Observed vs Expected

| | worker mode (observed, current behavior) | non-worker mode (for contrast — already correct) |
|---|---|---|
| `statuses` | `['sent', 'timeout']` | `['sent', 'sent']` |
| frames on B | **0** | **1** |
| B closed? | `{ reason: 'heartbeat timeout' }` | not closed |

Expected: worker mode should match the non-worker column — `['sent', 'sent']`, 1 frame on B, B stays
open. The stranded ref from A should have been cleared when B opened, exactly as it is for
`worker: false`.

## Proposed fix

Split `resetHeartbeat()`'s two responsibilities in `onConnOpen()` instead of gating both together —
hoist the ref clear above the gate, leave only the timer-arming call inside it:

```js
onConnOpen() {
  ...
  this.establishedConnections++;
  this.flushSendBuffer();
  this.pendingHeartbeatRef = null;   // moved out of the gate — unconditional on every open
  this.heartbeatSentAt = null;       // moved out of the gate — unconditional on every open
  if (this.autoSendHeartbeat) {
    this.resetHeartbeat();           // now only (re-)arms the auto-send timer
  }
  this.triggerStateCallbacks("open");
}
```

(`resetHeartbeat()` itself keeps clearing the ref when it runs — that's redundant but harmless for the
`autoSendHeartbeat: true` path, which already worked correctly.) This costs nothing for non-worker
clients and fixes worker-mode clients, whose external heartbeat driver (a Worker's `setInterval`, in
this app's case) still needs `pendingHeartbeatRef` cleared on open even though the library isn't the
one arming the next send.

To be explicit about what was already correctly fixed: **non-worker mode was correctly fixed by
`resetHeartbeat()` in the 0.4.x line** — the bug reported here is scoped entirely to
`autoSendHeartbeat === false` (i.e. `worker: true`).

## System information

- `@supabase/realtime-js` 2.108.2 (also affects 2.112.3, the latest checked — identical gate, one line
  later)
- `@supabase/phoenix` 0.4.5 (the vendored dependency where the gate actually lives —
  `onConnOpen`/`resetHeartbeat` in `priv/static/phoenix.cjs.js`)
- Client options: `{ worker: true, heartbeatIntervalMs: 25000, heartbeatCallback }`
- Repro requires no network and no browser — pure `RealtimeClient` + fake transport/Worker, as above

vjymisal0:
Closes #2625.

### Problem
When a Realtime client is configured in worker mode (`worker: true`), `autoSendHeartbeat` is disabled on the underlying Phoenix `Socket` (`autoSendHeartbeat = false`) because heartbeats are driven by the worker instead of the socket's internal `setInterval`.

In Phoenix, `onConnOpen()` gates `resetHeartbeat()` behind `if (this.autoSendHeartbeat)`. Because `autoSendHeartbeat` is false in worker mode, `resetHeartbeat()` is never called when the socket opens or reconnects. As a result, if a connection drops with a heartbeat in-flight (e.g. Wi-Fi roam, sleep, server disconnect), the stranded `pendingHeartbeatRef` survives onto the next healthy connection. The next heartbeat triggered by the worker on the new connection sees a non-null `pendingHeartbeatRef`, incorrectly enters the timeout branch, and force-closes the new socket with `"heartbeat timeout"` after having sent 0 frames.

### Solution
1. Added a setter for `pendingHeartbeatRef` to `SocketAdapter` and aligned `RealtimeClient.pendingHeartbeatRef` to delegate to the underlying socket adapter.
2. In `RealtimeClient._setupConnectionHandlers()`, unconditionally clear `pendingHeartbeatRef = null` on both socket `onOpen` and `onClose` events. This ensures that any pending heartbeat ref left in flight from an earlier connection is always reset before frames are sent on the new connection.
3. Added unit tests in `test/phoenix/socketAdapter.test.ts` and lifecycle tests in `test/RealtimeClient.lifecycle.test.ts` verifying that `pendingHeartbeatRef` is properly cleared across connection lifecycle transitions.

### Verification
- Ran vitest across all 26 test suites in `@supabase/realtime-js` (467 tests passed).
- Built packages cleanly via `pnpm --filter @supabase/realtime-js build`.

### DIFF
--- a/packages/core/realtime-js/src/RealtimeClient.ts
+++ b/packages/core/realtime-js/src/RealtimeClient.ts
@@ -184,11 +184,13 @@ export default class RealtimeClient {
     return this.socketAdapter.heartbeatTimer
   }
 
-  get pendingHeartbeatRef() {
-    if (this.worker) {
-      return this._pendingWorkerHeartbeatRef
-    }
+  get pendingHeartbeatRef(): string | null {
     return this.socketAdapter.pendingHeartbeatRef
+  }
+
+  set pendingHeartbeatRef(ref: string | null) {
+    this.socketAdapter.pendingHeartbeatRef = ref
+    this._pendingWorkerHeartbeatRef = ref
   }
 
   get reconnectTimer(): Timer {
@@ -702,6 +704,8 @@ export default class RealtimeClient {
   /** @internal */
   private _setupConnectionHandlers(): void {
     this.socketAdapter.onOpen(() => {
+      this.pendingHeartbeatRef = null
+
       const authPromise =
         this._authPromise ||
         (this.accessToken && !this.accessTokenValue ? this.setAuth() : Promise.resolve())
@@ -715,6 +719,7 @@ export default class RealtimeClient {
       }
     })
     this.socketAdapter.onClose(() => {
+      this.pendingHeartbeatRef = null
       if (this.worker && this.workerRef) {
         this._terminateWorker()
       }
--- a/packages/core/realtime-js/src/phoenix/socketAdapter.ts
+++ b/packages/core/realtime-js/src/phoenix/socketAdapter.ts
@@ -55,6 +55,10 @@ export default class SocketAdapter {
     return this.socket.pendingHeartbeatRef
   }
 
+  set pendingHeartbeatRef(ref: string | null) {
+    this.socket.pendingHeartbeatRef = ref
+  }
+
   get reconnectTimer(): Timer {
     return this.socket.reconnectTimer
   }
--- a/packages/core/realtime-js/test/RealtimeClient.lifecycle.test.ts
+++ b/packages/core/realtime-js/test/RealtimeClient.lifecycle.test.ts
@@ -316,4 +316,31 @@ describe('Race condition prevention', () => {
     assert.equal(client.isDisconnecting(), true)
     cleanup()
   })
+
+  test('clears pendingHeartbeatRef on socket open and close', async () => {
+    const { client, connect, socketConnected, cleanup } = setupRealtimeTest()
+
+    connect()
+    await socketConnected()
+
+    // Simulate a pending heartbeat ref left in flight
+    client.pendingHeartbeatRef = 'stale-ref-from-previous-conn'
+    expect(client.pendingHeartbeatRef).toBe('stale-ref-from-previous-conn')
+
+    // Simulate onOpen event (e.g. reconnect)
+    // @ts-ignore
+    client.socketAdapter.getSocket().triggerStateCallbacks('open')
+    expect(client.pendingHeartbeatRef).toBeNull()
+
+    // Simulate pending heartbeat ref again
+    client.pendingHeartbeatRef = 'another-pending-ref'
+    expect(client.pendingHeartbeatRef).toBe('another-pending-ref')
+
+    // Simulate onClose event (e.g. disconnect)
+    // @ts-ignore
+    client.socketAdapter.getSocket().triggerStateCallbacks('close')
+    expect(client.pendingHeartbeatRef).toBeNull()
+
+    cleanup()
+  })
 })
--- a/packages/core/realtime-js/test/RealtimeClient.lifecycle.test.ts
+++ b/packages/core/realtime-js/test/RealtimeClient.lifecycle.test.ts
@@ -328,7 +328,7 @@ describe('Race condition prevention', () => {
     expect(client.pendingHeartbeatRef).toBe('stale-ref-from-previous-conn')
 
     // Simulate onOpen event (e.g. reconnect)
-    // @ts-ignore
+    // @ts-expect-error - triggering private phoenix socket callback in tests
     client.socketAdapter.getSocket().triggerStateCallbacks('open')
     expect(client.pendingHeartbeatRef).toBeNull()
 
@@ -337,7 +337,7 @@ describe('Race condition prevention', () => {
     expect(client.pendingHeartbeatRef).toBe('another-pending-ref')
 
     // Simulate onClose event (e.g. disconnect)
-    // @ts-ignore
+    // @ts-expect-error - triggering private phoenix socket callback in tests
     client.socketAdapter.getSocket().triggerStateCallbacks('close')
     expect(client.pendingHeartbeatRef).toBeNull()
 
--- a/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
+++ b/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
@@ -2,6 +2,7 @@ import { beforeAll, afterAll, beforeEach, afterEach, test, expect, vi, describe 
 import { type TestSetup, setupRealtimeTest } from './helpers/setup'
 import Worker from 'web-worker'
 import path from 'path'
+import { pathToFileURL } from 'url'
 import RealtimeClient from '../src/RealtimeClient'
 
 let testSetup: TestSetup
@@ -10,7 +11,7 @@ beforeAll(() => {
   window.Worker = Worker
 })
 
-const workerUrl = path.join(__dirname, '/helpers/test_worker.js')
+const workerUrl = pathToFileURL(path.join(__dirname, '/helpers/test_worker.js')).href
 
 beforeEach(() => {
   testSetup = setupRealtimeTest({
@@ -36,7 +37,7 @@ test('sets worker URL', () => {
 })
 
 describe('when no workerUrl provided', () => {
-  const mockObjectURL = `file://${workerUrl}`
+  const mockObjectURL = workerUrl.startsWith('file://') ? workerUrl : `file://${workerUrl}`
   let originalCreateObjectURL: any
 
   beforeAll(() => {
--- a/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
+++ b/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
@@ -116,3 +116,39 @@ test('terminates worker on disconnect', async () => {
   expect(spy).toHaveBeenCalled()
   expect(testSetup.client.workerRef).toBeFalsy()
 })
+
+test('worker reconnect: in-flight heartbeat on socket A does not cause heartbeat timeout teardown on socket B after reconnect', async () => {
+  const heartbeatEvents: string[] = []
+  testSetup.cleanup()
+  testSetup = setupRealtimeTest({
+    worker: true,
+    workerUrl,
+    heartbeatCallback: (status: string) => {
+      heartbeatEvents.push(status)
+    },
+  })
+
+  testSetup.connect()
+  await testSetup.socketConnected()
+
+  // 1. Send first heartbeat on socket A
+  testSetup.client.sendHeartbeat()
+  expect(heartbeatEvents).toContain('sent')
+
+  // 2. Disconnect socket A while heartbeat was in flight (without reply or timeout)
+  await testSetup.disconnect()
+  await testSetup.socketClosed()
+
+  // 3. Reconnect on socket B
+  testSetup.connect()
+  await testSetup.socketConnected()
+
+  // 4. Send heartbeat on socket B - should send cleanly without teardown
+  expect(testSetup.client.pendingHeartbeatRef).toBeNull()
+  testSetup.client.sendHeartbeat()
+
+  // Verify connection remains open and exactly two sent events occurred
+  expect(testSetup.client.isConnected()).toBe(true)
+  expect(heartbeatEvents.filter((s) => s === 'sent').length).toBe(2)
+})
+
--- a/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
+++ b/packages/core/realtime-js/test/RealtimeClient.worker.test.ts
@@ -81,7 +81,7 @@ test('ensures single worker ref is started even with multiple connect calls', as
   await testSetup.socketConnected()
   const ref = testSetup.client.workerRef
 
-  // @ts-ignore - simulate another onOpen call
+  // @ts-expect-error - simulate another onOpen call
   testSetup.client.socketAdapter.getSocket().triggerStateCallbacks('open')
 
   expect(testSetup.client.workerRef).toBe(ref)
@@ -90,7 +90,7 @@ test('ensures single worker ref is started even with multiple connect calls', as
 test('throws error when Web Worker is not supported', () => {
   // Temporarily remove Worker from window
   const originalWorker = window.Worker
-  // @ts-ignore - Deliberately setting to undefined to test error case
+  // @ts-expect-error - Deliberately setting to undefined to test error case
   window.Worker = undefined
 
   expect(() => {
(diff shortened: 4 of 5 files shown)


UgaTheDev:
Independently reproduced the analysis against the published sources. Confirming the
mechanism, with the exact lines, since the stranding step isn't spelled out in the issue:

`@supabase/phoenix@0.4.5`, `assets/js/phoenix/socket.js`:

- `onConnClose()` (L547-556) calls `clearHeartbeats()` (L451-454) on every close.
  `clearHeartbeats()` only clears `heartbeatTimer` and `heartbeatTimeoutTimer` — it
  does **not** touch `pendingHeartbeatRef`. That is the stranding step: cancelling
  `heartbeatTimeoutTimer` is exactly what prevents `heartbeatTimeout()` (L473-487, the
  other place the ref gets cleared) from ever running for the in-flight beat.
- `onConnOpen()` (L456-466) gates `resetHeartbeat()` behind `if (this.autoSendHeartbeat)`
  at L463, and `resetHeartbeat()` (L489-494) is the only open-path clear of the ref.
- `RealtimeClient.ts:863` sets `result.autoSendHeartbeat = !this.worker`, so the gate is
  false for every `worker: true` client.
- On the next socket, `sendHeartbeat()` (L664) sees a truthy `pendingHeartbeatRef` at
  L673 and calls `heartbeatTimeout()` → `teardown(..., "heartbeat timeout")` before
  `push()` at L679 is ever reached. Zero frames sent, as reported.

The window is narrow but ordinary: a beat is sent and the socket drops before either the
reply or `heartbeatTimeoutTimer` fires. A Wi-Fi roam or laptop sleep is precisely that.

Two notes on this PR:

**1. `heartbeatSentAt` doesn't need clearing.** The issue's proposed patch also nulls
`heartbeatSentAt`; this PR doesn't, and I think the PR is right to skip it. A stale value
can only be read at L702 (`latency = this.heartbeatSentAt ? ... : undefined`), which is
reached only when an incoming ref matches `pendingHeartbeatRef` — and `sendHeartbeat()`
reassigns `heartbeatSentAt` at L678 on the same line it sets a new ref. So there is no
path where a stale `heartbeatSentAt` is observable. Worth stating explicitly so it doesn't
get added later as a cargo-culted line.

**2. Layer choice is worth a maintainer decision.** The root cause is the
`autoSendHeartbeat` gate in phoenix's `onConnOpen`, which conflates two responsibilities:
clearing connection-local heartbeat state (correct unconditionally) and arming the
auto-send timer (correctly gated). Fixing it in `RealtimeClient._setupConnectionHandlers`
leaves the gate itself intact, so any future `@supabase/phoenix` consumer passing
`autoSendHeartbeat: false` reintroduces the bug. The counter-argument is real though —
the realtime-js fix ships without a `@supabase/phoenix` release and version bump, and
realtime-js is effectively the only consumer of that fork. Flagging it as a tradeoff, not
a defect: if maintainers want the gate split upstream in `supabase/phoenix` as well, I'm
happy to open that PR.

Also worth confirming in review that the fix covers `_handleNodeJsRaceCondition`
(`RealtimeClient.ts:731-734`), which calls `socketAdapter.getSocket().onConnOpen()`
directly. It does — phoenix's `onConnOpen` ends in `triggerStateCallbacks("open")`, so the
adapter's `onOpen` handler fires on that path too — but it's the one path where the clear
is reached indirectly and is easy to regress.

Cypher-Aura-19:
Nice work getting this into the RealtimeClient lifecycle. One thing from the production confirmation on #2625 that may still be worth covering: InkHubHQ mentioned clearing heartbeatSentAt alongside pendingHeartbeatRef, since otherwise the first heartbeat after reconnect can report latency against the previous connection's timestamp.

Also, would it be worth adding the worker reconnect regression from the original repro rather than only asserting the ref is null? Specifically: heartbeat on A → disconnect/reconnect → first worker heartbeat on B, then assert ['sent', 'sent'], exactly one heartbeat frame on B, and no "heartbeat timeout" close. That would exercise the actual failure path from #2625.

UgaTheDev:
Agreed on the second half, but I'd still leave `heartbeatSentAt` alone, and the InkHubHQ
report on #2625 doesn't change that — the latency it describes can't actually be reported.

`heartbeatSentAt` is read in exactly one place, `@supabase/phoenix@0.4.5`
`assets/js/phoenix/socket.js:702`:

```js
latency = this.heartbeatSentAt ? Date.now() - this.heartbeatSentAt : undefined
```

That line is inside the branch guarded by the incoming ref matching `pendingHeartbeatRef`.
So to read a stale `heartbeatSentAt` you need a reply whose ref equals the current
`pendingHeartbeatRef` — and `sendHeartbeat()` assigns both on adjacent lines (L678):
the new ref and a fresh `heartbeatSentAt` are always written together. There is no
interleaving where the ref is current and the timestamp is from the previous connection.

After this PR the ordering makes it doubly unreachable: `pendingHeartbeatRef` is null on
open, so the first post-reconnect beat sets both fields before any reply can match.

That's why I'd skip it — not because it's harmless, but because adding it implies a
relationship between the two fields that doesn't exist, and the next person reading it has
to re-derive why. If the concern is that a future refactor decouples the two assignments,
the durable fix is a comment at L678 saying they must move together, not a defensive null
in the close path.

On the regression test: yes, and stronger than what's there. `expect(ref).toBeNull()`
asserts the mechanism, not the symptom — flip the fix to clear some other field and the
test still tells you nothing about whether a heartbeat goes out. The sequence you describe
is the right one, and the load-bearing assertion is the third:

1. `worker: true` client, heartbeat sent on socket A (`pendingHeartbeatRef` non-null).
2. Close A without a reply and without letting `heartbeatTimeoutTimer` fire — that's the
   window `clearHeartbeats()` (L451-454) leaves behind, since it clears the timers but not
   the ref.
3. On socket B, assert exactly one heartbeat frame is pushed, and that no
   `"heartbeat timeout"` teardown occurs.

Step 3 is what fails on `main` today: `sendHeartbeat()` sees the truthy ref at L673 and
calls `heartbeatTimeout()` → `teardown(..., "heartbeat timeout")` before reaching `push()`
at L679. Zero frames, which is the reported symptom.

Worth covering `_handleNodeJsRaceCondition` (`RealtimeClient.ts:731-734`) in the same test
if it's cheap — it calls `socketAdapter.getSocket().onConnOpen()` directly, so it reaches
the clear indirectly via `triggerStateCallbacks("open")` rather than through the normal
open path. It works today; it's the one path a refactor could quietly break.

Cypher-Aura-19:
Good catch on heartbeatSentAt — agreed, if the ref and timestamp are always refreshed together then clearing it separately isn't necessary.

+1 on testing the actual reconnect symptom rather than just the internal ref state. Covering _handleNodeJsRaceCondition too makes sense if it can fit cleanly into the same regression test.

vjymisal0:
Thanks @Cypher-Aura-19 and @UgaTheDev for the analysis! I've pushed an update adding the worker reconnect regression test in \packages/core/realtime-js/test/RealtimeClient.worker.test.ts\ that exercises the in-flight heartbeat sequence across disconnect/reconnect and asserts that the subsequent heartbeat succeeds without a timeout teardown.

UgaTheDev:
Thanks for adding the worker test. I ran it on `dfeb886c` and it fails as written, and it fails the same way with the source fix reverted, so it does not yet distinguish fixed from unfixed. Two separate causes, both confirmed by tracing the connection state through the sequence:

1. `socketConnected()` resolves on the mock server's `connection` event (`waitForHaveBeenCalled(emitters.connected)` in `test/helpers/setup.ts`), which fires before the client's own `onopen`. On the second `connect()` the client reports `isConnected() === false` at that point, so `sendHeartbeat()` is a no-op (no `sent` event), line 151 fails, and the connection only becomes open ~50ms later.
2. CodeRabbit's point holds: the helper auto-replies to every `heartbeat`, so socket A's heartbeat gets its `ok` before `disconnect()` runs and `pendingHeartbeatRef` is already cleared on the unfixed code too. Line 147 passes without the fix for that reason.

This variant passes with the fix and fails without it, at the real symptom (`isConnected()` goes false right after the second `sendHeartbeat()` because the stale ref triggers the heartbeat-timeout teardown on socket B):

```ts
test('worker reconnect: in-flight heartbeat on socket A does not tear down socket B', async () => {
  const heartbeatEvents: string[] = []
  let heartbeatsSeen = 0
  testSetup.cleanup()
  testSetup = setupRealtimeTest({
    worker: true,
    workerUrl,
    heartbeatCallback: (status: string) => { heartbeatEvents.push(status) },
    socketHandlers: {
      heartbeat: (socket, message) => {
        heartbeatsSeen += 1
        if (heartbeatsSeen === 1) return // leave socket A's heartbeat in flight
        const msg = JSON.parse(message as string)
        socket.send(JSON.stringify({ topic: msg.topic, event: 'phx_reply', ref: msg.ref, payload: { status: 'ok', response: {} } }))
      },
    },
  })

  testSetup.connect()
  await vi.waitFor(() => expect(testSetup.client.isConnected()).toBe(true))
  testSetup.client.sendHeartbeat()
  expect(heartbeatEvents).toEqual(['sent'])

  await testSetup.disconnect()
  await testSetup.socketClosed()

  testSetup.connect()
  await vi.waitFor(() => expect(testSetup.client.isConnected()).toBe(true))
  expect(testSetup.client.pendingHeartbeatRef).toBeNull()
  testSetup.client.sendHeartbeat()

  expect(testSetup.client.isConnected()).toBe(true)
  expect(heartbeatEvents.filter((s) => s === 'sent').length).toBe(2)
  expect(heartbeatEvents).not.toContain('timeout')
  await vi.waitFor(() => expect(heartbeatEvents).toContain('ok'))
  expect(testSetup.client.isConnected()).toBe(true)
})
```

Run with `pnpm vitest run test/RealtimeClient.worker.test.ts -t "worker reconnect"` in `packages/core/realtime-js`: 1 passed on the branch, 1 failed at the `isConnected()` line after the second heartbeat with `src/RealtimeClient.ts` and `socketAdapter.ts` reverted to the merge base. The rest of the diff still looks right to me.

vjymisal0:
Re-ran the worker file locally on the latest branch after the deterministic test update:\n\n`pnpm vitest run test/RealtimeClient.worker.test.ts`\n\nAll 9 tests pass, including the reconnect case. The earlier filtered run skipped everything because my test name filter didn't match, so I used the full worker file instead.

PR #2639, opened 2026-08-31
fix(auth): preserve verified passkey when webauthn register() enrollment fails

### LINKED ISSUE #2640: auth-js: mfa.webauthn.register() deletes an existing verified passkey when a re-registration fails
## Describe the bug

```ts
if (!factor) {
  await this.client.mfa
    .listFactors()
    .then((factors) =>
      factors.data?.all.find(
        (v) =>
      )
    )
}
```

## To Reproduce

## Expected behavior

## Environment

## Fix

pouyashahrdami:
Fixes #2640

## Description

`mfa.webauthn.register()` runs a cleanup step when its internal enrollment call fails. That cleanup looked up a webauthn factor matching the requested friendly name with `status !== 'unverified'` — i.e. a **verified** factor — and unenrolled it:

```ts
// packages/core/auth-js/src/lib/webauthn.ts
if (!factor) {
  await this.client.mfa
    .listFactors()
    .then((factors) =>
      factors.data?.all.find(
        (v) =>
          v.status !== 'unverified' // ← targets a VERIFIED factor
      )
    )
}
```

Enrollment commonly fails on a **friendly-name conflict** — precisely the case where a verified factor with that name already exists. So a user who retries registration with an existing name (or whose UI naively retries after a transient error) has their **existing, working passkey silently deleted**, while the call just returns the original enrollment error.

The sensible cleanup semantics here are the opposite: remove a **stale unverified** factor left behind by a previous incomplete attempt so a retry can succeed, and never touch a verified factor. This PR flips the condition to `status === 'unverified'`.

## Type of Change

- [x] Bug fix (fix)
- [ ] New feature (feat)
- [ ] Breaking change
- [ ] Documentation update (docs)
- [ ] Other

## Testing

Adds `packages/core/auth-js/test/webauthn.register.test.ts` covering the failed-enrollment cleanup path:

- a **verified** factor is preserved when enrollment fails (the regression),
- a **stale unverified** factor is cleaned up so a retry can succeed,
- a verified factor sharing a name with a stale unverified one is left intact.

All three tests fail against the previous `!== 'unverified'` behavior and pass with the fix. `nx lint auth-js`, `nx build auth-js`, and `prettier --check` all pass.

## Checklist

- [x] Tests added and passing
- [x] Builds passing (`nx build auth-js`)
- [x] Used conventional commits

### DIFF
--- a/packages/core/auth-js/src/lib/webauthn.ts
+++ b/packages/core/auth-js/src/lib/webauthn.ts
@@ -892,6 +892,10 @@ export class WebAuthnApi {
       })
 
       if (!factor) {
+        // Enrollment failed (commonly a friendly-name conflict). Clean up only a
+        // stale *unverified* factor left by a previous incomplete attempt so a
+        // retry can succeed. Never touch a verified factor: deleting the user's
+        // working passkey because a new enrollment failed would be data loss.
         await this.client.mfa
           .listFactors()
           .then((factors) =>
@@ -899,7 +903,7 @@ export class WebAuthnApi {
               (v) =>
                 v.factor_type === 'webauthn' &&
                 v.friendly_name === friendlyName &&
-                v.status !== 'unverified'
+                v.status === 'unverified'
             )
           )
           .then((factor) => (factor ? this.client.mfa.unenroll({ factorId: factor?.id }) : void 0))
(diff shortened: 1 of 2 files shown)


mandarini:
Hi @pouyashahrdami, thank you for contributing to Supabase! 💚 

Closing this in favour of #2641, which makes the same one character fix but puts its tests in the existing `passkey.methods.test.ts`, reusing the browser and client helpers already there instead of standing up a second setup.

Nothing wrong with your diff, it came down to test placement. The report in #2640 was yours and the trace you wrote made this quick to confirm, so credit for the fix is yours either way. Thanks again!

PR #2641, opened 2026-08-31
fix(auth): unenroll the unverified factor after a failed registration

### LINKED ISSUE #2640: auth-js: mfa.webauthn.register() deletes an existing verified passkey when a re-registration fails

Kjubikstronk:
Closes #2640.

When `_enroll` fails, `register()` tries to clean up after itself by finding a webauthn factor with the same friendly name and unenrolling it. The lookup selects on `v.status !== 'unverified'`, which matches a **verified** factor, so a failed re-registration can delete a passkey the user already had working.

The factor worth cleaning up is the unverified one the failed attempt just left behind, so the condition is inverted. That is the whole change:

```diff
-                v.status !== 'unverified'
+                v.status === 'unverified'
```

Two tests in `passkey.methods.test.ts`, which mocks fetch throughout and needs no server. One drives a failed enrollment with an existing verified factor of the same name and asserts `unenroll` is never called; the other does the same with an unverified factor and asserts it is unenrolled by id. Both fail on `master` and pass with the change, and the other 29 tests in that file are unaffected.

`prettier --check` is clean on both files.

### DIFF
--- a/packages/core/auth-js/src/lib/webauthn.ts
+++ b/packages/core/auth-js/src/lib/webauthn.ts
@@ -895,6 +895,10 @@ export class WebAuthnApi {
         await this.client.mfa
           .listFactors()
           .then((factors) =>
+            // Only an unverified factor is leftover from the enrollment that just
+            // failed, and removing it is what lets a retry with the same name
+            // succeed. A verified factor of that name is a passkey the user is
+            // already using, so it must never be selected here.
             factors.data?.all.find(
               (v) =>
                 v.factor_type === 'webauthn' &&
--- a/packages/core/auth-js/src/lib/webauthn.ts
+++ b/packages/core/auth-js/src/lib/webauthn.ts
@@ -899,7 +899,7 @@ export class WebAuthnApi {
               (v) =>
                 v.factor_type === 'webauthn' &&
                 v.friendly_name === friendlyName &&
-                v.status !== 'unverified'
+                v.status === 'unverified'
             )
           )
           .then((factor) => (factor ? this.client.mfa.unenroll({ factorId: factor?.id }) : void 0))
--- a/packages/core/auth-js/test/passkey.methods.test.ts
+++ b/packages/core/auth-js/test/passkey.methods.test.ts
@@ -697,3 +697,61 @@ describe('admin.passkey', () => {
     expect((error as AuthApiError).code).toEqual('passkey_not_found')
   })
 })
+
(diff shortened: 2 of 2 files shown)


mandarini (changes requested):
Hi @Kjubikstronk, thank you for contributing to Supabase! 💚 

Fix is correct. `FactorVerificationStatuses` is a closed union of `verified` and `unverified`, so `!== 'unverified'` was exactly `=== 'verified'`. Good catch, and thanks for keeping the diff to just that.

Two things before merge:

1. **Add a comment above the changed line.** A one character flip is easy to revert later by mistake. State the invariant: cleanup targets a stale unverified factor so a retry can succeed, and must never select a verified one.

2. **Two more tests.** Your existing two are solid. Please also cover a verified and an unverified factor sharing a name (verified survives), and no matching factor at all (`unenroll` never called).

Yours is the one we are taking of the PRs for this issue, the tests belong in `passkey.methods.test.ts` where you put them. Thanks again!

Kjubikstronk:
Both done in 97a16eb.

**The comment.** Above the status check, stating it in terms of what each status means rather than what the operator is:

```ts
// Only an unverified factor is leftover from the enrollment that just
// failed, and removing it is what lets a retry with the same name
// succeed. A verified factor of that name is a passkey the user is
// already using, so it must never be selected here.
```

**The two tests.** Both in `passkey.methods.test.ts`, next to the existing pair. The helper now takes a list of factors so the shared name case can be expressed directly:

- `removes only the unverified one when both share a name`, asserting `unenroll` is called exactly once and with the unverified id, so the test fails if the search ever picks the first match by name alone.
- `does nothing when no factor of that name exists`, with `unenroll` never called.

I checked the new tests actually bite: reverting the character back to `!== 'unverified'` fails three of the four, and the fourth is the no factor case, which is a control that should pass either way.

`passkey.methods.test.ts` is green, 33 of 33.

mandarini (approved):

mandarini:
@Kjubikstronk i'll fix the `pnpm audit` issues on `master` and I'll ask you to rebase

mandarini:
@Kjubikstronk can you please rebase with `master`?

PR #2642, opened 2026-08-31
fix(auth): stop register() cleanup from unenrolling a verified webauthn factor

### LINKED ISSUE #2640: auth-js: mfa.webauthn.register() deletes an existing verified passkey when a re-registration fails

samuelmbabhazi:
### What changed?

The cleanup block that `register()` runs after a failed enrollment selected a webauthn factor with the requested friendly name whose `status !== 'unverified'`, that is, a **verified** factor, and unenrolled it. The condition is now `status === 'unverified'`, so the cleanup removes only an unverified leftover from a previous failed attempt, and a comment states that intent. Three unit tests pin the behaviour down: a verified factor with the same friendly name is never unenrolled, only the unverified leftover is (when both coexist, exactly the leftover is removed), and factors with other names or types are untouched.

Enrollment most commonly fails on a friendly name conflict, which is precisely the situation where a verified factor with that name already exists. A user retrying `register()` with the same name therefore had their active passkey deleted by the failed attempt's cleanup. See the reproduction and analysis in the linked issue.

Closes #2640

## 📸 Screenshots/Examples

Not applicable.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

Verified locally on the auth-js package: the new tests fail against the previous condition and pass with this one, the full unit suite passes (240 tests), and lint, build and docs targets are green. The docker backed integration suites were not run.

### DIFF
--- a/packages/core/auth-js/src/lib/webauthn.ts
+++ b/packages/core/auth-js/src/lib/webauthn.ts
@@ -892,6 +892,9 @@ export class WebAuthnApi {
       })
 
       if (!factor) {
+        // Clean up only an unverified leftover from a previous failed attempt so the
+        // user can retry; a verified factor with this name is an active credential
+        // and must never be deleted here.
         await this.client.mfa
           .listFactors()
           .then((factors) =>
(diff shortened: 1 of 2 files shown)


samuelmbabhazi:
Closing this one: I had missed that #2639, from the issue author, predates it and covers the same fix. Deferring to it.

PR #2657, opened 2026-09-02
fix(postgrest): quote array literal elements in contains, overlaps and the like filters

### LINKED ISSUE #2669: postgrest-js: contains/overlaps and the likeAllOf family do not quote array literal elements, so a comma inside a value silently returns the wrong rows
## Describe the bug

A `{...}` filter value is a Postgres **array literal**, not a PostgREST list, and it has its own quoting rules. `in()` and `notIn()` quote their values for the `(...)` list context via `PostgrestReservedCharsRegexp`, but nothing covers the `{...}` context. Seven builders join their elements with a bare comma:

`contains`, `containedBy`, `overlaps`, `likeAllOf`, `likeAnyOf`, `ilikeAllOf`, `ilikeAnyOf`.

An element that itself contains a comma is therefore split by Postgres into two elements. The query does not fail — it answers a different question.

## To Reproduce

Table with two rows: id 1 holding `ARRAY['a,b']` (one element), id 2 holding `ARRAY['a','b']` (two elements).

```ts
await supabase.from('t').select('id').contains('tags', ['a,b'])
// URL:      tags=cs.{a,b}
// expected: id 1
// actual:   id 2
```

Measured on PostgREST 12 / Postgres 15:

| call | URL produced | actual result |
|---|---|---|
| `contains('tags', ['a,b'])` | `tags=cs.{a,b}` | id 2 |
| `overlaps('tags', ['a,b'])` | `tags=ov.{a,b}` | id 2 |
| `likeAllOf('name', ['*x,y*'])` | `name=like(all).{*x,y*}` | `[]` |
| `contains('tags', ['a{b'])` | `tags=cs.{a{b}` | error `22P02` malformed array literal |

The parsing rules, checked directly in Postgres 15:

```
'{a,b}'::text[]      -> 2 elements: a | b
'{"a,b"}'::text[]    -> 1 element:  a,b
'{a{b}'::text[]      -> ERROR: malformed array literal
'{NULL}'::text[]     -> SQL NULL, not the text 'NULL'
'{ a }'::text[]      -> 'a', surrounding spaces trimmed
```

## Expected behavior

An element containing a comma, brace, quote, backslash, leading/trailing whitespace, or the empty string should be quoted so Postgres receives the element the caller passed. `contains('tags', ['a,b'])` should return id 1.

## Why it matters

The first three rows above are the dangerous ones: no error is raised, and the caller gets rows that merely contain two different elements. A plausible wrong answer is harder to catch than a failure, and comma-bearing tags (`"Last, First"`, `"a,b"`) are ordinary user data.

## System information

- `@supabase/postgrest-js` on `master`
- PostgREST 12, Postgres 15

PR #2657 is open against this, with 12 assertions covering the seven builders and the cases that must not change (plain values, non-strings, the range and json branches of `contains`/`containedBy`/`overlaps`). 10 of the 12 fail on `master`.

PedroHenrique0713:
Fixes #2669

## The bug

A `{...}` filter value is a **Postgres array literal**, not a PostgREST list, so it needs its own quoting rules. `in()` and `notIn()` already quote for the `(...)` list context via `PostgrestReservedCharsRegexp`, but nothing covered `{...}`. Seven builders join their values with a bare comma:

`contains`, `containedBy`, `overlaps`, `likeAllOf`, `likeAnyOf`, `ilikeAllOf`, `ilikeAnyOf`.

## Measured, not read

PostgREST 12 over Postgres 15, on a table holding both `ARRAY['a,b']` (one element, id 1) and `ARRAY['a','b']` (two elements, id 2):

| call | URL produced | result on `master` |
|---|---|---|
| `contains('tags', ['a,b'])` | `tags=cs.{a,b}` | **id 2** — the two-element row |
| `overlaps('tags', ['a,b'])` | `tags=ov.{a,b}` | **id 2** |
| `likeAllOf('name', ['*x,y*'])` | `name=like(all).{*x,y*}` | **`[]`** |
| `contains('tags', ['a{b'])` | `tags=cs.{a{b}` | **`22P02` malformed array literal** |

With the quoting applied, all four return what the caller asked for (`cs.{"a,b"}` → id 1, and the brace no longer errors).

The first three are the dangerous ones: the caller asks for one element and gets rows that merely contain two different elements. A plausible wrong answer, not an error.

The parsing rules behind this, checked directly in Postgres 15:

```
'{a,b}'::text[]      -> 2 elements: a | b
'{"a,b"}'::text[]    -> 1 element:  a,b
'{a{b}'::text[]      -> ERROR: malformed array literal
'{NULL}'::text[]     -> SQL NULL, not the text 'NULL'
'{ a }'::text[]      -> 'a', the surrounding spaces are trimmed
```

## The fix

Quote an element only where Postgres needs it: a comma, a brace, a quote, a backslash, whitespace that would otherwise be trimmed, or the empty string (which `{}` drops entirely). Inside a quoted element, `"` and `\` are escaped.

Non-strings stay bare on purpose, so `null` reaches Postgres as SQL NULL rather than the text `'null'`. The range (`cs.[...)`) and json (`cs.{"a":...}`) branches of `contains`/`containedBy`/`overlaps` are untouched.

## Verification

12 new assertions in `test/array-literal-quoting.test.ts`, covering each of the seven builders plus the cases that must **not** change (plain values, non-strings, range and json forms).

- **10 of the 12 fail on `master`.**
- End to end against the PostgREST instance above, using the real client: without the change `contains(['a,b'])` returns id 2, with it returns id 1.

Comparing the full suite before and after by failing test **name**, the only difference is those 10 tests moving from failing to passing — no new failure. Locally the rest of the suite needs `supabase start`, which I did not run, so the integration specs are down to CI here. `tsc --noEmit` and `prettier --check` are clean.

### DIFF
--- a/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
+++ b/packages/core/postgrest-js/src/PostgrestFilterBuilder.ts
@@ -35,6 +35,29 @@ export type IsStringOperator<Path extends string> = Path extends `${string}->>${
 
 const PostgrestReservedCharsRegexp = new RegExp('[,()]')
 
+// A `{...}` filter value is a Postgres array literal, not a PostgREST list, so
+// it has its own set of characters that end an element. Whitespace is included
+// because Postgres trims it off an unquoted element, and the empty string
+// because `{}` is an empty array rather than an array holding one empty value.
+const PostgrestArrayLiteralReservedCharsRegexp = /[,{}"\\\s]|^$/
+
+/**
+ * Quotes an element of a Postgres array literal when it carries a character
+ * that would otherwise change how Postgres parses the literal.
+ *
+ * Only strings are quoted: numbers, booleans and `null` have to reach Postgres
+ * bare so that `null` stays SQL NULL instead of becoming the text `'null'`.
+ */
+const quoteArrayLiteralElement = (value: unknown): string => {
+  if (typeof value !== 'string') return `${value}`
+  if (!PostgrestArrayLiteralReservedCharsRegexp.test(value)) return value
+
+  return `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`
+}
+
+const toArrayLiteral = (values: readonly unknown[]): string =>
+  values.map(quoteArrayLiteralElement).join(',')
+
 // Match relationship filters with `table.column` syntax and resolve underlying
 // column value. If not matched, fallback to generic type.
 // TODO: Validate the relationship itself ala select-query-parser. Currently we
@@ -548,7 +571,7 @@ export default class PostgrestFilterBuilder<
   ): this
   likeAllOf(column: string, patterns: readonly string[]): this
   likeAllOf(column: string, patterns: readonly string[]): this {
-    this.url.searchParams.append(column, `like(all).{${patterns.join(',')}}`)
+    this.url.searchParams.append(column, `like(all).{${toArrayLiteral(patterns)}}`)
     return this
   }
 
@@ -567,7 +590,7 @@ export default class PostgrestFilterBuilder<
   ): this
   likeAnyOf(column: string, patterns: readonly string[]): this
   likeAnyOf(column: string, patterns: readonly string[]): this {
-    this.url.searchParams.append(column, `like(any).{${patterns.join(',')}}`)
+    this.url.searchParams.append(column, `like(any).{${toArrayLiteral(patterns)}}`)
     return this
   }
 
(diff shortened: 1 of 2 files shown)


PedroHenrique0713:
Confirmed, and measured on Postgres 15 before changing anything:

```
'{null}'::text[]    -> SQL NULL
'{NULL}'::text[]    -> SQL NULL
'{NuLl}'::text[]    -> SQL NULL
'{"null"}'::text[]  -> the text 'null'
'{nulls}'::text[]   -> the text 'nulls'
```

So any casing of the bare token is SQL NULL, while a value that merely starts with it is not, which keeps the check to the exact token.

End to end against PostgREST, on a table holding both `ARRAY['null']` and `ARRAY[NULL]`, `contains('tags', ['null'])` returned `[]` before and returns the text row now.

Pushed with `test.each(['null', 'NULL', 'NuLl'])` plus a case asserting that `nulls` and `null2` stay unquoted. A real `null` value still goes out bare, since only strings reach the quoting. 16 assertions in the file, all passing, and comparing the full suite by failing test name against master shows no new failure.

PR #2663, opened 2026-09-03
fix(supabase): warn when schema is passed outside db options

### LINKED ISSUE #969: Selecting a schema on createClient does not work

mandarini:
Warns once per process when `schema` is passed at the top level of the client options instead of under `db`. That shape (the v1 API) is not part of `SupabaseClientOptions` and is silently ignored today, so queries go to the default schema and the only signal is a confusing PostgREST "relation does not exist" error downstream.

The check mirrors `checkApiKeyFormat`: `console.warn`, never throws, no type changes, so nothing that compiles or runs today changes behavior. Throwing would crash any app that currently runs with a stray top-level `schema` (including configs spread from a shared object), which is a breaking change on 2.x; the warning lands next to the PostgREST error that already surfaces the problem, so the failure is explained rather than silenced. A `schema?: never` compile-time guard is worth its own PR, since it turns compiling code into non-compiling code.

Closes #969 — this covers the original report (wrong option shape). The other half of that thread, the `createClient<Database>` second-generic requirement, is documented in #2662 and supabase/supabase#49967.

### DIFF
--- a/packages/core/supabase-js/src/lib/helpers.ts
+++ b/packages/core/supabase-js/src/lib/helpers.ts
@@ -21,6 +21,42 @@ export function ensureTrailingSlash(url: string): string {
 
 export const isBrowser = () => typeof window !== 'undefined'
 
+let warnedTopLevelSchema = false
+
+/**
+ * Warn (once per process) when `schema` is passed at the top level of the client options
+ * instead of under `db`. A top-level `schema` is not part of the options shape and is
+ * ignored, so queries silently go to the default schema. Never throws.
+ */
+export function checkTopLevelSchemaOption(options: unknown): void {
+  if (warnedTopLevelSchema) {
+    return
+  }
+  if (
+    typeof options !== 'object' ||
+    options === null ||
+    !('schema' in options) ||
+    (options as { schema?: unknown }).schema === undefined
+  ) {
+    return
+  }
+  warnedTopLevelSchema = true
+  console.warn(
+    '@supabase/supabase-js: The "schema" option must be nested under "db", ' +
+      "e.g. createClient(url, key, { db: { schema: 'myschema' } }). " +
+      'A top-level "schema" is ignored and queries go to the default schema.'
+  )
+}
+
+/**
+ * For tests only. Resets the one-time top-level `schema` warning.
+ *
+ * @internal
+ */
+export function _resetTopLevelSchemaWarning(): void {
+  warnedTopLevelSchema = false
+}
+
 export type ResolvedSupabaseClientOptions<SchemaName> = Omit<
   Required<SupabaseClientOptions<SchemaName>>,
   'tracePropagation'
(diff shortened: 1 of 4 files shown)


packages/core/supabase-js/test/unit/helpers.test.ts
```diff
@@ -41,6 +41,49 @@ test('override setting defaults', async () => {
   expect(settings.db.schema).toBe(defaults.db.schema)
 })
 
+describe('checkTopLevelSchemaOption', () => {
+  let warnSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    helpers._resetTopLevelSchemaWarning()
+    warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {})
```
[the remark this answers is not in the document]
mandarini: Confirmed against the repo ESLint config and fixed on the lines this PR adds: this spy plus the two wiring tests in `SupabaseClient.test.ts`. The eleven pre-existing `() => {}` spies in these files hit the same rule and are left alone as out of scope.

packages/core/supabase-js/src/lib/helpers.ts
```diff
@@ -21,6 +21,42 @@ export function ensureTrailingSlash(url: string): string {
...
+ */
+export function checkTopLevelSchemaOption(options: unknown): void {
+  if (warnedTopLevelSchema) {
+    return
+  }
+  if (
+    typeof options !== 'object' ||
+    options === null ||
+    !('schema' in options) ||
+    (options as { schema?: unknown }).schema === undefined
```
QuintinWillison, line 42: Might be nothing or I might not understand what your goal is, but do you need to also consider other edge cases that could express the user's desire "not to set" this property? e.g. if a user explicitly writes `{ schema: null }`, `{ schema: false }`, or `{ schema: "" }`... in these cases they'll still get the warning logged.

QuintinWillison (dismissed):

packages/core/supabase-js/src/lib/helpers.ts
mandarini: Good question! The line is deliberately the one `postgrest-js` draws: in `PostgrestBuilder` the request path checks `this.schema === undefined` to skip the header, so `db: { schema: null } / '' / false` all send a broken `Accept-Profile` rather than falling back to the default. Only undefined means unset there, so only undefined is exempt here. A top-level `schema: null` is the same misplaced key with a value that would break in the right place too, so warning on it seems right. I'll note that in the JSDoc so it's explicit, just in case!! :D

QuintinWillison (approved):

PR #2666, opened 2026-09-04
fix(realtime): clear disconnect race timeout on successful close

Kamalpannu:
### What changed?

`SocketAdapter.disconnect()` starts a race timeout and discards the returned id:

```ts
return new Promise((resolve) => {
  setTimeout(() => resolve('timeout'), timeout)   // id not kept
  this.socket.disconnect(() => { callback(); resolve('ok') }, code, reason)
})
```

The returned value was never wrong on a clean close `resolve('ok')` settles the promise first, and the later `resolve('timeout')` is a no-op. But settling a promise does not cancel a `setTimeout`. The timer stays scheduled for the full `timeout` window, which defaults to 10000ms.

This PR keeps the id and clears it in the disconnect callback. The timeout branch needs no cleanup  if it fires, it has already run.

In Node a pending timer keeps the event loop alive. A process that awaits `client.disconnect()` and has nothing else to do therefore stays alive until the timer expires, and every `disconnect()` call schedules another one.

The impact is on Node scripts, test suites, CI jobs and serverless handlers, where a process that should exit immediately instead lingers. Browsers are unaffected there is no process to exit and long-running servers hold only one short-lived timer per disconnect. No public API behavior changes.

## 📸 Screenshots/Examples

Added a regression test in `test/phoenix/socketAdapter.test.ts`. On master it fails:

```
FAIL  test/phoenix/socketAdapter.test.ts > disconnect timer cleanup >
      does not leave the race timeout pending after a successful disconnect

AssertionError: expected 1 to be +0 // Object.is equality
- Expected
+ Received
- 0
+ 1

  71|     await socketAdapter.disconnect(() => {})
  72|
  73|     expect(vi.getTimerCount()).toBe(0)
    |                                ^
```

Worth flagging for reviewers: a naive before/after `getTimerCount()` comparison does **not** catch this. A heartbeat timer is cleared in the same tick, so the count goes 1 → 1 and the leak is hidden. Asserting that no timers remain at all is what exposes it.

Reviewers who want to see the event-loop effect directly can run any script that connects to a local WebSocket server, awaits `client.disconnect()`, and logs from `process.on('exit')` on master the exit is delayed by the timeout window; with this change it is immediate.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

Full `realtime-js` suite passes locally: 476 passed, 1 skipped.

Co-authored with Claude; I reproduced the failing test and verified the fix locally.

### DIFF
--- a/packages/core/realtime-js/src/phoenix/socketAdapter.ts
+++ b/packages/core/realtime-js/src/phoenix/socketAdapter.ts
@@ -94,9 +94,10 @@ export default class SocketAdapter {
     timeout: number = 10000
   ): Promise<'ok' | 'timeout'> {
     return new Promise((resolve) => {
-      setTimeout(() => resolve('timeout'), timeout)
+      const timer = setTimeout(() => resolve('timeout'), timeout)
       this.socket.disconnect(
         () => {
+          clearTimeout(timer)
           callback()
           resolve('ok')
         },
--- a/packages/core/realtime-js/test/phoenix/socketAdapter.test.ts
+++ b/packages/core/realtime-js/test/phoenix/socketAdapter.test.ts
@@ -62,3 +62,14 @@ describe('connection state', () => {
     expect(socketAdapter.isDisconnecting()).toBe(true)
   })
 })
+
+describe('disconnect timer cleanup', () => {
+  test('does not leave the race timeout pending after a successful disconnect', async () => {
+    testSetup.connect()
+    await testSetup.socketConnected()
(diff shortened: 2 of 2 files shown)


Kamalpannu:
Friendly ping the workflow runs are still pending approval, so CI hasn't run
and the required block-merge check can't report yet. Could a maintainer approve
the run when you get a chance? Happy to rebase if master has moved since.

PR #2671, opened 2026-09-06
fix(realtime): settle discarded push instead of leaving its promise pending

Kamalpannu:
### What changed?

When the push buffer is full, `ChannelAdapter.push()` drops the oldest push and calls `removedPush.cancelTimeout()`. That method only clears the timer:

```js
cancelTimeout(){
  clearTimeout(this.timeoutTimer)
  this.timeoutTimer = null
}
```

It fires no receive hook, so nothing tells the caller the push is gone. `RealtimeChannel.send()` resolves only from `receive('ok' | 'error' | 'timeout')`, so the discarded push's promise never settles.

This calls `removedPush.trigger('timeout', {})` instead. That routes through the reply handler registered in `startTimeout()`, which calls `cancelRefEvent()`, `cancelTimeout()` and `matchReceive()` so it does everything the old line did, plus removing the leftover `chan_reply_<ref>` binding and settling the promise.

`broadcast` pushes escape this: `send()` early-resolves them when `ack` is off. `presence` pushes do not so `await channel.track(...)` and `await channel.untrack(...)` hang forever if that push is the one discarded.

That is reachable in normal use: `track()` called rapidly before the channel has joined (a cursor-position channel, for example) fills the 100-entry buffer, and the earliest calls are silently dropped. The caller gets no error and no timeout — just a promise that never settles.

## 📸 Screenshots/Examples

Added a regression test. It subscribes a channel, leaves it unjoined so pushes buffer, overflows the cap, then advances timers well past the push timeout so every surviving push times out normally. On master:

pr2-screenshot1

```
AssertionError: expected false to be true // Object.is equality
```
pr2-screenshot2

The first push's promise is still pending; an in-buffer push under the same conditions settles as expected.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

Full `realtime-js` suite passes locally: 475 passed, 1 skipped.

### One open question

I used `trigger('timeout', {})` so the discarded push resolves as `'timed out'`. That reuses an existing status and adds no API surface, but the push wasn't slow it was dropped by the buffer cap so `'error'` may be the more honest signal. Happy to switch if you'd prefer; either way the promise settles instead of hanging.

### Unrelated, noting rather than changing

The cap check is `pushBuffer.length > MAX_PUSH_BUFFER_SIZE`, so the buffer reaches 101 before trimming. Left alone here to keep this PR to one concern.

Co-authored with Claude; I reproduced the failing test and verified the fix locally.

### DIFF
--- a/packages/core/realtime-js/src/phoenix/channelAdapter.ts
+++ b/packages/core/realtime-js/src/phoenix/channelAdapter.ts
@@ -84,7 +84,7 @@ export default class ChannelAdapter {
 
     if (this.channel.pushBuffer.length > MAX_PUSH_BUFFER_SIZE) {
       const removedPush = this.channel.pushBuffer.shift()!
-      removedPush.cancelTimeout()
+      removedPush.trigger('timeout', {})
       this.socket.log(
         'channel',
         `discarded push due to buffer overflow: ${removedPush.event}`,
--- a/packages/core/realtime-js/test/phoenix/channelAdapter.test.ts
+++ b/packages/core/realtime-js/test/phoenix/channelAdapter.test.ts
@@ -123,3 +123,36 @@ describe('updateJoinPayload', () => {
     })
   })
 })
+
+describe('push buffer overflow', () => {
+  let overflowSetup: TestSetup
+
+  beforeEach(async () => {
+    overflowSetup = setupRealtimeTest({ timeout: defaultTimeout, useFakeTimers: true })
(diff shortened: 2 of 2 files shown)


Kamalpannu:
Friendly ping the workflow runs are still pending approval, so CI hasn't run
and the required block-merge check can't report yet. Could a maintainer approve
the run when you get a chance? Happy to rebase if master has moved since.

PR #2644, opened 2026-08-31
feat(storage): object versioning updates

TylerHillery:
## 🔍 Description

Updates to support the new object versioning feature in storage.

### What changed?

Several new parameters were added. All changes are additive/optional:
- `Bucket` gains `versioning_status`
- `ListBucketOptions` response gains `version`, `archived_at`, `is_delete_marker`, `is_versioned` on `FileObject` (`list()`/`remove()`) and `SearchV2Object` (`listV2()`)
- `FileObjectV2` gains `archived_at`, `is_delete_marker`, `is_versioned`
- `SearchOptions`/`SearchV2Options` gain `noncurrentVersions`, `deleteMarkers`, `exactMatch` for `list()`/`listV2()`
- `DestinationOptions` (shared by `move()`/`copy()`) gains `sourceVersionId`
- New `DeleteObjectEntry` type. `remove()`'s `paths` param now accepts `(string | { path: string; versionId: string })[]` instead of `string[]`, to target an exact version for deletion
- New `VersioningStatus`, `CreateSettableVersioningStatus`, `UpdateSettableVersioningStatus` types: `createBucket()`/`updateBucket()` gain a `versioningStatus` option (`DISABLED`/`ENABLED` on create, `ENABLED`/`SUSPENDED` on update)
- `download()` and `info()` gain a `versionId` option to target a specific object version

PR #2678, opened 2026-09-07
feat(storage): add versionId support to create URL methods

TylerHillery:
## 🔍 Description

Follow up to #2644 forgot to also update the `createSignedUrl` and `getPublicUrl` methods to add the optional versonId to create a URL for specific object version

### DIFF
--- a/packages/core/storage-js/src/packages/StorageFileApi.ts
+++ b/packages/core/storage-js/src/packages/StorageFileApi.ts
@@ -643,6 +643,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
    * @param options.download triggers the file as a download if set to true. Set this parameter as the name of the file if you want to trigger the download with a different filename.
    * @param options.transform Transform the asset before serving it to the client.
    * @param options.cacheNonce Append a cache nonce parameter to the URL to invalidate the cache.
+   * @param options.versionId Create a signed URL for a specific object version rather than the current one.
    * @returns Promise with response containing signed URL or error
    *
    * @example Create Signed URL
@@ -699,6 +700,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       download?: string | boolean
       transform?: TransformOptions
       cacheNonce?: string
+      versionId?: string
     }
   ): Promise<
     | {
@@ -721,7 +723,11 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       let data = await post(
         this.fetch,
         `${this.url}/object/sign/${_path}`,
-        { expiresIn, ...(hasTransform ? { transform: options!.transform } : {}) },
+        {
+          expiresIn,
+          ...(hasTransform ? { transform: options!.transform } : {}),
+          ...(options?.versionId != null ? { versionId: options.versionId } : {}),
+        },
         { headers: this.headers }
       )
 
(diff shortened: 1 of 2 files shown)


packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -1112,6 +1120,7 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
     if (options?.download) query.set('download', options.download === true ? '' : options.download)
     if (options?.transform) this.applyTransformOptsToQuery(query, options.transform)
     if (options?.cacheNonce != null) query.set('cacheNonce', String(options.cacheNonce))
+    if (options?.versionId != null) query.set('versionId', String(options.versionId))
```
ferhatelmas, line 1123: would be nice to add a test that shows it's set when given

ferhatelmas: using existing pattern but check is a bit inconsistent with [create](https://github.com/supabase/supabase-js/pull/2678/changes#diff-bcad028f9e65ac415cf97b5fbd3e0e5267ea6299c479e187529c932d8c1ac091R729)

TylerHillery: addressed both

packages/core/storage-js/src/packages/StorageFileApi.ts
```diff
@@ -721,7 +723,11 @@ export default class StorageFileApi extends BaseApiClient<StorageError> {
       let data = await post(
         this.fetch,
         `${this.url}/object/sign/${_path}`,
-        { expiresIn, ...(hasTransform ? { transform: options!.transform } : {}) },
+        {
+          expiresIn,
+          ...(hasTransform ? { transform: options!.transform } : {}),
+          ...(options?.versionId != null ? { versionId: options.versionId } : {}),
```
ferhatelmas, line 729:
```suggestion
          ...(options?.versionId ? { versionId: options.versionId } : {}),
```

I would suggest this actually and update the other instead but leaving it to you and @mandarini for final

packages/core/storage-js/test/storageFileApi.test.ts
```diff
@@ -854,6 +854,46 @@ describe('Object API', () => {
       assert(parsedUrl.searchParams.has('token'))
     }
   })
+
+  it('will append a versionId parameter to getPublicUrl', () => {
```
ferhatelmas: nit: they are inside transformation describe, I think they don't belong here

ferhatelmas (dismissed):
inline comments are tiny, approving

packages/core/storage-js/src/packages/StorageFileApi.ts
mandarini: I think I will keep `!= null` because it matches the existing versionId/cacheNonce checks elsewhere in this file..

mandarini (dismissed):

packages/core/storage-js/src/packages/StorageFileApi.ts
TylerHillery: I don't have strong preference either way I was just following `options?.cacheNonce != null)` as convention for optional query param. Just let me know which way is preferred and I will implement.

packages/core/storage-js/test/storageFileApi.test.ts
TylerHillery: moved

mandarini (approved):

ferhatelmas (approved):

Issue #1661, opened 2023-09-26
`count: exact` and `head: true` hides errors


steve-chavez:
## Problem

If a table has a high count (say 30_000_000), doing:

```js
supabase
.from('table')
.select('id', {count: 'exact', head: true})
```

Will result in an uninformative error:

```js
{
error: { message: '' },
data: null,
count: null,
status: 500,
statusText: 'Internal Server Error'
}
```

By removing the `head: true` (untested, but assuming is the same as `curl`), we get the actual error:

```
{"code":"57014","details":null,"hint":null,"message":"canceling statement due to statement timeout"}
```

## Solution

Somehow expose the error message when this happens. Not sure if possible, this might need a special status code from upstream.

mandarini:
I will be closing this issue since it has no activity for more than 2 years. If you find that the issue persists in the latest version of the libraries, please feel free to open a new issue.

Thank you for your patience and for contributing to supabase!

laurenceisla:
@steve-chavez  I think this is solved by the [Proxy Status Header](https://docs.postgrest.org/en/v14/references/errors.html#proxy-status-header), right? Just tried it in Supabase-JS and it returns `proxy-status: PostgREST; error=57014` as a response header.

PR #2685, opened 2026-09-13
fix(postgrest): surface errors from failed HEAD requests

### LINKED ISSUE #2683: postgrest-js: a failed HEAD request (count: 'exact', head: true) yields an error with an empty message and no status
## Bug report

**Package:** `@supabase/postgrest-js` (`packages/core/postgrest-js`). Filing here because `supabase/postgrest-js` is archived and read-only.

A failed **HEAD** request — the shape `select('id', { count: 'exact', head: true })` produces — resolves with an error that carries **no information at all**: `{ message: '' }`, with no `code`, no `details`, no `hint` and no `status`. The HTTP status is on the *result*, not on the error, so any consumer that inspects `error` alone cannot tell a transient gateway failure from a permission denial.

The cause is that a HEAD response has no body to parse (HTTP forbids one), so `JSON.parse('')` throws and the catch assigns `error = { message: body }`, i.e. `{ message: '' }`.

### To reproduce

A gateway returning 504 for both methods. It sends the same body each time; HTTP strips it from the HEAD response.

```js
import { createServer } from 'node:http'
import { PostgrestClient } from '@supabase/postgrest-js'

const server = createServer((req, res) => {
  res.writeHead(504, { 'Content-Type': 'application/json' })
  if (req.method === 'HEAD') res.end()
  else res.end(JSON.stringify({ message: 'Gateway Timeout' }))
})
await new Promise((r) => server.listen(0, r))

const postgrest = new PostgrestClient(`http://localhost:${server.address().port}`)

console.log(await postgrest.from('t').select('id', { count: 'exact', head: true }))
console.log(await postgrest.from('t').select('id'))

server.close()
```

### Observed

```
// HEAD — select('id', { count: 'exact', head: true })
{ error: { message: '' }, data: null, count: null, status: 504, statusText: 'Gateway Timeout' }

// GET — select('id')                     same status, same server
{ error: { message: 'Gateway Timeout' }, data: null, count: null, status: 504, statusText: 'Gateway Timeout' }
```

`'status' in error` is `false` in both cases, which only matters for the HEAD one, where the error is otherwise empty.

Reproduced on **2.116.0** (latest at the time of writing). We hit it in production on 2.105.4, and the same code path is present in 2.100.0.

### Expected

A failed request should say *something* about why it failed, even with no body. Either would be enough:

- populate `error.status` from the response status, or
- when the body is empty, give `message` a non-empty value derived from the response, e.g. `HTTP 504` or the `statusText`.

### Why it matters

Anything that classifies failures reads the error object. With an empty message and no status, a bodiless 5xx is indistinguishable from an error the request earned by executing, so it cannot be retried or reported on its merits. In our case a platform gateway returned 504 on the first request of newly started serverless functions; every such failure on a GET or POST carried a body, was recognised as transient and recovered on retry, while the HEAD counts silently did not — for 24 h, 8 counts failed with "unknown error" and no retry, and the only usable signal was a status the library had already discarded.

The existing reports are close but neither covers this:

- #1661, "`count: exact` and `head: true` hides errors" (2023) — same empty-error shape, there from a statement timeout. Closed 2025-11-03 for inactivity, with the note to open a new issue if it persists in the latest version. It does (above). A comment there suggests PostgREST's [Proxy Status Header](https://docs.postgrest.org/en/v14/references/errors.html#proxy-status-header), which helps when PostgREST itself answers, but not when something in front of it returns a bodiless 5xx.
- supabase/postgrest-js#404 and #405 are about `statusText` on **success**, and supabase/postgrest-js#381 is an empty error where the response **did** have a JSON body.

### Workaround

Copy the result's status onto the error before classifying it, since the library keeps the two apart:

```js
const res = await postgrest.from('t').select('id', { count: 'exact', head: true })
const failure = res.error ? { ...res.error, status: res.error.status ?? res.status } : null
```

Ours is commit `c01ad47` (private repo): the retry helper now does exactly this, and the failure is reported as `HTTP 504` instead of "unknown error".

Happy to send a PR if one of the two suggested shapes is preferred.

vegtelenseg:
## 🔍 Description

Failed `HEAD` requests (`select(..., { count: 'exact', head: true })`) currently resolve with an empty error object, and an empty-bodied 404 is reported as success. This PR maps those failures onto the existing error shape.

### What changed?

`PostgrestBuilder.processResponse` parses a failed response body as the PostgREST error object and falls back to `{ message: body }` when that fails. HTTP forbids a body on a `HEAD` response, so a failed `HEAD` request always takes that fallback with `body === ''`. Three changes:

1. The issue 295 conversion of an empty-bodied 404 into `204 No Content` is now gated on the request being able to carry a body. A `HEAD` request cannot, so the failure is reported instead of dropped. The original 295 path (PATCH/DELETE with no match) is unchanged.
2. When there is no body to describe the failure, `message` falls back to `HTTP <status>` instead of `''`.
3. When PostgREST sets `Proxy-Status: PostgREST; error=<code>`, that code becomes `error.code`. Entries from other proxies are ignored, since `code` holds PostgREST and PostgreSQL codes.

`select('id', { count: 'exact', head: true })` resolved with `{ message: '' }` on every failure: no `code`, `details` or `hint`. Consumers that classify failures from `error` alone could not tell a transient gateway failure from a permission denial. The status was on the result, so the information existed, but nothing in `error` reflected it. PostgREST already publishes its code in `Proxy-Status` for exactly this case, and the client was not reading it.

Closes #2683

Refs #1661, which reported the same empty error from a statement timeout and is covered by change 3.

## 📸 Screenshots/Examples

Local PostgREST 14.16, same URL, `HEAD` versus `GET`:

| Request | Before | After |
| --- | --- | --- |
| `HEAD` bad filter | `400 { message: '' }` | `400 { message: 'HTTP 400', code: 'PGRST100' }` |
| `HEAD` missing table | `204 error: null, success: true` | `404 { message: 'HTTP 404', code: 'PGRST205' }` |
| `HEAD` bodiless gateway 504 | `504 { message: '' }` | `504 { message: 'HTTP 504' }` |
| `HEAD` healthy table | `200 count: 0` | unchanged |

## 🔄 Breaking changes

The `HEAD` 404 that previously resolved as `204` / `error: null` was the issue 295 workaround matching a request that cannot have a body. That is a dropped error, not a supported success path. Callers of `.update()` / `.delete()` with no match still get the same 204 conversion.

## 📋 Checklist

## 📝 Additional notes

`HTTP <status>` is the one new observable string here. Precedent is the network-error path in the same file, which synthesizes `"TypeError: fetch failed"` and is asserted with `toBe` in `fetch-errors.test.ts`. If you would rather not add it, dropping it is a one-line change and the other two fixes stand alone.

`details` and `hint` remain absent on this fallback, so `.throwOnError()` still yields a `PostgrestError` with those `undefined`. Filling them would add keys to every unparseable-body error, including the body-present case settled in #2398, so it seems to belong to the separate refactor described in that review.

No `migrations/` note: the 295 workaround is unchanged for methods that can carry a body, and there is no public API addition.

Tests: nine cases in `test/empty-body-errors.test.ts`. Seven of them fail with this PR's source change reverted; the other two assert behavior this PR deliberately leaves unchanged. The full `postgrest-js` suite passed locally (373 tests, 285 snapshots) plus 31 `tstyche` files. Affected packages (`postgrest-js`, `supabase-js`) build against `upstream/master`.

@AirRocker offered to test against their production setup. A maintainer can add the `trigger: preview` label if a `pkg.pr.new` build would help with that.


mavadojunior (commented):
K

PR #2686, opened 2026-09-13
fix(auth): return the stored session when a mid-flight refresh is discarded

ebroder:
### What changed?

`__loadSession` now treats `AuthRefreshDiscardedError` from `_callRefreshToken` as its own case. When a refresh is discarded because storage was rotated mid-flight, it re-reads storage and, if what it finds is a valid session whose access token has not itself expired, returns that session with `error: null` instead of translating the discard into `{ session: null }`. The cleared-storage case (a concurrent `signOut`) finds nothing to adopt and still returns `{ session: null }`. The adopted session comes from storage (the newer one the guard preserved); an adopted session must pass `_isValidSession` and still be within its own real expiry, so no path hands back a dead token as `error: null`.

`_callRefreshToken`'s commit-guard contract and the proactive-preserve branch for other refresh errors are unchanged. `setSession()` and `refreshSession({ refresh_token })` pass their own token to `_callRefreshToken` and never reach this path. No-arg `refreshSession()` does read the adopted session back through `__loadSession` and refreshes it rather than surfacing the discard (a concurrency signal, not a refresh failure, per #2668); genuine refresh failures still surface.

`_callRefreshToken` has a commit guard (introduced in #2392, extended in #2436): it snapshots storage before the `/token` fetch and re-reads after, and if the stored `refresh_token` changed in between it discards the rotated tokens and returns `AuthRefreshDiscardedError`. That is intended for two cases: storage cleared by a concurrent `signOut`, and storage replaced by another writer.

The problem is one level up. On any refresh error, `__loadSession` only preserved the current session when its access token was still valid *and* the same refresh token was still stored. In the replaced-storage case the second condition is false precisely because another writer rotated the token, so `getSession()` returned `{ session: null }` even though storage held a valid, newer session (the very session the guard preserved storage for). If the old access token had also passed its real expiry, the preserve branch was skipped and the result was `null` too. Every concurrent `getSession()` caller sharing that refresh via `refreshingDeferred` saw the same `null`.

It reproduces with two tabs sharing cookie storage (via `@supabase/ssr`), no server involvement:

| dt (ms) | tab | event |
|---|---|---|
| 0 | A | `POST /token` with R0 |
| 1 | B | `POST /token` with R0 (network-delayed 1.5s) |
| 158 | A | 200, R1 -> A writes storage |
| 285 | A | `GET /rest/v1/trips` with user JWT -> 200 |
| 1571 | B | 200, R1 (same child) |
| 1582 | B | `GET /rest/v1/trips` with the anon key -> 406 |

Storage changed exactly once (R0 -> R1, by tab A) and held a valid session throughout, but tab B's `getSession()` returned `null`. The same thing happens in a Next.js app when the `@supabase/ssr` middleware or route handler rotates the token and its `Set-Cookie` lands during the browser client's own refresh.

The downstream effect is what makes it sharp. `SupabaseClient._getSessionToken` destructures `{ data }` and ignores `error`, so `_getAccessToken` falls back to the anon key and every PostgREST/RPC request in that burst goes out anonymous. Under RLS that surfaces as a 406 on `.single()`, a 401 on tables without an anon grant, and a silent empty `200 []` on lists.

In the common case the discarded response even carries the *same* refresh token that is now in storage (GoTrue's v1 parent-of-active path hands the already-issued child to whoever reuses the parent), so the discard throws nothing away and then reports failure anyway.

A note on the history here. #2392's rationale for the guard was that cross-tab races need no client coordination because GoTrue's parent-of-active path handles them; returning `null` contradicts that. #2436 added the `stillStored` fallthrough specifically to stop `__loadSession` "translating the refresh error into `{ session: null, error }`", and this bug is that same failure mode reached through the replaced-storage path rather than the cleared-storage one it was written for. #2665 (fixed by #2668) is the only prior `AuthRefreshDiscardedError` report, but it covered the signOut path's logging noise only ("requests return HTTP 200"), so it never surfaced this.

## 📸 Screenshots/Examples

Covered by a new `describe('getSession() after a discarded mid-flight refresh')` block in `GoTrueClient.test.ts`, all driving the public `getSession()`:

- the old session is inside the 90s expiry margin (proactive refresh) but not actually expired,
- the old session is already past its real expiry,
- the same-token variant (the discarded response and the stored session carry the same refresh token),
- a regression guard that a mid-flight `signOut` still yields `session: null`,
- the adopted session is itself already past expiry, so it is not handed back as valid.

The adopt cases fail before the change (returning `AuthRefreshDiscardedError`) and pass after. The four existing commit-guard tests are unchanged. Locally: `pnpm nx build auth-js`, `pnpm nx format`, and the discard + commit-guard suites pass; CI runs the full Docker-backed suite.

## 🔄 Breaking changes

## 📋 Checklist

## 📝 Additional notes

This fix is intentionally scoped to the discard adoption. Adjacent concerns left out, as follow-ups:

- `supabase-js`'s `_getSessionToken` swallows the `error` from `getSession()`. Even with this change, any future path that returns `{ session: null, error }` will still fall back silently to the anon key. That destructure is worth revisiting on its own.
- No `TOKEN_REFRESHED` fires for a session adopted after a discard. That is arguably correct (this client did not perform the rotation), but listeners relying on the event won't see the adopted session until the next real refresh.
- The adopted session is read straight from primary storage, so under the experimental `userStorage` option its `user` lives in a separate record and would need rehydrating (as the non-expired path does) before return. Narrow, experimental-only; left for a follow-up rather than widening this diff.
- `getSession()` on the lockless path is best-effort against a concurrent `signOut`: the `userStorage` read (and any other await between reading storage and returning) is an interleaving point. Fully closing that window (and adopting the newest record if storage rotates again mid-read) belongs with the lock path, not here.


ebroder:
@coderabbitai full review

PR #2687, opened 2026-09-14
feat: hashed auth storage key to prevent collisions

staaldraad:
## 🔍 Description

Uses a new hash-key for the default auth storage key. This prevents collisions on self-hosted or shared domain deployments, or when two hosts share a common subdomain (such as `api`).

### What changed?

- The default auth storage key is now `sb-<hash>-auth-token`, where the hash is FNV-1a 32-bit of the Supabase URL origin. Path is ignored, so `https://x.supabase.co` and `https://x.supabase.co/` agree, while `localhost:54321` and `localhost:8000` no longer collide.
- `getDefaultStorageKey(url)` is exported from `@supabase/supabase-js` so `@supabase/ssr` and server cookie readers can compute the same key.
- auth-js gained a `legacyStorageKeys` option. On init it moves the session, its `-user` entry and pending PKCE verifiers from the old key to the new one, then removes the old entries. If the new key already holds a session, the old entries are just removed. supabase-js passes the old `sb-<ref>-auth-token` key only when the caller did not set `auth.storageKey`, so existing users stay signed in and custom keys are untouched.

In the case of two hostnames, api.hostname.com and api.hostname2.com sharing a common subdomain, the cache key would be shared.

## 🔄 Breaking changes

This should not be breaking, the `legacyStorageKeys` should allow existing sessions to migrate across.

## 📋 Checklist

## 📝 Additional notes

Mainly authored by Claude


mandarini:
@staaldraad I am not sure I will have time to look into in this week. Is this urgent? If yes, I will make time

staaldraad:
> @staaldraad I am not sure I will have time to look into in this week. Is this urgent? If yes, I will make time

no, not urgent. I can park it in draft for until after SELECT. Thanks! 🙏🏼

